PR #29, opened 2020-09-09
Increase Speed

nils-braun:
Currently, a lot of dask computation is spent in renaming or assigning new columns instead of the "real" calculation, as - to be consistent with what the Relational Algebra gives to us - the columns of the intermediate dataframes are renamed quite often.

This WIP PR includes a new data type, which contains the real data as well as a frontend column mapping. Typically, physical plans will only operate on the data itself and do not care so much about the column names, so we can store the columns separately to the real data.

In my very small tests, this brings dask-sql approximately en-par with usual dask calls in simple cases, such as 
```sql
SELECT a, MAX(b) FROM df
```

The PR still needs documentation and more unittests for the new classes.

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -4,6 +4,7 @@ from dask_sql.java import DaskSchema, DaskTable, RelationalAlgebraGenerator
 from dask_sql.mappings import python_to_sql_type
 from dask_sql.physical.rel import RelConverter, logical
 from dask_sql.physical.rex import RexConverter, core
+from dask_sql.datacontainer import DataContainer, ColumnContainer
 
 
 class Context:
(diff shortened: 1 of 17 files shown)


mrocklin:
We could also try to make Dask Dataframe renaming faster?

nils-braun:
@mrocklin That is of course also true! However, to make life a bit easier, there are many "unneeded" renames or column reordering done in `dask-sql` (which are only done for bookkeeping reasons). 
No matter what, the renaming will still need to go to each of the pandas dataframes and actually do the renaming, so there will always be a cost (and most of the renames can be completely ignored because they are just there for bookkeeping).

Maybe I should not call it renaming: it is also creating columns out of already present columns (again, mostly for bookkeeping) - which will always involve some data copying if I want to do it directly on the dataframe. So it is definitely not dask's fault - but more my misusing it :-)

PR #30, opened 2020-09-10
Added nicer error messages on parsing failures

### LINKED ISSUE #4: Show better error messages
When there is a SQL validation error, the error is shown on the screen. However, we always get a nested Java and Python exception, which might lead to confusion.
The blazingSQL project created their own SQL exception in Java for this - we might also get along with just catching and parsing the Java exception in python and printing out a piece of more useful information to the user in the `Context.sql` function.

nils-braun:
Fixes #4 

Instead of showing all the traceback from calcite (java) -> jpype -> python we only show a single, hand-made exception now, which gives information on the parsing error and where it happened.

For the example given in the issue #4, this would now look like

```
Traceback (most recent call last):
  File "test_me.py", line 7, in <module>
    context.sql("SELECT * from timeseries where name=Tim")
  File "/home/nils/projects/private/dask-sql/dask_sql/context.py", line 92, in sql
    raise ParsingException(sql, str(e.message())) from from_chained_exception
dask_sql.utils.ParsingException: Can not parse the given SQL: org.apache.calcite.runtime.CalciteContextException: From line 1, column 37 to line 1, column 39: Column 'Tim' not found in any table

The problem is probably somewhere here:

        SELECT * from timeseries where name=Tim
                                            ^^^
```

I took the idea to show ^ indicators from clang and blazingSQL.

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -1,9 +1,16 @@
 import dask.dataframe as dd
 
-from dask_sql.java import DaskSchema, DaskTable, RelationalAlgebraGenerator
+from dask_sql.java import (
+    DaskSchema,
+    DaskTable,
+    RelationalAlgebraGenerator,
+    ValidationException,
+    SqlParseException,
+)
 from dask_sql.mappings import python_to_sql_type
 from dask_sql.physical.rel import RelConverter, logical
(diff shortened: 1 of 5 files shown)


mrocklin:
Woot.

PR #35, opened 2020-09-13
[WIP] Added a cast operator

nils-braun:
As the title says, this PR implements the `CAST` operator, which converts between different types.
The PR is only WIP so far because I guess that many translations, which are able to do in SQL, are not implemented so far an I need to this this properly.


nils-braun:
This PR is very outdates, therefore I am closing it

PR #39, opened 2020-09-17
Documentation

nils-braun:
Added documentation and readthedocs integration. Currently available at https://dask-sql.readthedocs.io/en/feature-documentation/index.html.

@kyprifog Would you mind having a look into the created documentation and tell me, if you think there is something missing? Would be very nice!


nils-braun:
I will merge this PR now, so that the deviation between the documentation and the actual code does not increase that much.
If someone would like to comment later, I am more than happy to hear any thoughts on the docu!

mrocklin:
You may want to consider using the Dask Sphinx theme:
https://github.com/dask/dask-sphinx-theme

nils-braun:
Thanks @mrocklin, that makes sense. I have added it in #52

Issue #38, opened 2020-09-16
Implement <>


kyprifog:
Currently
```SELECT * from $table WHERE NOT(col_b = 123.0)```
results in

```
    def convert(
        self, rex: "org.apache.calcite.rex.RexNode", df: dd.DataFrame
    ) -> Union[dd.Series, Any]:
        # Prepare the operands by turning the RexNodes into python expressions
        operands = [RexConverter.convert(o, df) for o in rex.getOperands()]
    
        # Now use the operator name in the mapping
        operator_name = str(rex.getOperator().getName())
    
        try:
>           operation = self.OPERATION_MAPPING[operator_name]
E           KeyError: '<>'
```

Issue #37, opened 2020-09-16
Implement IS NOT NULL


kyprifog:
Reminder for later.  Currently
```SELECT * from $table WHERE col_b IS NOT NULL```
results in:

```    
try:
            operation = self.OPERATION_MAPPING[operator_name]
        except KeyError:
>           raise NotImplementedError(f"{operator_name} not (yet) implemented")
E           NotImplementedError: IS NOT NULL not (yet) implemented
```

Issue #11, opened 2020-08-27
Implement more expression operators


nils-braun:
The `RexCallPlugin` currently can only handle the most common expressions (binary operations, or and and). There are many more of them that could and should be implemented. 
Similar to the aggregate functions in #9: One should also check, how many of them are actually in the calcite language standard and if choosing a different SQL dialect makes more sense.

nils-braun:
First (maybe unfinished) set of operations, taken from [SqlFunctions.java](https://github.com/apache/calcite/blob/master/core/src/main/java/org/apache/calcite/runtime/SqlFunctions.java)

- [x] -
- [x] *
- [x] /
- [x] +
- [x] <
- [x] <=
- [x] =
- [x] <>
- [x] >
- [x] >=
- [ ] ABS
- [ ] ACOS
- [ ] ASIN
- [ ] ATAN
- [ ] ATAN2
- [ ] BIT_AND
- [ ] BIT_OR
- [ ] BIT_XOR
- [ ] CBRT
- [ ] CEIL
- [ ] COS
- [ ] COSH
- [ ] COT
- [ ] DEGREES
- [ ] EXP
- [ ] FLOOR
- [ ] LOG10
- [ ] MOD
- [ ] POWER
- [ ] RADIANS
- [ ] ROUND
- [ ] SIGN
- [ ] SIN
- [ ] SINH
- [ ] TAN
- [ ] TANH
- [ ] TRUNCATE

PR #40, opened 2020-09-18
Implement not and neq

nils-braun:
Fixing #37 and #38.

@kyprifog Would you like to check, if this solves your use-case?

### DIFF
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -111,7 +111,10 @@ class IsNullOperation(Operation):
         """
         Returns true where `df` is null (where `df` can also be just a scalar).
         """
-        return np.isnan(df)
+        if is_frame(df):
+            return df.isna()
+
+        return df is None or np.isnan(df)
 
 
 class LikeOperation(Operation):
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -64,6 +64,22 @@ class CaseOperation(Operation):
             return then if where else other  # pragma: no cover
 
 
+class NotOperation(Operation):
+    """The not operator"""
+
+    def __init__(self):
+        super().__init__(self.not_)
+
+    def not_(self, df: Union[dd.Series, Any],) -> Union[dd.Series, Any]:
+        """
+        Returns not `df` (where `df` can also be just a scalar).
+        """
+        if is_frame(df):
+            return ~df
+        else:
+            return not df  # pragma: no cover
+
+
 class LikeOperation(Operation):
     """The like operator (regex for SQL with some twist)"""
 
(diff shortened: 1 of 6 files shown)


nils-braun:
Also related to #11

kyprifog:
I was able to test
```
SELECT * from table where NOT col = 'value'
```
and this seems to work properly.   I'm still trying to figure out how to test
```
SELECT * from table where col IS NOT NULL
```
where i'm getting an error related to isnan.   I have to decide if this even makes sense, this aspect of dask/pandas is a bit more confusing than spark.

martindurant:
^ @tomaugspurger , want to comment on the state of pandas nullable floats? I would say keep using pandas style `isnan`, and don't worry about NaN

TomAugspurger:
Open PR at https://github.com/pandas-dev/pandas/pull/34307.

nils-braun:
@kyprifog Thanks for testing! Could you maybe share some test data and the error message for the `NULL` case? Because for me, it seems to work:

```python
import dask
import dask_sql
import numpy as np

context = dask_sql.Context()
df = dask.datasets.timeseries()

# Fill in some NaN here and there
df["x"] = df["x"].apply(lambda x: np.NaN if x > 0 else x, meta=("x", "float64"))

context.register_dask_table(df, "timeseries")
print(context.sql("SELECT * FROM timeseries WHERE x IS NOT NULL", debug=True,).compute())
```

@martindurant I tried to read through the linked discussion (thanks for that!), but I am now a bit unsure what the current best usage is. Are you proposing to move to `df.isna` instead of `np.isnan`? Sorry if I understood this incorrectly...

kyprifog:
@nils-braun I think the problem I was facing is most closely exemplified if you change your query above:
```
SELECT * FROM timeseries WHERE name IS NOT NULL
```
IE looking for `None` or its equivalent in a column with dtype O (object), which results in
```
E           TypeError("ufunc 'isnan' not supported for the input types, and the inputs could not be safely coerced to any supported types according to the casting rule ''safe''")
```
Object is the default datatype inferred by dask for any string columns with missing values so this seems like a pretty common occurence.   Digging deeper it looks like that since object is a less specific in these cases the type for missing values is float.nan (even though the other values have type str), so this may be the connection that @martindurant is driving at.   This isn't strictly a sql-dask problem, but looking at pandas documentation it seems like there is still talk of a concept of nullable type for object dtype:  
https://pandas.pydata.org/pandas-docs/stable/user_guide/missing_data.html#missing-data

IE it should behave roughly like:
```
df.dropna(subset=["name"])
```
which works as expected.

nils-braun:
Thank you @kyprifog for the excellent description. Now I finally understand :-)
This means, if moving to `df.isna` (which I did in the PR), should already solve the problem. At least your example use case works now:

```python
df["name"] = df["name"].apply(
    lambda x: None if x.startswith("A") else x, meta=("x", "object")
)

sql = "SELECT name IS NOT NULL FROM timeseries"
```

I have also added tests for this.

kyprifog:
Tested on my end and works, and fix makes sense :+1:

PR #56, opened 2020-10-13
Implement a SQL server speaking the presto protocol

nils-braun:
This server can be used to run dask-sql in a standalone application.
It is e.g. possible to run it in a dask-cluster and answer SQL queries from external.

More information can be found in the README or the documentation.


nils-braun:
Fixing #6 .

PR #63, opened 2020-10-28
Unify dask-sql API with blazing SQL

### LINKED ISSUE #58: Unify the API of dask-sql and blazing-sql
After a very nice discussion with @felipeblazing (thanks again for reaching out!), we decided to unify the APIs of dask-sql and [blazing-sql](https://github.com/BlazingDB/blazingsql), so that users can switch between the different packages more easily (and therefore switch between "normal" CPU dask and GPU-enabled rapids). As blazing has been around much longer already, I think it would make sense to use blazings API as the target for dask-sql.

The entrypoint to all applications with dask-sql or blazing is to instantiate a context and then call methods on it.
blazing currently has the following methods on the context:

```
File System:
	.s3
	.localfs
	.hdfs
	.gs
	.show_filesystems

SQL:
	.explain
	.create_table
	.drop_table
	.sql
```

The SQL methods can be implemented quite quickly as they are basically already present (just with different names).
The FS methods are very interesting to have (see #28), so they definitely should come next. 
There is another method, `.log`, which however only makes sense in the blazing context and is very hard to port. 

Apart from the context API, there is of course also the very important question if the same SQL queries are understood. 
That should be handled next.

nils-braun:
Fixes #58 

I have implemented (parts of) the blazingSQL API of the context - especially `.sql`, `.create_table`, `.explain` and `.drop_table`.

What is still missing:
- [x] let create_table also work on a hive cursor
- [ ] ~think about implementing all the file system functions (e.g. `.s3`). However - in dask they are not really needed as all the parameters can just be passed via the call to `create_table`.~ -> I decided against implementing them for now. Everything can be achieved with the methods already present. Maybe in a later step we can add them also here.

### DIFF
--- a/README.md
+++ b/README.md
@@ -49,7 +49,7 @@ c = Context()
 # Load the data and register it in the context
 # This will give the table a name
 df = timeseries()
-c.register_dask_table(df, "timeseries")
+c.create_table("timeseries", df)
 
 # Now execute an SQL query. The result is a dask dataframe
 # The query looks for the id with the highest x for each name
--- a/README.md
+++ b/README.md
@@ -19,7 +19,7 @@ Read more in the [documentation](https://dask-sql.readthedocs.io/en/latest/).
 
 You can try out `dask-sql` quickly by using the docker command
 
-    docker run --rm -it -p 8080:8080 nils-braun/dask-sql
+    docker run --rm -it -p 8080:8080 nbraun/dask-sql:0.1.2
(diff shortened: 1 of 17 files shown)


nils-braun:
The PR contains a first experimental implementation of an hive input.
It is not really well tested so far (and I have also not written unittests for it) - so I guess this part should be improved in the next release.

PR #68, opened 2020-11-02
Input from cluster memory

nils-braun:
This PR allows to input published datasets from the cluster memory.

This means, if you have published a dataset, e.g. with a distributed client
```python
client.publish_dataset(my_table=df)
```

you can now register it with
```sql
CREATE TABLE some_name WITH (location='my_table', format='memory')
```

and start querying it as normal.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,0 +1,3 @@
+import pytest
+
+pytest_plugins = ["tests.integration.fixtures", "distributed.utils_test"]
--- a/dask_sql/java.py
+++ b/dask_sql/java.py
@@ -10,16 +10,20 @@ import logging
 
 import jpype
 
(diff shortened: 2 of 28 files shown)


mrocklin:
This is a nice intersection of an old Dask feature (published datasets) and
the Dask-SQL context.  I'm curious if there are things we should consider
doing on the Dask side that we could learn from the context approach.

nils-braun:
Thanks for commenting, @mrocklin (I am always amazed how fast you find out about what is going on in the ecosystem).

I am not sure if these things can be compared so easily because the context object is local to the process it is running on (or, if running a SQL server, still local to the application but can be controlled via SQL) and the published datasets are always shared among the full cluster.

PR #75, opened 2020-11-06
Increase speed and parallelism of the limit algorithm and implement descending sorting

nils-braun:
This PR introduces four changes:
* it simplifies the logic for getting a piece of the dask dataframe starting at OFFSET and ending at LIMIT.
* this allows to not needing to calculate the dask dataframe before, but really only at the actual calculation. The calculated used to be needed for getting the partition boundaries, but they can now also only be calculated "lazily"
* as there is no need to do a full recalculation, I introduced a quick shortcut: if the first partition is already enough, just return this. This is a very typical use case when you just do a "LIMIT 10".
* it allows for descending sorting also in the first column

### DIFF
--- a/dask_sql/physical/rel/logical/sort.py
+++ b/dask_sql/physical/rel/logical/sort.py
@@ -76,7 +76,7 @@ class LogicalSortPlugin(BaseRelPlugin):
         if not first_sort_ascending:
             # As set_index().reset_index() always sorts ascending, we need to reverse
             # the order inside all partitions and the order of the partitions itself
-            df = df.map_partitions(self.reverse_partition, meta=df)
+            df = df.map_partitions(lambda partition: partition[::-1], meta=df)
             df = df.partitions[::-1]
 
         # sort the remaining columns if given
@@ -154,8 +154,3 @@ class LogicalSortPlugin(BaseRelPlugin):
                 for partition_number, partition in enumerate(df.partitions)
             ]
         )
-
-    @staticmethod
-    def reverse_partition(partition):
-        """Small helper function to revert a partition in dask"""
-        return partition[::-1]
--- a/dask_sql/physical/rel/logical/sort.py
+++ b/dask_sql/physical/rel/logical/sort.py
@@ -61,11 +61,6 @@ class LogicalSortPlugin(BaseRelPlugin):
         first_sort_column = sort_columns[0]
         first_sort_ascending = sort_ascending[0]
 
-        # Sort the first column with set_index. Currently, we can only handle ascending sort
-        if not first_sort_ascending:
-            raise NotImplementedError(
-                "The first column needs to be sorted ascending (yet)"
-            )
         # We can only sort if there are no NaNs or infs.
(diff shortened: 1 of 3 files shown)


dask_sql/physical/rel/logical/sort.py
```diff
@@ -79,6 +73,12 @@ def _apply_sort(
             raise ValueError("Can not sort a column with NaNs")
 
         df = df.set_index(first_sort_column, drop=False).reset_index(drop=True)
+        if not first_sort_ascending:
+            # As set_index().reset_index() always sorts ascending, we need to reverse
+            # the order inside all partitions and the order of the partitions itself
+            df = dd.from_delayed(
+                reversed([self.reverse_partition(p) for p in df.to_delayed()])
+            )
```
mrocklin, line 81: You may want to become familiar with the `df.partitions` iterable. 

```python
dd.concat(df.partitions[::-1])
```

Going between delayed and dataframes introduces some inconvenience in graph-handling.

dask_sql/physical/rel/logical/sort.py
```diff
@@ -104,24 +103,32 @@ def _apply_offset(self, df: dd.DataFrame, offset: int, end: int) -> dd.DataFrame
...
         """
-        # As we need to calculate the partition size, we better persist
-        # the df. I think...
-        # TODO: check if this is the best thing to do
-        df = df.persist()
+        if not offset:
+            # We do a (hopefully) very quick check: if the first partition
+            # is already enough, we will just ust this
+            first_partition_length = df.map_partitions(lambda x: len(x)).to_delayed()[0]
+            first_partition_length = first_partition_length.compute()
```
mrocklin, line 110:
```python
len(df.partitions[0])
```

dask_sql/physical/rel/logical/sort.py
nils-braun: Uh nice! That looks awesome. Yes, this is definitely much better.

mrocklin: At some point I think that we'll want to go through and remove all of the `to/from_delayed` calls.  This isn't critical, but it's nice to do from a performance standpoint (happy to go into this in more depth if you like).

nils-braun: Thanks for your advice, I would actually be really interested in your thoughts about [this part](https://github.com/nils-braun/dask-sql/pull/75/files#diff-1fcfb9a4ff196f6a0e8fb3847cc3d2300b9b3a91af77f0555e6a61a179757802R153). 

Probably you see it right away, but what I would like to achieve is basically two things
1. call a function on every partition, while also having the partition index as an input argument
2. calculate `partition_borders`  only when executing the DAG - not already when creating it (this is what I have done before this PR, and it slows down the process).

So far, the `delayed` function was all I could come up with, but I definitely think there is a better way.

nils-braun: I have created an issue to fix this, so we can go on with this PR.

PR #98, opened 2020-12-30
DROP TABLE and IF NOT EXISTS/REPLACE

### LINKED ISSUE #97: `DROP table x` support
A user can do:

`CREATE table demo as select 1 as n;`

But then how to delete it?

nils-braun:
This PR introduces a new SQL statement
```sql
DROP TABLE [IF EXISTS] table
```
and adds support for `CREATE OR REPLACE` and `IF NOT EXISTS` to `CREATE TABLE` and `CREATE VIEW`.
It fixes #97.

During this, I have also refactored the freemarker templates and the implemented SqlNodes (and based them on top of the buildin objects in calcite if possible).

Still to do in this PR:
* documentation
* some code comments why I am not using the Apache Calcite classes directly


nils-braun:
This PR is waiting for #104 - except that it is already finished.

Issue #104, opened 2021-01-03
Add extended SQL documentation


nils-braun:
With the most recent additions, the SQL statements that dask-sql understands have grown. It might make sense to add a page for every SQL statement instead of having them all on a single documentation page

PR #105, opened 2021-01-03
ANALYZE TABLE

### LINKED ISSUE #96: Support Analyze Query
This could very well be difficult to implement but the [ANALYZE](https://dev.mysql.com/doc/refman/5.6/en/analyze-table.html) query would be useful to have.  For reference this is what spark currently supports with respect to `ANALAYZE`:
https://spark.apache.org/docs/3.0.0/sql-ref-syntax-aux-analyze-table.html

nils-braun:
Fixes #96 

This PR introduces a new SQL statement: 
```sql
ANALYZE TABLE <table> COMPUTE STATISTICS [FOR ALL COLUMNS | FOR COLUMNS a, b, ...]
```

Thanks to @quasiben for bringing it up. The query will not do any optimizations (as its pendant in e,g, spark), but will just output a table computed on the fly with the following properties:
```
"count",
"mean",
"std",
"min",
"25%",
"50%",
"75%",
"max",
"data_type",
"col_name",
```

If you have any more ideas on what to implement, I am happy to add this.

The PR is waiting for #104 for the documentation.


quasiben:
Thanks @nils-braun !

PR #118, opened 2021-01-20
add fsql

goodwanghan:


dask_sql/integrations/fugue.py
```diff
@@ -50,3 +54,23 @@ def __init__(self, *args, **kwargs):
...
         super().__init__(*args, **kwargs)
         self._default_sql_engine = DaskSQLEngine(self)
+
+
+def fsql(
+    sql: str,
+    ctx: Optional[Context] = None,
+    register: bool = False,
+    fugue_conf: Any = None,
+) -> Dict[str, dd.DataFrame]:
```
nils-braun, line 64: Would you like to add some docstring? 
We could then also add it to the documentation page.

nils-braun (commented):
That is super cool!

dask_sql/integrations/fugue.py
goodwanghan: yes, added

PR #135, opened 2021-02-09
Add implementation of SQL Except operation

demianw:
There is also a change in an error message which is needed for the presto SQL adapter to work.

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -86,6 +86,7 @@ class Context:
         RelConverter.add_plugin_class(logical.LogicalSortPlugin, replace=False)
         RelConverter.add_plugin_class(logical.LogicalTableScanPlugin, replace=False)
         RelConverter.add_plugin_class(logical.LogicalUnionPlugin, replace=False)
+        RelConverter.add_plugin_class(logical.LogicalMinusPlugin, replace=False)
         RelConverter.add_plugin_class(logical.LogicalValuesPlugin, replace=False)
         RelConverter.add_plugin_class(logical.SamplePlugin, replace=False)
         RelConverter.add_plugin_class(custom.AnalyzeTablePlugin, replace=False)
@@ -515,6 +516,10 @@ class Context:
                 nonOptimizedRelNode = generator.getRelationalAlgebra(validatedSqlNode)
                 rel = generator.getOptimizedRelationalAlgebra(nonOptimizedRelNode)
                 rel_string = str(generator.getRelationalAlgebraString(rel))
+                logger.debug(
+                    f"Non optimised query plan: \n "
+                    f"{str(generator.getRelationalAlgebraString(nonOptimizedRelNode))}"
+                )
         except (ValidationException, SqlParseException) as e:
             logger.debug(f"Original exception raised by Java:\n {e}")
             # We do not want to re-raise an exception here
--- a/dask_sql/physical/rel/logical/minus.py
+++ b/dask_sql/physical/rel/logical/minus.py
@@ -1,0 +1,71 @@
+import dask.dataframe as dd
+
+from dask_sql.physical.rex import RexConverter
+from dask_sql.physical.rel.base import BaseRelPlugin
+from dask_sql.datacontainer import DataContainer, ColumnContainer
+
+
+class LogicalMinusPlugin(BaseRelPlugin):
+    """
+    LogicalUnion is used on EXCEPT clauses.
+    It just concatonates the two data frames.
+    """
+
+    class_name = "org.apache.calcite.rel.logical.LogicalMinus"
+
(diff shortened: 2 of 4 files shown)


dask_sql/context.py
```diff
@@ -515,6 +516,10 @@ def _get_ral(self, sql):
                 nonOptimizedRelNode = generator.getRelationalAlgebra(validatedSqlNode)
                 rel = generator.getOptimizedRelationalAlgebra(nonOptimizedRelNode)
                 rel_string = str(generator.getRelationalAlgebraString(rel))
+                logger.debug(
+                    f"Non optimised query plan: \n "
+                    f"{str(generator.getRelationalAlgebraString(nonOptimizedRelNode))}"
```
nils-braun, line 521: Ok, makes sense to do that. Good idea.

dask_sql/physical/rel/logical/minus.py
```diff
@@ -0,0 +1,69 @@
+import dask.dataframe as dd
+
+from dask_sql.physical.rex import RexConverter
+from dask_sql.physical.rel.base import BaseRelPlugin
+from dask_sql.datacontainer import DataContainer, ColumnContainer
+
+
+class LogicalMinusPlugin(BaseRelPlugin):
+    """
+    LogicalUnion is used on EXCEPT clauses.
```
nils-braun, line 10: I guess you still would like to update that docstring :-)

dask_sql/physical/rel/logical/minus.py
```diff
@@ -0,0 +1,69 @@
...
+    ) -> DataContainer:
+        first_dc, second_dc = self.assert_inputs(rel, 2, context)
+
+        first_df = first_dc.df
+        first_cc = first_dc.column_container
+
+        second_df = second_dc.df
+        second_cc = second_dc.column_container
+
+        # For concatenating, they should have exactly the same fields
```
nils-braun, line 27:
```suggestion
        # For subtracting, they should have exactly the same fields
```

dask_sql/physical/rel/logical/minus.py
```diff
@@ -0,0 +1,69 @@
...
+
+        # To concat the to dataframes, we need to make sure the
+        # columns actually have the specified names in the
+        # column containers
+        # Otherwise the concat won't work
+        first_df = first_dc.assign()
+        second_df = second_dc.assign()
+
+        self.check_columns_from_row_type(first_df, rel.getExpectedInputRowType(0))
+        self.check_columns_from_row_type(second_df, rel.getExpectedInputRowType(1))
```
nils-braun, line 55: There is now a lot of code duplication in this and the `LogicalUnion` plugin. I think it would make sense to extract the basic functionalities (the column name cleaning) into a function in `utils.py` and then reuse it here - or what do you think @demianw?

dask_sql/physical/rel/logical/minus.py
```diff
@@ -0,0 +1,69 @@
...
+
+        df = first_df.merge(
+            second_df,
+            how="left",
+            indicator=True,
+        )
+
+        df = df[df.iloc[:, -1] == "left_only"].iloc[:, :-1]
```
nils-braun, line 63: That is super cool!

tests/integration/test_except.py
```diff
@@ -0,0 +1,29 @@
...
+            UNION
+            SELECT 3 as "a"
+        )
+        EXCEPT
+        SELECT 2 as "a"
+        """
+    )
+    result_df = result_df.compute()
+    assert result_df.columns == "a"
+    assert set(result_df["a"]) == set([1, 3])
```
nils-braun, line 29: Would you mind adding a test with NaNs? You can also use one of the prepared tables (e.g. `user_table_nan`) if it makes sense. 
It might also make sense to test this functionality against sqlite - I am just scared that especially on NULL (NaN) pandas/dask and SQL have different opinions (as I have seen so often, unfortunately...)

tests/integration/test_except.py
```diff
@@ -0,0 +1,29 @@
+def test_except_empty(c, df):
```
nils-braun, line 1: I think you do not need the `df` parameter here

nils-braun (commented):
Thank you so much @demianw! That is a very nice addon! 
If you want, you can also add it into the documentation in `select.rst` (e.g. below the UNION).

I have a few comments on the changes, but in general I am already quite fine with it.

nils-braun:
Oh, and if you want to get rid of the style errors: make sure you use black version 19.10 (I have seen that that might make some difference).

nils-braun:
@demianw - would you still like to work on the PR? I think your changes are absolutely worth being included and just need some small amount of tweaking!

PR #137, opened 2021-02-10
Docker image improvements

nils-braun:
Publish two images:
* the main image
* and one which is prepared to use it on the cloud (e.g. with Faregate)


nils-braun:
Test failure ja unrelated

Issue #83, opened 2020-11-20
connect to databricks cluster


raybellwaves:
I'm taking a look at https://dask-sql.readthedocs.io/en/latest/pages/data_input.html#load-hive-data and i'm curious if a variant exists to connect to a databricks cluster (https://docs.microsoft.com/en-us/azure/databricks/integrations/bi/jdbc-odbc-bi) to extract tables/info from the Hive metastore.

https://stackoverflow.com/a/61806107/6046019 may also be relevant

nils-braun:
That is a very interesting issue. Do you have - by any chance - a working databricks setup where you would be able to test this?

To be honest, I have never worked with Azure Databricks before, so I do not have any experience. However, I have found [this](https://kb.databricks.com/python/sql-in-python.html). In the code, there is the creation of a hive cursor described. This is also exactly what it needs as input to the `create_table` function. 

Concerning the stackoverflow link you shared: so far, I have always tested with a `pyhive.Cursor`. However, there is no reason it should not also work with a sqlalchemy connection. I have added a PR to also allow for these. 

So in my test setup with a "real" hive, both these code blocks work now:

```python
import sqlalchemy

cursor = sqlalchemy.create_engine("hive://localhost:10000").connect()

from dask_sql import Context

c = Context()
c.create_table("hh", cursor)

print(c.sql("SELECT * FROM hh").compute())
```

```
from pyhive.hive import connect

cursor = connect("localhost", 10000).cursor()

from dask_sql import Context

c = Context()
c.create_table("hh", cursor)

print(c.sql("SELECT * FROM hh").compute())
```

Would be super interesting, if this also works with the databricks cluster (looking at the code in the link I shared, I guess the `sqlalchemy`  solution you referenced is much simpler).
If you do not have the chance to test it, I can also try to set up a cluster on my own.

raybellwaves:
Just making some notes here

Some common terms:

token = TOKEN # "xxxx" see https://docs.microsoft.com/en-us/azure/databricks/dev-tools/api/latest/authentication#token-management
auth = "token:{}".format(token)
host = HOST # "adb-x.x.azuredatabricks.net
port = 443
database = DATABASE # "tmp"
http_path = "sql/protocolv1/o/x1/x2"
workspace_url = "x" # Name of the databricks instance
workspace_id = "x1" # x1 in http_path
cluster_id = "x2" # x2 in http_path

uri = "hive://token:{}@{}:{}/{}".format(token, host, port, database)

[...]

PR #140, opened 2021-02-21
Fix integration with dbfs using the newest fsspec version

nils-braun:
I finally found some time to debug the dbfs (databricks) integration mentioned in #83. After the integration of dbfs into fsspec (which is, so far, not released - so you need the HEAD version to try this) I could follow the descriptions by @raybellwaves in the issue:
* setup a databricks cluster on Azure (the community version does not work)
* then execute the lines under "Steps to reproduce" from [here](https://github.com/nils-braun/dask-sql/issues/83#issuecomment-736968990) (thanks for the nice intro, Ray!)
* With the fix from this PR, I am now able to execute this:
```python
from sqlalchemy import create_engine
from dask_sql import Context

dbfs_engine = create_engine(f"databricks+pyhive://token:{token}@{host}:{port}/", 
                            connect_args={"http_path": http_path}
                           ).connect()

c = Context()
c.create_table("df", cursor, hive_table_name="tmp.df", storage_options={"instance": host, "token": token})
```
and it works :-)

### DIFF
--- a/dask_sql/input_utils/hive.py
+++ b/dask_sql/input_utils/hive.py
@@ -89,7 +89,15 @@ class HiveInputPlugin(BaseInputPlugin):
             raise AttributeError(f"Do not understand hive's table format {format}")
 
         def _normalize(loc):
-            return os.path.join(loc.lstrip("file:"), "[A-Za-z0-9-]*")
+            if loc.startswith("dbfs:/") and not loc.startswith("dbfs://"):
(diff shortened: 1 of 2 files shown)


raybellwaves:
assume this works ok/the same now with latest fsspec

nils-braun:
Yes, just tested it :-)
I will add a notebook to describe how to use it and some documentation.

PR #146, opened 2021-03-08
IPython Magic

nils-braun:
This PR allows to register a new ipython magic function in jupyter notebooks and write cells like

```
%%sql
SELECT * FROM df
```

where df could just be a dataframe variable in the notebook.

I took some ideas from the fugue project for that.


nils-braun:
Build failure addressed in #147

PR #160, opened 2021-04-27
upgrade fugue dependency

goodwanghan:
Integration of newer fugue version


nils-braun:
Looks good to me, thanks @goodwanghan !

Issue #163, opened 2021-05-01
OVER Aggregate functions get replace by Calcite


nils-braun:
When writing SQL queries such as `SELECT SUM(a) OVER(...)` or `SELECT AVG(x) OVER(...)`, Apache Calcute automatically replaces the aggregation functions with "more basic" ones, like avg -> sum / count and sum -> case(count > 0, $sum0, NULL).

That makes sense in most of the cases, but Dask can handle avg, sum etc. quite well (and actually the Calcute implementation is even slower). Therefore I would like to control how this replacement is done.

For normal groupby aggregations, there is a OptimizationRule to tune for that. However for the window calls (over), this is currently hardcoded in the `SqlToRelConverter` and in the `AvgVarianceConvertlet`.

Issue #43, opened 2020-09-24
Implement windowing functions


nils-braun:
So far, aggregation can only work without windowing (without the `OVER` keyword).

Unfortunately, I do (still) not know how to implement the very broad windowing techniques from SQL in dask.

beckernick:
There's been a good chunk of effort put into an efficient hash-partitioning shuffle in Dask this year. Perhaps this could be leveraged here?

I could imagine parsing the query plan and mapping the `func() OVER (PARTITION BY hash_cols ORDER BY col) AS ...` to a custom task composed of several standard Dask operations. With the hash repartition, all unique rows corresponding to specific hash values in `col` would be wholly contained in the same partition, somewhat equivalent to if we collected each individual PARTITION BY partition into larger sets of partitions. This should allow for within-group operations for the functional logic, assuming we preserve ordering of rows within groups (implicitly or explicitly, depending on the groupby).

As a small example:

```python
import pandas as pd
from pyspark import SparkContext
from pyspark.sql import SQLContext, SparkSession
from pyspark.sql import functions as F
import dask.dataframe as dd
import dask
from distributed import Client, LocalCluster, wait
from dask.datasets import timeseries

spark = SparkSession.builder \
    .master("local") \
    .getOrCreate()

time_df = timeseries(
    start='2000-01-01',
    end='2000-01-31',
    freq='30s'
)
time_df = time_df.persist()
time_pdf = time_df.compute()

sdf = spark.createDataFrame(time_pdf)
sdf.createOrReplaceTempView("time_df")
sdf.cache()
```

Rank (dense_rank in this case)
```python
q = """
SELECT
    id,
    name,
    x,
    y,
    CAST(dense_rank() OVER (PARTITION BY name ORDER BY y ASC) AS DOUBLE) as rank
FROM
    time_df
"""


def rankfunc(df, keys, col, ascending):
    keys = [keys] if isinstance(keys, str) else keys
    col = [col] if isinstance(col, str) else col

    new = df.copy()
    new['rank'] = new.groupby(keys)[col].rank(method="dense", ascending=ascending)
    
    order_by_col_order = [True] * len(keys) + [ascending]
    new = new.sort_values(keys + col, ascending=order_by_col_order).reset_index(drop=True)
    return new
    
def sql_window_rank(df, partition_cols, order_by_col, ascending=False):
    new = df.copy()
    new = new.shuffle(on=partition_cols)
[...]

PR #164, opened 2021-05-01
Implement `OVER` for arbitrary windows

nils-braun:
This is another step towards #43.

It introduces handling for arbitrary window definitions, including `UNBOUNDED`, `CURRENT ROW` and fixed offsets.
The code is also using the pandas rolling functions now if possible, which should speed up the calculation.

Still, it is super slow due to multiple reasons (one being #163) - but this is something we should fix in the upcoming development.

### DIFF
--- a/.github/docker-compose.yaml
+++ b/.github/docker-compose.yaml
@@ -1,0 +1,17 @@
+# Docker-compose setup used during tests
+version: '3'
+services:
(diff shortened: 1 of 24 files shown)


nils-braun:
@goodwanghan - if you maybe find some time, would you like to give it a try?

PR #180, opened 2021-05-19
Add better partition string escaping

gallamine:
Based on the https://github.com/nils-braun/dask-sql/issues/179 this PR adds string escaping to non-integer partition values. Partitions like:

```
forecasted_business_day=2021-05-27
```

will now be wrapped like 
```
forecasted_business_day="2021-05-27"
```
when executing the SQL

### DIFF
--- a/dask_sql/input_utils/hive.py
+++ b/dask_sql/input_utils/hive.py
@@ -1,6 +1,7 @@
 import ast
 import logging
 import os
+import re
 from functools import partial
 from typing import Any, Union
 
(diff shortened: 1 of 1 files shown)


nils-braun:
Thank you so much @gallamine! I have added some thoughts to https://github.com/gallamine/dask-sql/pull/1 on that and #181. I think we need to solve both issues at the same time.

GPUtester:

Admins can comment `ok to test` to allow this one PR to run or `add to allowlist` to allow all future PRs from the same author to run.

PR #184, opened 2021-05-20
Sarg and search

### LINKED ISSUE #175: ValueError: invalid literal for int() with base 10: 'Sarg[(0..1)]'
I'm experiencing issues when using `CASE` and checking whether a value is in a given range. The issue seems to occur when using `BETWEEN` as well as a manual range select.
In any case the execution fails with `ValueError: invalid literal for int() with base 10: 'Sarg[(0..1)]'`.
You can reproduce the error when using the code snippet below. Thanks!

```python
import dask
import dask_sql

import pandas as pd

df = pd.DataFrame({"A": [4.1] * 5})
data = dask.dataframe.from_pandas(df, npartitions=1)

sql_context = dask_sql.Context()
sql_context.create_table("data", data)

result = sql_context.sql(
    """
        SELECT
            CASE
                WHEN data.A > 0 AND data.A < 1 THEN 'in-between'
                ELSE 'out-of-range'
            END AS is_in_range
        FROM data
        """,
    return_futures=False,
)

print(result)
```

**Execution Environment**
Python 3.8.10
dask==2021.4.1
dask-sql==0.3.4

nils-braun:
Fixes #175 

`SEARCH` is a not-so-frequent SQL operation, but Calcite uses it internally for optimizing calls such as shown in the issue.
This PR implements in implementation for the search operation as well as the Search Args (Sargs).

I tried to cover some edge cases like infinity, but I am not sure if I caught all of them.

### DIFF
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -17,6 +17,7 @@ from dask_sql.datacontainer import DataContainer
 from dask_sql.mappings import cast_column_to_type, sql_to_python_type
 from dask_sql.physical.rex import RexConverter
 from dask_sql.physical.rex.base import BaseRexPlugin
+from dask_sql.physical.rex.core.literal import SargPythonImplementation
 from dask_sql.utils import (
     LoggableDataFrame,
     convert_to_datetime,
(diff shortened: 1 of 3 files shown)


nils-braun:
@mschroederi, would you like to have a look or shall I merge?

mschroederi:
lgtm

mschroederi:
@nils-braun Thanks!

PR #185, opened 2021-05-24
ML model improvement : Added SHOW MODELS and DESCRIBE MODEL

rajagurunath:
ML Model Improvement 

1.SHOW MODELS
    List down the registered model  in the context
2. DESCRIBE MODEL <modelName>
      provides hyper parameters used to train a model


romainr:
Great functionalities!

notebooks/Feature Overview.ipynb
```diff
@@ -571,13 +601,20 @@
...
+   "execution_count": null,
+   "metadata": {},
+   "outputs": [],
+   "source": []
   }
  ],
  "metadata": {
   "kernelspec": {
-   "display_name": "Python 3",
+   "display_name": "dask-sql",
```
nils-braun, line 615: I think it is probably more general to keep the original "Python" kernel name, as not every user might have named their kernel "dask-sql" (and it will also work better with Binder).

planner/src/main/codegen/includes/show.ftl
```diff
@@ -46,21 +46,33 @@ SqlNode SqlShowColumns() :
     }
 }
 
-// DESCRIBE "table"
+// DESCRIBE TABLE "table"
```
nils-braun, line 49: While I generally like to be as specific as I can, I am a bit worried because `DESCRIBE <table>`  is actually a well-known SQL command (see e.g. [here](https://www.geeksforgeeks.org/sql-describe-statement/#:~:text=So%20desc%20or%20describe%20command,time%20of%20Creation%20of%20table.)). How about allowong both `DESCRIBE TABLE <table>` and `DESCRIBE <table>` (and also `DESCRIBE MODEL <model>`)? You can use a or-clause (with a `|`) to distringuish the different cases.
(Feel free to ask me, if that was not clear enough).

planner/src/main/codegen/includes/show.ftl
```diff
@@ -46,21 +46,33 @@ SqlNode SqlShowColumns() :
...
-
+// DESCRIBE MODEL "model_name"
+SqlNode SqlDescribeModel() :
+{
+    final Span s;
+    final SqlIdentifier modelName;
+}
+{
+    <DESCRIBE> { s = span(); } <MODEL>
+    modelName = CompoundTableIdentifier()
```
nils-braun, line 71: You can replace the `<MODEL> modelName` with `ModelIdentifier`, like e.g. in `SqlPredictModel` in `model.ftl`. With that, you can also use code similar as in https://github.com/nils-braun/dask-sql/blob/main/dask_sql/physical/rel/custom/predict.py#L52 in your python code without the need for a custom function. This is not a must and your code will also work, but I think it might be a bit inconsistent between a call to `PREDICT` and a call to `DESCRIBE MODEL` - as one of them will allow for a compound identifier (which is not used anyway) and the other will not. Also, it will probably make the custom function `get_model_from_compound_identifier` unnecessary (or there is another reason for having a compound identifier which I miss?)

planner/src/main/java/com/dask/sql/parser/SqlShowModels.java
```diff
@@ -0,0 +1,37 @@
...
+import org.apache.calcite.sql.SqlCall;
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.SqlNode;
+
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+public class SqlShowModels extends SqlCall {
+    public SqlIdentifier catalog;
```
nils-braun, line 15: You do not need those two variables.

tests/integration/test_model.py
```diff
@@ -250,3 +296,55 @@ def test_drop_model(c, training_df):
...
+            wrap_predict = True,
+            target_column = 'target'
+        ) AS (
+            SELECT x, y, x*y > 0 AS target
+            FROM timeseries
+            LIMIT 100
+        )
+    """
+    )
+    expected_dict = {
```
nils-braun, line 315: Would it make sense to get those parameters dynamically from an instance of the model (e.g. created with `sklearn.ensemble.GradientBoostingClassifier()` directly)? I am a bit worried that we will need to change this code often whenever there is a new version of `sklearn`.

nils-braun (commented):
Hi @rajagurunath - thanks again for your work and your really nice PR.
Sorry, it took a bit longer than anticipated...

I had some smaller comments (with are mostly irrelevant), but the two important comments are the one on the `DESCRIBE TABLE` and the `SqlModelIdentifier`, which will probably improve readability and maintainability.
Please feel free to get in touch with me if the comments were not clear enough.

planner/src/main/codegen/includes/show.ftl
nils-braun: And it will automatically fix the missing coverage, as the function will be gone then ;-)

notebooks/Feature Overview.ipynb
rajagurunath: sure, makes sense,  I will revert this changes back to previous config

planner/src/main/codegen/includes/show.ftl
rajagurunath: I may want your input here @nils-braun , Actually I tried with `DESCRIBE TABLE <table_name>` and `DESCRIBE <table_name>` , But I think  parser on seeing "DESCRIBE" it automatically redirects to `ShowColumnsPlugin ` so `DESCRIBE MODEL <model_name>` was parsed incorrectly that's y added TABLE keyword to distinguish between this two situation , please let me know if some other things I can try to rectify this situation .
Please find below screenshot for further information.
describeTable
MODELkeyword-interpretedasTableName
ThisErrorisExpected

planner/src/main/codegen/includes/show.ftl
rajagurunath: GOT it, makes perfect sense, I have replaced type of modelName as modelIdentifier and made changes in python code as well , but currently blocked by above parsing issue .

planner/src/main/java/com/dask/sql/parser/SqlShowModels.java
rajagurunath: I will remove this variables , like and catalog

tests/integration/test_model.py
rajagurunath: Sure, this is good idea , I will get model from the context and the model parameters as expected data .And compare it with result of the Describe model output

planner/src/main/codegen/includes/show.ftl
nils-braun: Sure! Let me try to give you some more context. 
The `ftl` files are Java freemarker template files. They are turned into a parsing logic by a maven plugin, which is then integrated into the jar. There exist multiple ways how to specify the grammar you want to describe, and one is also the "alternative". It looks similar to this:
```freemarker
SqlNode Something() :
{
    ....
}
{
    <DESCRIBE>
    (
        tableName = CompoundTableIdentifier()
        {
            return ...
        }
    |
        <MODEL>
        modelIdentifier = ...
        {
            return ...
         }
    )
}
```
With that, we define two possibilities: `DESCRIBE "table"` and `DESCRIBE MODEL "model"`, which could also return different Java classes (and they should :-)).
 It would even be possible to include an optional `[ <TABLE> ]`, to also allow for a `DESCRIBE TABLE "table"`.

One thing to note when writing these alternatives (or branches), is that the parser will decide on the very first token, which route to take. This means, if you have two alternatives `A B C` and `A B D`, it can not decide after `A` already. One way to circumvent this, is to use a `LOOKAHEAD(2)` command, which will tell the parser to decide after two tokens.

So to summarize: you need to combine the model and table `DESCRIBE` statements into a single function, which will look similar to this:
```freemarker
SqlNode SqlDescribeTableOrModel() :
{
    final Span s;
    final SqlIdentifier schemaName;
    final SqlIdentifier tableName;
    final SqlModelIdentifier modelName;
}
{
    <DESCRIBE> { s = span(); }
    (
        LOOKAHEAD(2)
        modelName = ModelIdentifier()
        {
            return new SqlShowModelParams(s.end(this), modelName);
        }
    |
        [ <TABLE> ]
        tableName = CompoundTableIdentifier()
        {
            return new SqlShowColumns(s.end(this), tableName);
        }
    )
}
```
(please note that I did not test the code, so there might be some syntax errors)

planner/src/main/codegen/includes/show.ftl
nils-braun: Great!

planner/src/main/java/com/dask/sql/parser/SqlShowModels.java
nils-braun: Thanks

tests/integration/test_model.py
nils-braun: Makes sense!

planner/src/main/codegen/includes/show.ftl
nils-braun: Please also not the order of the two branches: first check the `MODEL` alternative.

rajagurunath: wow, Thanks for the great explanation, Trying this now , will let you know shortly .

rajagurunath:
Hi @nils-braun , Thanks for the inputs/guidance, Please review the PR, let me know if there is any further corrections . 
Thanks

nils-braun (approved):
I just added a very tiny change to also test the exception. The rest was very well done! lgtm

PR #193, opened 2021-06-17
Feature/regr count

rajagurunath:
Added REGR_COUNT:

Calcite makes use of regr_count and implements the following :
1. REGR_SXX
2. REGR_SYY
3. COVAR_POP
4. COVAR_SAMP


dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -109,6 +110,9 @@ class LogicalAggregatePlugin(BaseRelPlugin):
         "max": AggregationSpecification("max", AggregationOnPandas("max")),
         "min": AggregationSpecification("min", AggregationOnPandas("min")),
         "single_value": AggregationSpecification("first"),
+        "regr_count": AggregationSpecification(
```
nils-braun, line 113: This is just a normal "count", isn't it?
In this case, I would prefer to have 
```suggestion
        "regr_count": AggregationSpecification("count", AggregationOnPandas("count"))
```
as it might be faster than a "handmade" implementation. Or did I miss something?

Dask might be able to use some nice cython-optimized implementations on built-in methods. That might also be the case in your definition, but I would rely on the Dask developers to probably have a better importation here :-)

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -243,13 +249,25 @@ def _collect_aggregations(
...
+            elif aggregation_name == "regr_count":
+                is_null = IsNullOperation()
+                two_columns_proxy = new_temporary_column(df)
+                col1 = cc.get_backend_by_frontend_index(inputs[0])
+                col2 = cc.get_backend_by_frontend_index(inputs[1])
+                # both cols should be not null
+                df = df.assign(
+                    **{two_columns_proxy: (~is_null(df[col1]) & (~is_null(df[col2])))}
+                )
+                input_col = two_columns_proxy
```
nils-braun, line 270: Nice, looks really good to me

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -290,8 +305,8 @@ def _collect_aggregations(
             value = (input_col, output_col, aggregation_function)
             collected_aggregations[key].append(value)
             output_column_order.append(output_col)
-
-        return collected_aggregations, output_column_order
+        # mutated df - dono how to avoid this !!!
```
nils-braun, line 308: You can remove this comment. It is totally fine to return it here and dask-sql does this all the time. Your solution is ok.

dask_sql/physical/rel/logical/aggregate.py
nils-braun: And, one additional thought: can you include a comment why you can use a count here (basically, because you do the pre-transformation earlier)?

tests/integration/fixtures.py
```diff
@@ -5,10 +5,19 @@
 import numpy as np
 import pandas as pd
 import pytest
+from dask.datasets import timeseries
 from dask.distributed import Client
 from pandas.testing import assert_frame_equal
 
 
+@pytest.fixture()
+def training_df(c):
```
nils-braun, line 14: Good idea! But maybe a more telling name would be good? Something like "timeseries_df"?

tests/integration/test_groupby.py
```diff
@@ -238,3 +238,84 @@ def test_aggregations(c):
 
     expected_df = pd.DataFrame({"max": ["a normal string"], "min": ["%_%"]})
     assert_frame_equal(df.reset_index(drop=True), expected_df)
+
```
nils-braun, line 241: I like your tests, they are all very good.
However you could even increase the test cases for your new cool feature by adding a test in the comparison with SQL lite and/or postgreSQL. There are already tests for this and you probably only need to add one or two lines of SQL that call your new function (s). The benefit of having this additional tests is, that by now you only test the "definition". If we have misunderstood the definition of regr_count (for whatever reason and it is not count(y and x not null)), we would not see it here.

tests/integration/fixtures.py
```diff
@@ -5,10 +5,19 @@
...
+    df = timeseries(freq="1d").reset_index(drop=True)
+    c.create_table("timeseries", df, persist=True)
```
nils-braun, line 16: The only problem with this df is, it has no NULL values - and as regr_count only gives non-trivial results in this case...
Could you introduce some NaN values (maybe even randomly)? Maybe 10-30% of the values (independently in x and y, so that we have pairs of both null and only one null and none null)

nils-braun (commented):
Again, very good work!
I have (as always ;-)) some comments.

dask_sql/physical/rel/logical/aggregate.py
rajagurunath: Yes @nils-braun, I have tried that first, like using `AggregationSpecification("count", AggregationOnPandas("count"))` 
but got following error (refer below screenshot)
image

This Error was raised when calling **regr_sxx,regr_syy** functions, As you already know our aggregation_dict looks like this 
```
# we are preparing this aggregation dict here in our code 
for aggregation in aggregations:
    input_col, output_col, aggregation_f = aggregation
```

And Dask expects (aggregation_f, input_column) combination to be unique in the final dask graph plan, it makes sense because dask doesn't need to repeat the operation twice, And  our aggregation function produces (count,sepal_length ) twice this is why we are getting this error (for more reference into dask code base refer [here](https://github.com/dask/dask/blob/8aea537d925b794a94f828d35211a5da05ad9dce/dask/dataframe/groupby.py#L689)

To understand why  (count,sepal_length) was produced twice please refer to the below plan produced by the calcite 
image

And since both normal count and regr_count using the same name **count** on `AggregationSpecification `, dask not allowing this duplicate combination I guess, and raising the error.  that's why I differentiated these two functions with a different name (count and regr_count). Do you know any other way to avoid this duplicate combination, may be calcite plan itself can avoid duplicate combinations by adding any optimization rule?  pleased to know your thoughts on this.

dask_sql/physical/rel/logical/aggregate.py
rajagurunath: ❤️

dask_sql/physical/rel/logical/aggregate.py
rajagurunath: 👍 got it

tests/integration/fixtures.py
rajagurunath: sure 👍

tests/integration/test_groupby.py
rajagurunath: Thanks for the explanation,  💯 makes sense !!

Added following function only in test_postgres.py not added in test_sqlite.py (not sure whether the SQLite supports those function)

```
def test_statistical_functions(assert_query_gives_same_result):

    # test regr_count
    assert_query_gives_same_result(
        """
        select user_id,regr_count(a,b) from df1 group by user_id
        """
    )

    assert_query_gives_same_result(
        """
        select user_id, regr_sxx(a,b) from df1 group by user_id
        """
    )

    assert_query_gives_same_result(
        """
        select user_id, regr_syy(a,b) from df1 group by user_id
        """
    )

    assert_query_gives_same_result(
        """
        select user_id,COVAR_POP(a,b) from df1 group by user_id
        """
    )

    assert_query_gives_same_result(
        """
        select user_id,COVAR_SAMP(a,b) from df1 group by user_id
        """
    )

```
Is this something you are expecting?

tests/integration/fixtures.py
rajagurunath: makes perfect sense, I will introduce some NULL values in timeseries_df

dask_sql/physical/rel/logical/aggregate.py
nils-braun: Ah thanks, very detailed description.
However, I still think that using count would be a good idea and I actually think that this is a bug, because regr_count should not be applied on the bare column, but on your temporary column (created after applying not is nan). In your code, you are only doing the pre-processing if we have two inputs, but in the code example you are showing there is actually only one input to "regr_count". However, to be consistent, we should also create a temporary column with the not is nan pre-processing in this case. I guess all you need to do is change the if-elif logic.

rajagurunath: Thanks a lot for pointing out the bug and clearly explaining why it was Occurring, I think I have misunderstood the underlying problem now I am clear and able to implement this functionality with inbuilt dask functions instead of using custom functions, currently, I am using the following specification 

```
# is null was checked earlier, now only need to compute the sum of the non-null values
"regr_count": AggregationSpecification("sum", AggregationOnPandas("sum"))
```
if we are using built-in methods, I guess I need to change it into **sum** instead of **count** because, a sum only sums the boolean values (only True values) of the temporary column, and count counts both True and False

For regr_count function above implementation works and matches with the Postgres result.

rajagurunath: But After testing with nan introduced data frame, `regr_sxx` and `regr_syy` results are not exactly matching with Postgre's results and the testing formula (regr_count (y,x) * var_pop (y)), but matching with the relative tolerance to set to 1 in `assert_frame_equal`, is this okay to set this parameter, Dono Why I am getting that difference. Any idea here?

rajagurunath:
Hi @nils-braun,

Not able to find why test cases are getting dask client instead of dask_sql.context.Context in place "**c**", that is the reason all the test cases are failing I think,  Any pointers where I am going wrong?

nils-braun:
Hi @rajagurunath 
oh, that one was nasty! The bug was totally unrelated to your changes. It just happened to be another `distributed` version, which changed the pytest fixtures of `distributed`. And as we specify the `distributed` fixtures to be included in our tests as well, the `c` fixture was now overwritten by the `distributed` fixture for `c` (which is a dask client). 

There is an easy fix for this, and that is changing the order of the fixtures: first, the distributed one, then ours - so the `c` will be defined by `dask-sql`. Don't know why I haven't done so already... I will do a PR to fix this.

nils-braun:
Hi @rajagurunath 
oh, that one was nasty! The bug was totally unrelated to your changes. It just happened to be another `distributed` version, which changed the pytest fixtures of `distributed`. And as we specify the `distributed` fixtures to be included in our tests as well, the `c` fixture was now overwritten by the `distributed` fixture for `c` (which is a dask client). 

There is an easy fix for this, and that is changing the order of the fixtures: first, the distributed one, then ours - so the `c` will be defined by `dask-sql`. Don't know why I haven't done so already... I will do a PR to fix this.

nils-braun:
I uploaded the fixes you did to the mlflow test and the change in the fixture order that I have implemented. I think that now all tests should be green - but let's wait for it.

nils-braun:
I was able to remove the `rtol=1`:
* it seems postgres and calcite handle the return type of the `REGR_SXX` (and yy) calls differently: postgres always returns a float, whereas calcite returns an integer if one of the inputs is an integer. Therefore, I am now converting to floats in both cases (leading to the same result)
* in the tests where you tested for the definition of `REGR_SXX`, you allowed for all cases. However, the formular for `REGR_SXX` is only true, when both `x` and `y` are not-null. Now the tests pass.

rajagurunath:
Amazing ❤️ ,  Thanks again for finding this Bug and Correcting it !!.  Perfectly makes sense now!!!!

please let me know if something is pending in this functionality, like adding the functions in documentation, etc. Happy to take that !.

nils-braun:
Right, food you mentioned that! Sorry for not providing this info before.

It would be nice to put the new SQL commands into the SQL documentation, just where all the other aggregation functions are mentioned. After you have done that, we can merge the PR!

rajagurunath:
Added in the documentation , Please let me know if any other things need to be added

nils-braun:
Looks really good!

PR #194, opened 2021-06-20
Fix test fixture order

nils-braun:
First import external fixtures, then ours, to make sure we overwrite any other definitions.
See https://github.com/nils-braun/dask-sql/pull/193#issuecomment-864606251 for more explanation.

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,3 @@
 import pytest
 
-pytest_plugins = ["tests.integration.fixtures", "distributed.utils_test"]
+pytest_plugins = ["distributed.utils_test", "tests.integration.fixtures"]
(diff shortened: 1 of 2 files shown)


rajagurunath:
Hi @nils-braun, out of two test cases that are failing, I think I kind of know the reason why the MLflow model export was failing. 
MLflow's latest release includes logging/saving pip dependency (requirements.txt) as well in addition to the conda.yaml 
https://github.com/mlflow/mlflow/pull/4409

I think there is no need to assert for number of files generated while exporting the model into MLflow model format

I am guessing the second issue is related to the pyarrow version but not sure about that!

nils-braun:
Good catch. Could you try to fix the mlflow error? I will take care of the pyarrow problem. You can just open a new PR with the target set to this branch.

rajagurunath:
Have raised a PR https://github.com/nils-braun/dask-sql/pull/195 , let me know if this what you are expecting.

PR #199, opened 2021-07-03
Feature/automl

rajagurunath:
This pull request adds a Hyperparameter tuner and automl behavior  as part of https://github.com/nils-braun/dask-sql/issues/130

1. Uses dask_ml for hyperparameter tuning 
2. tpot for automl

(Added dask_ml and tpot in github action for test coverage)

**some known issues so far :**

1. Dependency problem 
    * Hyperparameter tuner requires `dask_ml` (latest) which requires `dask_xgboost` (latest) which inturn requires `xgboost`<=0.90 
    * But `tpot 0.11.7` (latest) requires `xgboost`>=1.1.0 so there is a conflict that breaks some hyperparameter tuning and some model prediction as well, so a temporary solution downgraded the `tpot==0.11.6.post1` which works fine as of now.

Please let me know if there is any better solution to handle this situation .

As always please provide your valuable feedback 😊


rajagurunath:
I think the test cases are failing due to the dtype difference (float64 vs int64) in data frames and I suspect this may be due to the latest dask update 🤔  Any thoughts ?

nils-braun:
Concerning the xgboost dependency, please see [my comment](https://github.com/nils-braun/dask-sql/issues/130#issuecomment-873657281) :-)

Regarding the code! Looks quite nice, I will try to have a detailed look soon.

And to the CI/CD failure: it seems we are often spotting the new "features" of new dask versions :-) Well, that is the fate of having such a large and good pydata ecosystem (there are so many moving parts). I just triggered a re-run of the corresponding test pipeline on our current main branch - I expect it will also fail. In this case we can fix it on main first and then merge back here.

nils-braun:
I did find the issue and have fixed it in #202. Once it is merged, feel free to also include it here. Hope this will help!

rajagurunath:
Hi @nils-braun, I tried using the auto-detect feature, but failed with following error: 

`gaierror: [Errno 8] nodename nor servname provided, or not known`


image

Anything I am missing here?  Please let me know !

nils-braun:
Hi @rajagurunath 
I finally found some time to try it our by myself. I am using slightly different versions of the libraries
(ask = '2021.06.2, xgboost = 1.4.0), but I do not think that matters (see below).
Some things I changed before playing around with the code:
* We need to do one pre-processing step, as the `iris` table contains a string column (and xgboost does not know how to handle this):
```sql
CREATE OR REPLACE TABLE enriched_iris AS (
    SELECT 
        sepal_length, sepal_width, petal_length, petal_width,
        CASE 
            WHEN species = 'setosa' THEN 0 ELSE CASE 
            WHEN species = 'versicolor' THEN 1
            ELSE 2 
        END END AS "species"
    FROM iris 
)
```
* And we should not wrap the predict call, because `xgboost` already knows how to handle a dask dataframe:
```sql
    model_class = 'xgboost.dask.DaskXGBClassifier',
    target_column = 'species'
) AS (
    SELECT * FROM "enriched_iris"
)
```

For me, that will produce a nice xgboost model, that I can also use for prediction:
```sql
SELECT
    *
FROM PREDICT(
    MODEL my_model,
    SELECT * FROM enriched_iris
)
```

Now - why did it work for me but not for you? The error message seems to point to some network problems. I guess the software wants to get the hostname from 127.0.0.1, which should be localhost (or the other way round). Normally, it uses the entry in the hosts file (for me, that is `/etc/hosts`, but I do not know which OS you are working with). I am wondering why it fails to resolve localhost... Could you try to find the hosts-file for your OS and paste the lines related to localhost or 127.0.0.1 here (if any)?

rajagurunath:
Hi @nils-braun 

Thanks a lot for the detailed comments with a code example:

you are right in both cases: 

**Dependency conflict:**
Actually, it's my mistake, dask-xgboost is an optional dependency for dask-ml, I just got confused while executing the model `dask_ml.xgboost.XGBClassifier` in a notebook (which raised ImportError for dask-xgboost, so I installed it and everything that happened later was history 😅 ). TLDR,  there are no dependency conflicts.

**Network Error:**
AS you have suggested I have added one entry in my laptop (macOS)  in `/etc/hosts (127.0.0.1   node_name )` which solved this issue (i.e resolved localhost address). and auto-detect feature detects the existing dask client, like a charm. (but faced some hiccups when there are more than one dask-scheduler is available, the test fails with the status XFails (Worker failed to start)- exploring more on this !).

Once again thanks for your time, kindly review the changes and let me know if I missed something!

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -0,0 +1,207 @@
...
+        )
+        model_class = None
+        automl_class = None
+        experiment_class = None
+        if "model_class" in kwargs:
+            model_class = kwargs.pop("model_class")
+            # when model class was provided, must provide experiment_class also for tuning
+            if "experiment_class" not in kwargs:
+                raise ValueError(
+                    f"Parameters must include a 'experiment_class' parameter for tunning {model_class}."
```
nils-braun, line 122:
```suggestion
                    f"Parameters must include a 'experiment_class' parameter for tuning {model_class}."
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
     -- DESCRIBE MODEL <model_name>.
     DESCRIBE MODEL my_model
 
+5. Hyperparameter Tuning
+-------------------------
+Wants to increase the performance of your model by tuning the
```
nils-braun, line 105:
```suggestion
Want to increase the performance of your model by tuning the
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
+parameters ? Use the Hyperparameter tuning directly
```
nils-braun, line 106:
```suggestion
parameters? Use hyperparameter tuning directly
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
+in SQL using below SQL syntax, choose different tuners
+from dask_ml package based on memory and compute constraints and
```
nils-braun, line 108:
```suggestion
from the dask_ml package based on memory and compute constraints and
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
+for more details refer  `dask ml documentation <https://ml.dask.org/hyper-parameter-search.html#incremental-hyperparameter-optimization>`_
```
nils-braun, line 109:
```suggestion
for more details refer  to the `dask ml documentation <https://ml.dask.org/hyper-parameter-search.html#incremental-hyperparameter-optimization>`_
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
 
-5. Export Trained Model
+.. code-block:: sql
+
+ CREATE EXPERIMENT my_exp WITH(
```
nils-braun, line 113:
```suggestion
 CREATE EXPERIMENT my_exp WITH (
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
+    target_column = 'target'
+    ) AS (
+        SELECT x, y, x*y > 0 AS target
+        FROM timeseries
+        LIMIT 100
+    )
+
+5.1 Automl in SQL
+-----------------
+Wants to try different models with different parameters in SQL ?, Now you can
```
nils-braun, line 129:
```suggestion
Want to try different models with different parameters in SQL? Now you can
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
+start automl experiment with the help ``tpot`` framework which trains
```
nils-braun, line 130:
```suggestion
start automl experiments with the help of the ``tpot`` framework which trains
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
+and evaluates number of different sklearn compatible models and uses dask for
```
nils-braun, line 131:
```suggestion
and evaluates a number of different sklearn compatible models and uses dask for
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
+distributing the work across the dask clusters.
+use below SQL syntax for automl and for more details refer
```
nils-braun, line 133:
```suggestion
Use below SQL syntax for automl and for more details refer to the
```

docs/pages/machine_learning.rst
```diff
@@ -100,8 +100,62 @@ following sql statements
...
+            max_eval_time_mins=1),
+            target_column = 'target'
+            ) AS (
+                SELECT x, y, x*y > 0 AS target
+                FROM timeseries
+                LIMIT 100
+            )
+
+After the experiment was completed, both hyperparameter tuner and
+automl experiments stores the best model of the experiment in sql context
```
nils-braun, line 155:
```suggestion
automl experiments stores the best model of the experiment in the sql context
```

planner/src/main/java/com/dask/sql/parser/SqlCreateExperiment.java
```diff
@@ -0,0 +1,69 @@
...
+import org.apache.calcite.sql.SqlIdentifier;
+import org.apache.calcite.sql.SqlKind;
+import org.apache.calcite.sql.SqlNode;
+import org.apache.calcite.sql.SqlOperator;
+import org.apache.calcite.sql.SqlSpecialOperator;
+import org.apache.calcite.sql.SqlWriter;
+import org.apache.calcite.sql.parser.SqlParserPos;
+
+public class SqlCreateExperiment extends SqlCreate {
+    private static final SqlOperator OPERATOR = new SqlSpecialOperator("CREATE MODEL", SqlKind.OTHER_DDL);
```
nils-braun, line 17:
```suggestion
    private static final SqlOperator OPERATOR = new SqlSpecialOperator("CREATE EXPERIMENT", SqlKind.OTHER_DDL);
```

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -0,0 +1,207 @@
...
+from dask_sql.java import org
+from dask_sql.physical.rel.base import BaseRelPlugin
+from dask_sql.utils import convert_sql_kwargs, import_class
+
+logger = logging.getLogger(__name__)
+
+
+class CreateExperimentPlugin(BaseRelPlugin):
+    """
+    Creates /initiate  Experiment for hyperparameter tuning or Automl like behaviour,
```
nils-braun, line 17:
```suggestion
    Creates an experiment for hyperparameter tuning or automl like behaviour,
```

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -0,0 +1,207 @@
...
+    i.e evaluates model with different hyperparameters and registers the best performing
```
nils-braun, line 18:
```suggestion
    i.e evaluates models with different hyperparameters and registers the best performing
```

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -0,0 +1,207 @@
...
+    model in the context which can be used for prediction
+
+    sql syntax:
+        CREATE EXPERIMENT <> WITH ( key = value )
```
nils-braun, line 22:
```suggestion
        CREATE EXPERIMENT <name> WITH ( key = value )
```

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -0,0 +1,207 @@
...
+
+from dask_sql.datacontainer import ColumnContainer, DataContainer
+from dask_sql.java import org
+from dask_sql.physical.rel.base import BaseRelPlugin
+from dask_sql.utils import convert_sql_kwargs, import_class
+
+logger = logging.getLogger(__name__)
+
+
+class CreateExperimentPlugin(BaseRelPlugin):
```
nils-braun, line 15: General comment (applies to this class and all the examples/documentation below): the model name is currently independent of the experiment name. Would you think it makes sense to change that, so that the best performing model has the same name as the experiment? From my view, that would make it easier for people to do the connection. Because now I fear that having multiple experiments will override the output model.
In any case, we should mention in the documentation somewhere how the resulting model is named.

nils-braun (approved):
Very good work. I have just found some small typos here and there that you can fix right from the GitHub web interface.
The only real comment I had was on the model name, but apart from that this PR is ready to be merged!

dask_sql/physical/rel/custom/create_experiment.py
rajagurunath: Hi @nils-braun, Thanks for taking the time to review the code and clearly explaining the bugs and typos.

I also like the idea of storing the best-performing model with the experiment name. and added the same in the documentation.

Please review and let me know if any further changes are needed

nils-braun:
That's in! Another very nice addition to the ML functionality

PR #205, opened 2021-08-04
Multiple schemas allowed

### LINKED ISSUE #203: Allow for multiple schemas
Currently, all tables, functions and other context variables are supposed to belong to the single schema "schema".
In principle, both `Calcite` and the SQL parser statements allows for multiple schemas, we would just need to implement:

- [ ] dictionaries instead of lists for functions, tables and models in the context
- [ ] passing the schema information on into the `RelationalAlgebraGenerator` object.
- [ ] statements for creating a schema, deleting it (with cascade) and switching the root schema (= the schema which is inherently used when no schema is defined).

nils-braun:
Fixes #203 

Most of the hard work of the implementation was done by @rajagurunath - I am just finishing up.
This PR allows to have multiple schemas defined and use them as usual: there is a `USE SCHEMA name` SQL command to select a root-schema, but you can also reference tables, functions, models or experiments with a fully qualified name like

```sql
SELECT * FROM schema.table
```

Technically, this is implemented by storing a dictionary for tables/functions/etc keyed by the schema name. For calcite, we pass in a list of schema definitions (via a builder pattern).

Still missing: some documentation on the new features


nils-braun:
@rajagurunath - in the end, that was more complicated than I thought so it took a bit longer... Please have a look if you have some time. 

I moved back to an implementation without a `add` function. The reason is that I just have a lot more control over where it is added (tables/functions) etc. than a generic `add` function. I still like the idea, but I am a bit scared that the dispatch might not do what we want in all cases.

rajagurunath:
> @rajagurunath - in the end, that was more complicated than I thought so it took a bit longer... Please have a look if you have some time.
> 
> I moved back to an implementation without a `add` function. The reason is that I just have a lot more control over where it is added (tables/functions) etc. than a generic `add` function. I still like the idea, but I am a bit scared that the dispatch might not do what we want in all cases.

Wow, Really the code looks clean and to the point as always ❤️. 

Indeed it's Quite a lot of work,  you have done the real heavy lifting 🥇 , yeah I just learned and tried the function dispatch, but your concern really makes sense. Thanks for pitching in sailing the ship in the right direction as always 😄

rajagurunath (approved):
Looks Really Great ❤️

nils-braun:
Thanks for the review!

PR #207, opened 2021-08-05
Updates to `aggregate.py` and `join.py`

sarahyurick:
Two files have been modified here: 

1. In `aggregate.py`, I changed `_perform_aggregation()` to construct a dict-of-lists, rather than a dict-of-dicts. Although Dask still supports the ability to pass a dict-of-dicts into `groupby().agg()`, Pandas previously deprecated and has now removed this functionality (see here: https://pandas.pydata.org/pandas-docs/stable/whatsnew/v0.20.0.html#deprecate-groupby-agg-with-a-dictionary-when-renaming). 
2. In `join.py`, there were a couple function calls that were Pandas-specific, so I changed them to call to `dask.dataframe` instead.

Let me know what you think!

### DIFF
--- a/dask_sql/physical/rel/logical/aggregate.py
+++ b/dask_sql/physical/rel/logical/aggregate.py
@@ -1,3 +1,24 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2021, NVIDIA CORPORATION.
+#
+# Permission is hereby granted, free of charge, to any person obtaining a copy of
+# this software and associated documentation files (the "Software"), to deal in
+# the Software without restriction, including without limitation the rights to
+# use, copy, modify, merge, publish, distribute, sublicense, and/or sell copies of
+# the Software, and to permit persons to whom the Software is furnished to do so,
+# subject to the following conditions:
+#
+# The above copyright notice and this permission notice shall be included in all
+# copies or substantial portions of the Software.
+#
+# THE SOFTWARE IS PROVIDED "AS IS", WITHOUT WARRANTY OF ANY KIND, EXPRESS OR
+# IMPLIED, INCLUDING BUT NOT LIMITED TO THE WARRANTIES OF MERCHANTABILITY, FITNESS
+# FOR A PARTICULAR PURPOSE AND NONINFRINGEMENT. IN NO EVENT SHALL THE AUTHORS OR
+# COPYRIGHT HOLDERS BE LIABLE FOR ANY CLAIM, DAMAGES OR OTHER LIABILITY, WHETHER
+# IN AN ACTION OF CONTRACT, TORT OR OTHERWISE, ARISING FROM, OUT OF OR IN
+# CONNECTION WITH THE SOFTWARE OR THE USE OR OTHER DEALINGS IN THE SOFTWARE.
+
 import logging
 import operator
 from collections import defaultdict
--- a/dask_sql/physical/rel/logical/aggregate.py
+++ b/dask_sql/physical/rel/logical/aggregate.py
@@ -341,15 +341,18 @@ class LogicalAggregatePlugin(BaseRelPlugin):
         grouped_df = tmp_df.groupby(by=group_columns_and_nulls)
 
         # Convert into the correct format for dask
-        aggregations_dict = defaultdict(dict)
+        aggregations_dict = defaultdict(list)
+        input_output_cols = []
         for aggregation in aggregations:
             input_col, output_col, aggregation_f = aggregation
 
-            aggregations_dict[input_col][output_col] = aggregation_f
+            aggregations_dict[input_col].append(aggregation_f)
+            input_output_cols.append((input_col, output_col))
 
         # Now apply the aggregation
         logger.debug(f"Performing aggregation {dict(aggregations_dict)}")
         agg_result = grouped_df.agg(aggregations_dict)
+        agg_result.columns = input_output_cols
 
         # ... fix the column names to a single level ...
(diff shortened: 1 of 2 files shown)


dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -1,3 +1,24 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
sarahyurick, line 3: We include this in the header by default, but you may request to remove it.

dask_sql/physical/rel/logical/join.py
```diff
@@ -1,3 +1,24 @@
+# The MIT License (MIT)
+#
+# Copyright (c) 2021, NVIDIA CORPORATION.
```
sarahyurick, line 3: We include this in the header by default, but you may request to remove it.

dask_sql/physical/rel/logical/join.py
```diff
@@ -117,7 +138,7 @@ def merge_single_partitions(lhs_partition, rhs_partition):
                 # which is definitely not possible (java dependency, JVM start...)
                 lhs_partition = lhs_partition.assign(common=1)
                 rhs_partition = rhs_partition.assign(common=1)
-                merged_data = pd.merge(lhs_partition, rhs_partition, on=["common"])
+                merged_data = dd.multi.merge(lhs_partition, rhs_partition, on=["common"])
```
nils-braun, line 141: That makes sense! I had a look into the dask-dataframe functions and they have a fallback to the pandas functionality (if it is actually a pandas object) - so there is no difference for dask-pandas (but for other dask-like libraries, there might be). Cool!

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -341,15 +362,18 @@ def _perform_aggregation(
...
             input_col, output_col, aggregation_f = aggregation
 
-            aggregations_dict[input_col][output_col] = aggregation_f
+            aggregations_dict[input_col].append(aggregation_f)
+            input_output_cols.append((input_col, output_col))
 
         # Now apply the aggregation
         logger.debug(f"Performing aggregation {dict(aggregations_dict)}")
         agg_result = grouped_df.agg(aggregations_dict)
+        agg_result.columns = input_output_cols
```
nils-braun, line 376: This will lead to problems with the order if you have multiple input columns with multiple aggregations each.
Imagine you have an list of aggregations something like this:
```python
aggregations = [
    ("A", "output_1", "sum"),
    ("B", "output_2", "sum"),
    ("A", "output_3", "mean"),
    ("B", "output_4", "mean"),
]
```
After your code, the `dict` will contain `{'A': ['sum', 'mean'], 'B': ['sum', 'mean']}` (which is expected) and the list of columns will be `[('A', 'output_1'), ('B', 'output_2'), ('A', 'output_3'), ('B', 'output_4')]` - as the order will be exactly as in the `aggregations` list. 
The problem however is, that the `dict` does not have this order and the grouping will be ordered by input column:
```python
df = pd.DataFrame({'A': [1, 2, 3, 4], 'B': [1, 2, 3, 4], "id": [1, 1, 2, 2]})
result = df.groupby("id").agg(aggregations_dict)
result.columns.to_list()
```
gives `[('A', 'sum'), ('A', 'mean'), ('B', 'sum'), ('B', 'mean')]` (for dask it will be similar). If you now apply the column names from your  `input_output_cols`, we will map the wrong columns (and this is also why the test in the CI is failing).

I can totally understand why you implemented the change and I definitely also do not want to use a deprecated function, but solving this differently might be hard. We would probably need to use `ordereddicts` (to make sure we understand the order of the input columns) and then build a new list based on input column and aggregation list by ourselves. 

One question though: is it also deprecated for dask or only for pandas? Because I still see it mentioned in the [dask function which handles this](https://github.com/dask/dask/blob/main/dask/dataframe/groupby.py#L585) without any warning.

dask_sql/physical/rel/logical/aggregate.py
nils-braun: I think in this case I would vote to remove it, if you can. Because with just the single copyright text from NVIDIA it looks like this is a NVIDIA-owned project (even though copyright != ownership). I am not sure what your policy is though - do you need to have some kind of mentioning of NVIDIA? I am happy to add a "authors" section or alike.

nils-braun:
Thanks @sarahyurick for the change and for providing the code for your issues right away!

dask_sql/physical/rel/logical/aggregate.py
sarahyurick: Ah okay thanks, I see the issue! Using a dict-of-dicts is still supported in Dask, although it's not supported in cuDF or Dask cuDF, which is what motivated the update. If it's alright with you, I'll look into the `ordereddicts` approach and get back to you?

sarahyurick: I think you're right - the `OrderedDict` approach is turning out to be more complicated than desired. We're currently looking into supporting dict-of-dicts support on the Dask cuDF side now. If it's okay, I'm going to keep this PR open for now as we're running additional tests with it, but I think eventually I'll remove the changes from `aggregate.py`.

nils-braun: Sure, fine with me. Take your time!

sarahyurick:
UPDATE: Moved `join.py` changes to #211, so hopefully that can be merged easily.

PR #211, opened 2021-08-13
Update `join.py`

sarahyurick:
Moved the `join.py` changes from #207 here, will remove them from 207. Should (hopefully!) be OK to merge.

### DIFF
--- a/dask_sql/physical/rel/logical/join.py
+++ b/dask_sql/physical/rel/logical/join.py
@@ -5,7 +5,6 @@ from functools import reduce
 from typing import List, Tuple
 
 import dask.dataframe as dd
-import pandas as pd
 from dask.base import tokenize
 from dask.highlevelgraph import HighLevelGraph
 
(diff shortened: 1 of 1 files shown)


dask_sql/physical/rel/logical/join.py
```diff
@@ -117,7 +116,7 @@ def merge_single_partitions(lhs_partition, rhs_partition):
                 # which is definitely not possible (java dependency, JVM start...)
                 lhs_partition = lhs_partition.assign(common=1)
                 rhs_partition = rhs_partition.assign(common=1)
-                merged_data = pd.merge(lhs_partition, rhs_partition, on=["common"])
+                merged_data = dd.multi.merge(lhs_partition, rhs_partition, on=["common"])
```
rjzamora, line 119:
```suggestion
                merged_data = lhs_partition.merge(rhs_partition, on=["common"])
```

This is a good idea, but it will not preduce a cudf DataFrame (which is what you would want for the output of a single partition), even if lhs_partition and rhs_partition are both cudf DataFrames.  This is because `dd.multi.merge` has an explicit check for `pd.DataFrame` instances, but not for cudf.

nils-braun:
@sarahyurick - thanks for splitting, that will make it easier to review.
To get rid of the style check error, I would recommend to use `pre-commit` as described in the [docu](https://dask-sql.readthedocs.io/en/latest/pages/installation.html#for-development), which will eventually call `black`.

dask_sql/physical/rel/logical/join.py
sarahyurick: That worked, and made the fix at https://github.com/dask/dask/pull/7990 unnecessary! (It also made it so we only have to call `.compute()` once, as desired.) Thanks!

nils-braun:
Cool, looks good now!

PR #216, opened 2021-08-18
Use the mambaforge installer to speed up the build process

nils-braun:
Our current CI build process involves multiple conda installations. I could already speed it up in the past by switching from conda to mamba - with the drawback that the setup of the conda installation (miniconda) takes a long time as it needs to install mamba.
Let's hope that starting with "mambaforge" makes it faster.


nils-braun:
Its hard to tell quantitatively, but it seems the build is at least 1-2 min faster (even without caching).
While doing this test, I realized that the windows build does not pick up the conda cache correctly. It states that the `repodata_record.json` file is corrupted and therefore needs to reinstall all the packages, which takes at least twice as long. This should be fixed in a second iteration.

PR #219, opened 2021-08-18
Creating tables with a dask-cudf DataFrame

sarahyurick:
Dask-SQL can create tables directly from storage in SQL syntax:
```
CREATE TABLE my_data WITH (
    format = 'csv',
    location = './iris.csv'
)
```
As suggested by @randerzander it would be nice to have a way to create this backed by a dask-cudf DF instead of a Dask CPU DF. This is my first step toward achieving that. 

So far, I just added a boolean parameter `gpu`, although perhaps something like `HOW='gpu'` vs default `HOW='cpu'` would be better. These changes are also only for reading in a single file. Mainly want to get your thoughts about how to support this @nils-braun :)

For example, something like this works with this PR: 
```
CREATE OR REPLACE TABLE my_data WITH (
    format = 'csv',
    location = './iris.csv',
    gpu=True
)
```

### DIFF
--- a/dask_sql/physical/rel/custom/create_table.py
+++ b/dask_sql/physical/rel/custom/create_table.py
@@ -62,11 +62,17 @@ class CreateTablePlugin(BaseRelPlugin):
         except KeyError:
             raise AttributeError("Parameters must include a 'location' parameter.")
 
+        try:
+            gpu = kwargs.pop("gpu")
+        except KeyError:
+            gpu = False
+
         context.create_table(
             table_name,
             location,
             format=format,
             persist=persist,
             schema_name=schema_name,
+            gpu=gpu,
             **kwargs,
         )
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -125,6 +125,7 @@ class Context:
         format: str = None,
         persist: bool = True,
         schema_name: str = None,
+        gpu: bool = False,
         **kwargs,
     ):
         """
@@ -199,6 +200,7 @@ class Context:
             table_name=table_name,
             format=format,
             persist=persist,
+            gpu=gpu,
             **kwargs,
         )
         self.schema[schema_name].tables[table_name.lower()] = dc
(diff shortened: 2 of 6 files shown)


sarahyurick:
Looks like the tests are failing from everywhere else the new parameter would need to be added...

dask_sql/physical/rel/custom/create_table.py
```diff
@@ -62,11 +62,17 @@ def convert(
         except KeyError:
             raise AttributeError("Parameters must include a 'location' parameter.")
 
+        try:
+            gpu = kwargs.pop("gpu")
+        except KeyError:
+            gpu = False
+
```
nils-braun, line 69: The `pop` function allows for a default argument, so you can write
```suggestion
        gpu = kwargs.pop("gpu", False)
```

nils-braun (commented):
Thanks @sarahyurick !
I see two possibilities to work around the additional-keyword-argument-problem:
* Either you add into each input plugin which does not support gpu a first line with throws an exception when gpu is set to true
* or you do not specifically use a gpu argument in the main functions and let only the locations input plugin handle it. Additional arguments are passed via `kwargs`, so you could just extract it from there (for example the intake plugin has a `intake_table_name`, which is only used for this class)

sarahyurick:
Okay, it looks like there were only 2 other places in the code (in `pandas.py` and `intake.py`) where it was expecting the extra parameter, so I updated those. What do you think? I also like your idea to let only the locations input plugin handle it, so I could definitely do that instead, especially if you think there are any places I am potentially overlooking.

nils-braun:
Sorry @sarahyurick - I just realised that the coverage was not shown correctly (because we moved to the new git space) and that the hive test was disabled.
I will fix both in #221 - you might need to update to the newest main version once merged. After that, we can check together if you maybe need some additional tests or any `pragma: no cover` annotations.

sarahyurick:
Sounds good! There were some conflicts that required me to re-open this PR as #224.

nils-braun:
Ok, thanks! Just as a note (but you might already know this): you do not need to create a new PR if there are conflicts - you can just rebase and force-push or (preferred) merge the new main branch into your branch, solve the merge conflicts and push again. But also opening a new PR is fine :-)

PR #221, opened 2021-08-20
Re-enable the hive tests

nils-braun:
Due to me being stupid, there was no hive test in the moment.
Additionally, the move to a new git space disabled all coverage tests and display of the results.
We are fixing both here and bring the coverage to 100% back again.


nils-braun:
The dark-cluster test fails until the dask docker image is also updated, so we can proceed here anyways.

PR #223, opened 2021-08-23
Improve Left/Right/Inner Join

### LINKED ISSUE #190: Full Join becomes Inner Join for inequality conditions
Code:
```python
import pandas as pd
import numpy as np
from dask_sql import Context
client = Client()
cont = Context()

df1 = pd.DataFrame({
    'dated': pd.date_range(pd.Timestamp('2021-01-01'), pd.Timestamp('2021-01-10')),
    'var1': np.ones(10)})
df2 = pd.DataFrame({
    'startdate': [pd.Timestamp('2020-12-30'), pd.Timestamp('2021-01-09')],
    'enddate': [pd.Timestamp('2021-01-03'), pd.Timestamp('2021-01-20')],
    'var2': np.array([2.0, 3.0])})

cont.create_table('df1', df1)
cont.create_table('df2', df2)

df3 = cont.sql(
    """select a.*, b.var2
    from df1 a left join df2 b
    on b.startdate<=a.dated and a.dated<=b.enddate""").compute()
```
Results:
* `df1`:
```
       dated  var1
0 2021-01-01   1.0
1 2021-01-02   1.0
2 2021-01-03   1.0
3 2021-01-04   1.0
4 2021-01-05   1.0
5 2021-01-06   1.0
6 2021-01-07   1.0
7 2021-01-08   1.0
8 2021-01-09   1.0
9 2021-01-10   1.0
```
* `df2`:
```
   startdate    enddate  var2
0 2020-12-30 2021-01-03   2.0
1 2021-01-09 2021-01-20   3.0
```
* `df3`:
```
        dated  var1  var2
0  2021-01-01   1.0   2.0
2  2021-01-02   1.0   2.0
4  2021-01-03   1.0   2.0
17 2021-01-09   1.0   3.0
19 2021-01-10   1.0   3.0
```
This is an Inner Join, not Left Join.

The correct output should be as follows, using sqlite3:
```python
import sqlite3
# Connect database
conn = sqlite3.connect(':memory:')
df1.to_sql('df1', conn, index=False)
df2.to_sql('df2', conn, index=False)

df3 = pd.read_sql_query(
    """select a.*, b.var2
    from df1 a left join df2 b
    on b.startdate<=a.dated and a.dated<=b.enddate""", 
    conn, 
    parse_dates=['dated'])
```
where `df3` is
```
       dated  var1  var2
0 2021-01-01   1.0   2.0
1 2021-01-02   1.0   2.0
2 2021-01-03   1.0   2.0
3 2021-01-04   1.0   NaN
4 2021-01-05   1.0   NaN
5 2021-01-06   1.0   NaN
6 2021-01-07   1.0   NaN
7 2021-01-08   1.0   NaN
8 2021-01-09   1.0   3.0
9 2021-01-10   1.0   3.0
```

flcong:
* Fix #190
* For left (right) join, set `pd.NA` to observations not matched by equality or inequality condition from the right (left) table.
* Add new data sets and tests for complex join.

### DIFF
--- a/dask_sql/physical/rel/logical/join.py
+++ b/dask_sql/physical/rel/logical/join.py
@@ -5,9 +5,6 @@ from functools import reduce
 from typing import List, Tuple
 
 import dask.dataframe as dd
-
-# Need pd.NA
-import pandas as pd
 from dask.base import tokenize
 from dask.highlevelgraph import HighLevelGraph
 
@@ -95,32 +92,12 @@ class LogicalJoinPlugin(BaseRelPlugin):
         # 4. dask can only merge on the same column names.
         # We therefore create new columns on purpose, which have a distinct name.
         assert len(lhs_on) == len(rhs_on)
-        # Add two columns (1,2,...) to keep track of observations in left and
-        # right tables. They must be at the end of the columns since
-        # _join_on_columns needs the relative order of columns (lhs_on and rhs_on)
-        # Only dask-supported functions are used (assign and cumsum) so that a
-        # compute() is not triggered.
-        df_lhs_renamed = df_lhs_renamed.assign(left_idx=1)
-        df_lhs_renamed = df_lhs_renamed.assign(
-            left_idx=df_lhs_renamed["left_idx"].cumsum()
-        )
-        df_rhs_renamed = df_rhs_renamed.assign(right_idx=1)
-        df_rhs_renamed = df_rhs_renamed.assign(
-            right_idx=df_rhs_renamed["right_idx"].cumsum()
-        )
-
         if lhs_on:
             # 5. Now we can finally merge on these columns
             # The resulting dataframe will contain all (renamed) columns from the lhs and rhs
             # plus the added columns
-            # Need the indicator for left/right join
             df = self._join_on_columns(
-                df_lhs_renamed,
-                df_rhs_renamed,
-                lhs_on,
-                rhs_on,
-                join_type,
-                indicator=True,
+                df_lhs_renamed, df_rhs_renamed, lhs_on, rhs_on, join_type,
             )
         else:
             # 5. We are in the complex join case
@@ -139,7 +116,10 @@ class LogicalJoinPlugin(BaseRelPlugin):
                 # which is definitely not possible (java dependency, JVM start...)
                 lhs_partition = lhs_partition.assign(common=1)
                 rhs_partition = rhs_partition.assign(common=1)
-                merged_data = lhs_partition.merge(rhs_partition, on=["common"])
+                # Need to drop "common" here, otherwise metadata mismatches
+                merged_data = lhs_partition.merge(rhs_partition, on=["common"]).drop(
+                    columns=["common"]
+                )
 
                 return merged_data
 
@@ -171,28 +151,10 @@ class LogicalJoinPlugin(BaseRelPlugin):
                 ResourceWarning,
             )
 
-            # Add _merge to be consistent with the case lhs_on=True
-            df["_merge"] = "both"
-            df["_merge"] = df["_merge"].astype("category")
-        # Put newly added columns to the end
-        df = df[
-            df.columns.drop("left_idx").insert(
-                df.columns.get_loc("right_idx") - 1, "left_idx"
-            )
-        ]
-
-        # Completely reset index to uniquely identify each row since there
-        # could be duplicates. (Yeah. It may be better to inform users that
-        # index will break. After all, it is expected to be broken since the
-        # number of rows changes.
-        df = df.assign(uniqid=1)
-        df = df.assign(uniqid=df["uniqid"].cumsum()).set_index("uniqid")
-
         # 6. So the next step is to make sure
         # we have the correct column order (and to remove the temporary join columns)
-        # Need to exclude temporary columns left_idx and right_idx
-        correct_column_order = list(df_lhs_renamed.columns.drop("left_idx")) + list(
-            df_rhs_renamed.columns.drop("right_idx")
+        correct_column_order = list(df_lhs_renamed.columns) + list(
+            df_rhs_renamed.columns
         )
         cc = ColumnContainer(df.columns).limit_to(correct_column_order)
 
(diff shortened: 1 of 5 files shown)


nils-braun:
Thanks @flcong - looks like a lot of work! I might need some time checking the content. Make sure to ping me if I take too long... In the meantime, maybe @rajagurunath can have a first look (although this is also new code for you, sorry!)

flcong:
> Thanks @flcong - looks like a lot of work! I might need some time checking the content. Make sure to ping me if I take too long... In the meantime, maybe @rajagurunath can have a first look (although this is also new code for you, sorry!)

No problem! First time contributing to a python package.

rajagurunath:
Hi @flcong,

Thanks for fixing this issue and welcome to dask-sql and open source :)

tests/integration/test_join.py
```diff
@@ -184,3 +184,358 @@ def test_join_literal(c):
...
+        "enddate": "Int64",
+        "lk_int": "Int64",
+        "pn_str": "string",
+        "lk_str": "string",
+    }
+    for k, v in change_types.items():
+        dfcorrpn[k] = dfcorrpn[k].astype(v)
+
+    # Left Join
+    querypnl = """
```
rajagurunath, line 358: I have tested this Query using PostgresSQL, it somewhat matches with the correct dataset specified here except for some `dtype` mismatch,  I have one suggestion can Please you add this testing on `test_postgres.py`  as well, where you can make use of `assert_query_gives_same_result` to compare your result directly with Postgres result. let me know if you need any help.

dask_sql/physical/rel/logical/join.py
```diff
@@ -177,8 +218,91 @@ def merge_single_partitions(lhs_partition, rhs_partition):
...
+                df = df.merge(
+                    (df.groupby(idx_varname)["filter_condition"].agg("sum") < 1)
+                    .rename("ineq_unmatched")
+                    .to_frame(),
+                    left_on=idx_varname,
+                    right_index=True,
+                    how="left",
+                )
+                # Assign pd.NA
+                for v in df.columns[df.columns.str.startswith(other_varpre)]:
```
rajagurunath, line 261: Just tried running the tests and the same code in jupyter notebook , and got the following error ```TypeError: float() argument must be a string or a number, not 'NAType' ```, if you are not getting this error then it is probably due to environment mismatch error I guess, Or any other guess regarding this Error?
 maybe due to setting `pd.NA` ? what do you think @flcong?

tests/integration/test_join.py
```diff
@@ -184,3 +184,358 @@ def test_join_literal(c):
...
+        select a.*, b.startdate, b.enddate, b.lk_nullint, b.lk_int, b.lk_str,
```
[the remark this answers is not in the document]
flcong: All right. Let me check. To do the Postgre tests, I need to install docker, right?

dask_sql/physical/rel/logical/join.py
flcong: Which test triggers the error? It seems to be the conversion `float(pd.NA)`?

tests/integration/test_join.py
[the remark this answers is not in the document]
rajagurunath: Yes :), if not, a quick hack for testing purpose is you can hardcore any available postgresSQL ip address in the `engine` fixture of test_postgres.py.whichever is easier.

dask_sql/physical/rel/logical/join.py
rajagurunath: For me, first query (left join) was failing in `test_join_lricomplex`
 I will try to provide more context by tomorrow :)

flcong: OK. BTW, I use Python 3.8.10 and Pandas 1.3.2 to run the tests.

rajagurunath: That's a good catch @flcong, The Above Error was raised on  pandas version `1.2.4` and working fine in `pandas==1.3.2`

tests/integration/test_join.py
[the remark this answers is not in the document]
flcong: OK. I've added the tests in `test_postgres.py`. Also I don't know why but I cannot connect to the postgres container using the original `test_postgres.py`. So I exposed port 5432 in `client.containers.run` and change `address` to `"localhost"`. Then, the tests work fine.

rajagurunath (commented):
Maybe Once these changes are done, I request @nils-braun  for further review since I am also no expert in the `dask-sql/physical/rel/logical` part.

nils-braun (commented):
Ui @flcong, that was probably quite difficult to find out. Nice job. It took me quite some time to dig through the code, so I think it would be very cool if you can add a bit more documentation. Maybe you can use a small example Dataframe in the comments to show, what is actually wrong and how each step helps in solving it (e.g. the three cases you have and then the different merges later which rows they actually add). If this is too much to ask for, I can also take care of it.

I had some smaller comments in the code, but there is also a more general issue I would like to discuss here (and maybe it is just me being too naive): let me recap the problem. When we do a LEFT or RIGHT join with a filter condition, it happens that we get rid of all NAN-rows in the other Dataframe (and therefore the JOIN turns into an inner join) - I did understand this part. In my naive understanding (please correct me) the solution should be rather easy: if in the resulting Dataframe in the end a certain line from the left (or right) Dataframe is missing and we are doing a LEFT (or RIGHT) join, we still need to keep it. Shouldn't here a very simple `join` with the initial left (or right) Dataframe before the merge and on one of your newly created index columns be enough here? Your solution might also work (I did not think through all corner cases, but looks good so far) but it is quite complex and involves multiple merges (which are shuffles in the Dask world). Am I too simple?

tests/integration/test_postgres.py
```diff
@@ -126,6 +128,92 @@ def test_join(assert_query_gives_same_result):
...
+    assert_query_gives_same_result(
+        """
+        select a.*, b.startdate, b.enddate, b.lk_nullint, b.lk_int, b.lk_str,
+            b.lk_float, b.lk_date
+        from user_table_ts a inner join user_table_lk2 b
+        on b.startdate<=a.dates and a.dates<=b.enddate
+        """,
+        ["dates", "startdate", "enddate"],
+        force_dtype="dask",
+        check_dtype=True,
```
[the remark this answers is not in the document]
flcong: I'm not sure how `assert_frame_equal` works to determine if, for example, `2.0` and `2` are identical when `check_dtype=False`. So here I try to make it more explicit that a type cast to the dask dataframe makes the two dataframes identical even if `check_dtype=True`. I guess it conveys more information for developers? (Maybe just my OCD.)

flcong:
> > Am I too simple?
> > Just to make my point clear: if you replace your changes in `join.py` inside the `filter_condition` check just with
> 
> ```
> logger.debug(f"Additionally applying filter {filter_condition}")
> df = filter_or_scalar(df, filter_condition)
> 
> # make sure we recover any lost rows in case of left, right or outer joins
> if join_type in ["left", "outer"]:
>     df = df.merge(df_lhs_renamed, on=list(df_lhs_renamed.columns), how="right")
> elif join_type in ["right", "outer"]:
>     df = df.merge(df_rhs_renamed, on=list(df_rhs_renamed.columns), how="right")
> 
> df = df.drop(columns=["left_idx", "right_idx"])
> ```
> 
> your tests still succeed

That's brilliant! I didn't think of that. I rolled back `join.py` to the previous version and added the lines you mentioned. 

One caveat is that in this case, if any input column has type int and there are unmatched rows, `np.nan` will be added and the column type will be cast to float, instead of nullable int. Since we don't know whether or not there are unmatched rows without triggering `compute()` (I guess), we have to either leave it as is (users may want to type-cast float into nullable int later) or we cast all int columns to nullable int (which may not be good either). What do you think?

By the way, I also find a bug in the `merge_single_partitions` function in `join.py`. (I don't know how you quote code in the conversation XD.) In the returned dataframe, there is an extra column `"common"` which causes a metadata mismatch for left join when there are only inequality conditions (e.g. my time-series table tests), so I dropped the column.

tests/integration/fixtures.py
```diff
@@ -86,6 +86,88 @@ def datetime_table():
     )
 
 
+@pytest.fixture
+def user_table_lk():
```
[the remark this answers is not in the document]
flcong: Yeah. I realize that. I simplified the new tests.

PR #224, opened 2021-08-23
Reading tables with a dask-cudf DataFrame

sarahyurick:
Updated version of #219. Also tagging @ayushdg if you have time to double check the `pandaslike.py` changes specifically?

### DIFF
--- a/dask_sql/input_utils/pandaslike.py
+++ b/dask_sql/input_utils/pandaslike.py
@@ -18,6 +18,20 @@ class PandasLikeInputPlugin(BaseInputPlugin):
         is_cudf_type = cudf and isinstance(input_item, cudf.DataFrame)
         return is_cudf_type or isinstance(input_item, pd.DataFrame) or format == "dask"
 
-    def to_dc(self, input_item, table_name: str, format: str = None, **kwargs):
+    def to_dc(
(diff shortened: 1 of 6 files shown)


sarahyurick:
Sounds good - I've updated `hive.py` and added some `# pragma: no cover`s - let me know if I missed any!

dask_sql/input_utils/pandaslike.py
```diff
@@ -18,6 +18,20 @@ def is_correct_input(
...
         npartitions = kwargs.pop("npartitions", 1)
-        return dd.from_pandas(input_item, npartitions=npartitions, **kwargs)
+        if gpu:  # pragma: no cover
+            import dask_cudf
+
+            return dask_cudf.from_cudf(
+                cudf.from_pandas(input_item), npartitions=npartitions, **kwargs,
+            )
+        else:
+            return dd.from_pandas(input_item, npartitions=npartitions, **kwargs)
```
ayushdg, line 37: Given that this input util accepts both `cudf` and `pandas` dataframes as valid inputs, you'd probably need an additional check here to check if `input_item` is a pandas dataframe or not, and call the `from_pandas` function only for that case.

PR #228, opened 2021-08-26
Add DISTRIBUTE BY to dask-sql grammar

jdye64:
Currently `DISTRIBUTE BY` is missing from the `dask-sql` grammar which is extended from `Postgres`. There are situations where it would be beneficial to allow for data to be distributed/shuffled across a Dask environment much like Spark (which has DISTRIBUTE BY) allows to ensure data locality.


nils-braun:
Sorry, or seems I gave you a non-working commit :-/ I can fix the style issues and I think also the first test problem with the missing "ANALYZE" later - but I think the missing /int comes from the fact, that the newer calcite version you introduced in the pom.xml seems to handle some operations differently. 
I have an implementation of /int in #226 (due to other reasons), so you can either wait for that or revert the switch to the newer calcite version here (which, I guess, is not strictly needed?)

jdye64:
Ahh yeah so I don’t need to introduce that newer version of calcite anymore longer. I was experimenting for a bit thinking I could include spark dialect in the RelationalAlgebraBuilder amd it “just work” but that spark dialect wasn’t in calcite-core until 1.27 so that is literally the only reason I version bumped. Now I don’t need it

nils-braun:
Thanks for the response @jdye64 - I did revert to the earlier version and also fixed the styling and the remaining test issue(s).
The only thing missing to do is to add a small test for the new "distributed" keyword. Would you like to do so or shall I do it?

jdye64:
I am happy to write the test. I wanted your opinion on the best way to perform the actual assertion in the pytest however? Seems like checking things like partition counts and where the data logically ended up might be platform independent? Therefore I am a little lost on what that test condition should actually look like / be testing? Do you have any ideas there?

jdye64:
@nils-braun I have discovered a situation where this grammar does not work as expected and was hoping you had some input. For Example if I have a query like 

`SELECT age, name FROM
people DISTRIBUTE BY age
ORDER BY name`

the parsing fails since it encounters the ORDER token and is expecting the EOF token. This is one of those situations where I really wish we could modify the base calcite grammar but as we already mentioned that doesn't seem feasible. What I would like to do is add something like another `OrderedQueryOrExpr` call in `distribute.ftl` after the DISTRIBUTE BY but that doesn't seem possible either? If the logic in `OrderedQueryOrExpr` definition was more granular I think it would be possible but it isn't. Any ideas?

tests/integration/test_distributeby.py
```diff
@@ -0,0 +1,30 @@
...
+    df_1 = pd.DataFrame({"id": [1, 2, 3], "val": [1, 2, 3]})
+
+    ddf_0 = dd.from_pandas(df_0, npartitions=1)
+    ddf_1 = dd.from_pandas(df_1, npartitions=1)
+    ddf = dd.concat([ddf_0, ddf_1])
+
+    partitioned_ddf = ddf.shuffle(on=["id"])
+    part_0_ids = partitioned_ddf.get_partition(0).compute().id.unique()
+    part_1_ids = partitioned_ddf.get_partition(1).compute().id.unique()
+    assert bool(set(part_0_ids) & set(part_1_ids)) is False
```
charlesbluca, line 17: If this is always a given, do we need to make this assertion? Or is it serving as some form of input validation / sanity check

tests/integration/test_distributeby.py
```diff
@@ -0,0 +1,30 @@
+import dask.dataframe as dd
+import pandas as pd
+
+
+def test_distribute_by(c):
```
charlesbluca, line 5: We might want to parametrize this to also run a GPU version of the test, which I think can be done by switching out the library used to create `df_0`/`df_1`.

tests/integration/test_distributeby.py
```diff
@@ -0,0 +1,30 @@
...
+    df_0 = pd.DataFrame({"id": [0, 1, 2], "val": [0, 1, 2]})
+
+    df_1 = pd.DataFrame({"id": [1, 2, 3], "val": [1, 2, 3]})
+
+    ddf_0 = dd.from_pandas(df_0, npartitions=1)
+    ddf_1 = dd.from_pandas(df_1, npartitions=1)
+    ddf = dd.concat([ddf_0, ddf_1])
```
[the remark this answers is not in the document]
jdye64: probably not. I will make that change now

tests/integration/test_distributeby.py
```diff
@@ -0,0 +1,34 @@
...
+
+@pytest.mark.parametrize("gpu", [False, pytest.param(True, marks=pytest.mark.gpu)])
+def test_distribute_by(c, gpu):
+    df = pd.DataFrame({"id": [0, 1, 2, 1, 2, 3], "val": [0, 1, 2, 1, 2, 3]})
+
+    if gpu:
+        gdf = cudf.from_pandas(df)
+        ddf = dask_cudf.from_cudf(df, npartitions=2)
+    else:
+        ddf = dd.from_pandas(df, npartitions=2)
```
charlesbluca, line 21: Small nitpick but we can probably simplify the dataframe creation a bit:

```python
if gpu:
    xd = cudf
else:
    xd = pd

df = xd.DataFrame({"id": [0, 1, 2, 1, 2, 3], "val": [0, 1, 2, 1, 2, 3]})
ddf = dd.from_pandas(df, npartitions=2)
```

Despite its name, `dd.from_pandas` will create a dask-cudf dataframe if `df` is a cuDF dataframe, which should allow us to circumvent importing dask-cudf here

tests/integration/test_distributeby.py
```diff
@@ -0,0 +1,33 @@
...
+        """
+    SELECT
+    id
+    FROM test
+    DISTRIBUTE BY id
+    """
+    )
+    part_0_ids = partitioned_ddf.get_partition(0).compute().id.unique()
+    part_1_ids = partitioned_ddf.get_partition(1).compute().id.unique()
+    assert bool(set(part_0_ids) & set(part_1_ids)) is False
```
jakirkham, line 33: This [fails in the GPU case]( https://gpuci.gpuopenanalytics.com/job/dask/job/dask-sql/job/prb/job/dask-sql-prb/151/CUDA_VER=11.2,LINUX_VER=ubuntu18.04,PYTHON_VER=3.8,RAPIDS_VER=21.12/testReport/tests.integration/test_distributeby/test_distribute_by_True_/ ) because neither of these are iterable

```python
>       assert bool(set(part_0_ids) & set(part_1_ids)) is False

tests/integration/test_distributeby.py:33: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
/opt/conda/envs/dask_sql/lib/python3.8/site-packages/cudf/core/single_column_frame.py:92: in __iter__
    cudf.utils.utils.raise_iteration_error(obj=self)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

obj = 0    3
Name: id, dtype: int64

    def raise_iteration_error(obj):
>       raise TypeError(
            f"{obj.__class__.__name__} object is not iterable. "
            f"Consider using `.to_arrow()`, `.to_pandas()` or `.values_host` "
            f"if you wish to iterate over the values."
        )
E       TypeError: Series object is not iterable. Consider using `.to_arrow()`, `.to_pandas()` or `.values_host` if you wish to iterate over the values.
```

May need to coerce them to something that is iterable first. Maybe like this?

```suggestion
    if gpu:
        part_0_ids = part_0_ids.to_pandas()
        part_1_ids = part_1_ids.to_pandas()
    assert bool(set(part_0_ids) & set(part_1_ids)) is False
```

charlesbluca (approved):
Looks like that resolved the GPU failures - happy to merge this once the tests pass, thanks for the work here @jdye64 😄

Issue #227, opened 2021-08-26
Strategy for running CI-tests against GPU machines


nils-braun:
With the recent integration of dask-cudf into dask-sql, it would be wise to start testing against GPU libraries and machines in the automated build triggered by GitHub actions.

@ayushdg mentioned on [#220](https://github.com/dask-contrib/dask-sql/issues/220#issuecomment-904761024_), that:
> GpuCI is one option that could be used to test integration with rapids libraries. It's what the dask repository uses as well (https://github.com/dask/community/issues/138).
> I don't think it uses GitHub actions specifically but is enabled via a webhook to GitHub.

I would be very much interested, if this is also an option for dask-contrib packages (and if this would be supported by RAPIDS) and what the current experience with the setup at the dask main repository is. Maybe, @quasiben, @jrbourbeau, can you shortly summarize the current setup in the dask main repository and how the experiences are? Thanks!

charlesbluca:
The basic setup for Dask's gpuCI consisted of:

- Marking relevant GPU tests with a custom pytest marker (see dask/dask#7876)
- Enabling gpuCI on a fork of Dask and getting the required setup files working there (see charlesbluca/dask#1); these files consist of
  - a [shell script](https://github.com/dask/dask/blob/main/continuous_integration/gpuci/build.sh) defining the steps to build the test environment and run the desired tests
  - a [YAML file](https://github.com/dask/dask/blob/main/continuous_integration/gpuci/axis.yaml) defining the build matrix
- Once gpuCI works in the forked branch, merging this branch into the main repo and enabling gpuCI there

Once enabled, gpuCI runs on all PRs opened by members on an allowlist (in Dask's case, this contains all members of the organization as well as some other people added using the commands listed [here](https://docs.dask.org/en/latest/develop.html#gpu-ci)); initially, there was some noisiness as the gpuCI bot pinged a large chunk of the open Dask PRs for approval to run GPU tests, but this should be avoidable here if we populate the allowlist before enabling.

Another thing to consider is that gpuCI on Dask and Distributed uses Docker images (generated nightly based on specifications [here](https://github.com/rapidsai/dask-build-environment/)) to speed up the overall test time; we might want to do something similar for dask-sql, but I'm interested in if it would be possible to reuse one of those images here.

PR #229, opened 2021-08-27
[Review] Add fast path for multi-column sorting

quasiben:
This PR adds an optional fast path for multi-column sorting for libraries like Dask-cuDF was has a work implementation.  The implementation currently does not support ascending/na_position kwargs -- we might be able to do something similar to what is currently implemented in [_sort_first_column](https://github.com/dask-contrib/dask-sql/blob/8002ad83653721a544905647695fd1c0871f336f/dask_sql/physical/utils/sort.py#L70) where we check for NAs/Descending.  For now, though, if the users needs those options the fall back is to the default implementation. 

We can add tests for this PR after https://github.com/dask-contrib/dask-sql/issues/227 is resolved.

Lastly, if we want to resolve multi-column sorting for Dask, we could use similar implementation in dask-cudf which relies on [multi-column quantiles](https://github.com/rapidsai/cudf/blob/4d8e401f778396bd909f5ea0f75b80ec52ebd74c/python/dask_cudf/dask_cudf/sorting.py#L120) .  This would mean adding a `quantiles` method to pandas

```python
In [30]: df = pd.DataFrame({'a': [1, 0, 11, 12, 2], 'b': [1, 2, 3, 4, 5], 'c': [0, 1, 5, 2, 3]})

In [31]: cdf = cudf.from_pandas(df)

In [32]: cdf[['c', 'a']].quantiles([0, 0.5, 1])
Out[32]:
     c   a
0.0  0   1
0.5  2  12
1.0  5  11

In [33]: df[['c', 'a']].quantile([0, 0.5, 1])
Out[33]:
       c     a
0.0  0.0   0.0
0.5  2.0   2.0
1.0  5.0  12.0

In [34]:
```

cc @rjzamora

### DIFF
--- a/dask_sql/physical/utils/sort.py
+++ b/dask_sql/physical/utils/sort.py
@@ -2,7 +2,6 @@ from typing import List
 
 import dask.dataframe as dd
 import pandas as pd
-from dask.array.routines import isin
 from dask.utils import M
 
 from dask_sql.utils import make_pickable_without_dask_sql


dask_sql/physical/utils/sort.py
```diff
@@ -6,12 +6,34 @@
...
     sort_columns: List[str],
     sort_ascending: List[bool],
     sort_null_first: List[bool],
 ) -> dd.DataFrame:
+
+    # Try fast path for multi-column sorting before falling back to
+    # sort_partition_func.  Tools like dask-cudf have a limited but fast
+    # multi-column sort implementation.  We check if any sorting/null sorting
+    # is required.  If so, we fall back to default sorting implementation
+    if any(sort_null_first) is False and all(sort_ascending) is True:
```
nils-braun, line 31: So by default, the multi column sort in dask-cudf sorts nulls last and sorts everything ascending? (I guess so, just wanted to double check)

dask_sql/physical/utils/sort.py
```diff
@@ -6,12 +6,34 @@
 from dask_sql.utils import make_pickable_without_dask_sql, new_temporary_column
 
 
+def multi_col_sort(
```
nils-braun, line 9: Maybe we can call it `fast_multi_col_sort` (or alike), because the other function also does a multi column sort in some way?

nils-braun (approved):
Cool, I like that! Once we are sure about the testing (as you have said) we can merge.

Regarding the multi-column sorting in dask: one problem is (as you mentioned) the quantile method, another one is however: how to handle the index, as Dask does currently not support multi-index (multi-column) divisions (as far as I recall). How do you handle this in dask-cudf?

nils-braun:
(side note: why is the code coverage still 100%... Shouldn't the sort_values function not be implemented? Strange. I am currently not on my laptop, but I will need to check later)

dask_sql/physical/utils/sort.py
nils-braun: Because (and that might be actually a problem), the `dask.DataFrame.sort_values` function does some "strange" (at least from the point of SQL) things to NaN values. As it is basically computing new divisions, which can not contain NaN values, it will just remove all NaN rows. You can test this:
```python
import pandas as pd
import dask.dataframe as dd

df = pd.DataFrame({"a": list(range(200)) + [None]*100})
ddf = dd.from_pandas(df, npartitions=11)

ddf.sort_values('a')
```

will return a Dataframe with length 200 (instead of 300). Even if the user did not specify to sort NaN values in a non-default way, I think we should not have this function as default - for Dask. Is this different in Dask-cudf (or cudf)? I am fine to have a special case handling for dask-cudf if it makes sense.

We actually should have a test for this. I just pushed one (dirty without PR, ehem) onto main in 9ebe16a53e7bbb89314061b38014e03f663e6a5b. This works on main, but fails with your PR (due to the issue I described).

dask_sql/physical/utils/sort.py
charlesbluca: I was thinking something closer to the original function name that also references the fact that this relies on dask-cudf's codepath (such as `cudf_apply_sort`) - how does that sound?

rjzamora: I think I'm somewhere in-between these two opinions :)

I'm hoping that `sort_values()` to work for both pandas and cudf-backed data in the future, so it would probably be better to avoid using `cudf` in the function name.  Also, the current multi-column sort for pandas-backed data produces correct ordering, but it does not produce "proper" partitioning.  My understanding is that the percentiles of the first column are used to define new divisions, which can lead to some ugly partitioning when the number of partitions is not small relative to the cardinality of the first column.

Therefore, I personally feel that it makes the most sense to start by trying to call the "proper" sort, and then fall back on the "workaround" for pandas data.  In other words, it seems to me like the pandas-based version should have a "special" name to make it clear that dask-sql is using a work-around for missing functionality.

dask_sql/physical/utils/sort.py
charlesbluca: Thanks for pointing this out! I think the additional condition `isinstance(df, dask_cudf.DataFrame)` might be necessary here to avoid Dask's single column case.

To answer your question, dask-cudf doesn't implement null sorting:

```python
import dask_cudf

gddf = dask_cudf.from_dask_dataframe(ddf)
gddf.sort_values("a")
```

Results in

```
ValueError                                Traceback (most recent call last)                                        
<ipython-input-7-6fc4356cae9b> in <module>                                                                         
----> 1 gddf.sort_values("a")                                                                                      
                                                                                                                   
/raid/charlesb/miniconda3/envs/dask/lib/python3.8/site-packages/dask_cudf/core.py in sort_values(self, by, ignore_i
ndex, max_branch, divisions, set_divisions, **kwargs)                                                              
    244             df = self.map_partitions(M.sort_values, by)                                                    
    245         else:                                                                                              
--> 246             df = sorting.sort_values(                                                                      
    247                 self,                                                                                      
    248                 by,                                                                                        
                                                                                                                   
/raid/charlesb/miniconda3/envs/dask/lib/python3.8/site-packages/dask_cudf/sorting.py in sort_values(df, by, max_bra
nch, divisions, set_divisions, ignore_index)                                                                       
    223     # Step 1 - Calculate new divisions (if necessary)                                                      
    224     if divisions is None:                                                                                  
--> 225         divisions = quantile_divisions(df, by, npartitions)                                                
    226                                                                                                            
    227     # Step 2 - Perform repartitioning shuffle                                                              
                                                                                                                   
/raid/charlesb/miniconda3/envs/dask/lib/python3.8/site-packages/dask_cudf/sorting.py in quantile_divisions(df, by, 
npartitions)                                                                                                       
    188         divisions.iloc[-1] += 1                                                                            
    189         divisions = sorted(                                                                                
--> 190             divisions.drop_duplicates().astype(dtype).values.tolist()                                      
    191         )                                                                                                  
    192     else:                                                                                                  
                                                                                                                   
/raid/charlesb/miniconda3/envs/dask/lib/python3.8/site-packages/cudf/core/frame.py in values(self)                 
   4528         <class 'cupy.core.core.ndarray'>                                                                   
   4529         """                                                                                                
-> 4530         return self._column.values                                                                         
   4531                                                                                                            
   4532     @property                                                                                              
                                                                                                                   
/raid/charlesb/miniconda3/envs/dask/lib/python3.8/site-packages/cudf/core/column/column.py in values(self)         
    142                                                                                                            
    143         if self.has_nulls:                                                                                 
--> 144             raise ValueError("Column must have no nulls.")                                                 
    145                                                                                                            
    146         return cupy.asarray(self.data_array_view)                                                          
                                                                                                                   
ValueError: Column must have no nulls. 
```

With this in mind, we should probably adjust our try/except block to look out for errors on dask-cudf's end that can be handled by dask-sql (i.e. presence of nulls in the dataframe).

dask_sql/physical/utils/sort.py
charlesbluca: Just to clarify, would that look like moving the original content of `apply_sort`:

https://github.com/dask-contrib/dask-sql/blob/4d5f7dd4c133f7f2c584db2492c8f0ef3cba7e31/dask_sql/physical/utils/sort.py#L15-L38

To its own function that specifies it is a workaround, and then have `apply_sort` do something like:

```python
if isinstance(df, dask_cudf.DataFrame) and not any(sort_null_first) and all(sort_ascending):
    try:
        df = df.sort_values(sort_columns)
        return df.persist()
    except ValueError:
        pass

return workaround_sort_function(...) 
```

charlesbluca:
I'm going to be taking this PR over from @quasiben - could I get workflow approval?

quasiben:
> I'm going to be taking this PR over from @quasiben - could I get workflow approval?

Thanks @charlesbluca I approved so you should be able to get CI runs now

dask_sql/physical/utils/sort.py
rjzamora: Yes - That is exactly what I have in mind.  It feels to me like it makes more sense to isolate the "workaround" in a special function, rather than the proper sort_values usage.  However, I'd be happy to defer to @nils-braun if he has a different preference.

charlesbluca: For now, I'm going to opt for a conservative middle ground by placing the content of `multi_col_sort` into `apply_sort`, since it's only 2 lines. If @nils-braun is okay with the restructuring as you've described, then I'll go ahead and split off dask-sql's sorting implementation into a separate function.

quasiben:
@nils-braun if you have time I think this is ready for another review

charlesbluca:
FYI I am currently working on implementing `ascending` in Dask's `sort_values`:

- https://github.com/dask/dask/pull/8130

If that gets merged first, we could update the sorting logic here to reflect this change

dask_sql/physical/utils/sort.py
```diff
@@ -12,6 +17,22 @@ def apply_sort(
...
+    # multi-column sort implementation.  We check if any sorting/null sorting
+    # is required.  If so, we fall back to default sorting implementation
+    if (
+        dask_cudf is not None
+        and isinstance(df, dask_cudf.DataFrame)
+        and all(sort_ascending)
+        and not any(sort_null_first)
+    ):
+        try:
+            df = df.sort_values(sort_columns)
```
ayushdg, line 31: It might make sense to `ignore_index` here to make this behavior consistent with Dask-sql when we don't fallback to the fast path

charlesbluca:
In some cases, dask-cudf's sorting doesn't match up with dask-sql's; for example:

```python
import cudf
import pandas as pd
from dask_sql import Context

df = pd.DataFrame({"user_id": [2, 1, 2, 3], "b": [3, 3, 1, 3]})

c = Context()

c.create_table("gdf", df, gpu=True, npartitions=2)
c.create_table("pdf", df, gpu=False, npartitions=2)

got, expect = (c.sql(
    """
SELECT
    *
FROM %s
ORDER BY user_id
""" % s
).compute() for s in ("gdf", "pdf"))

print(got)
print(expect)

   user_id  b
0        1  3
0        2  3
1        2  1
2        3  3
   user_id  b
0        1  3
0        2  1
1        2  3
2        3  3
```

I'm not sure if the dask-sql's sorting of rows with identical values in the sort-by column is intentional, but for now I worked around this by testing the SQL sorted dask-cudf table against the same table sorted by calling `sort_values(by=..., ignore_index=True)` directly.

PR #230, opened 2021-08-28
Introduces parallel tests to speed up the processing

nils-braun:
This PR adds pytest-xdist on Github Actions


nils-braun:
It is approx. 30s faster in the tests, which is ~20%. So I think this is worth it.

PR #235, opened 2021-09-02
Add flake8 pre-commit hook

charlesbluca:
Adds flake8 pre-commit hook, along with some exclusions / changes to resolve the majority of resulting style failures.

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -2,7 +2,6 @@ import asyncio
 import inspect
 import logging
 import warnings
-from collections import namedtuple
 from typing import Any, Callable, Dict, List, Tuple, Union
 
 import dask.dataframe as dd
@@ -600,7 +599,7 @@ class Context:
         """
         Stop a SQL server started by ``run_server`.
         """
-        if not self.sql_server is None:
+        if self.sql_server is not None:
             loop = asyncio.get_event_loop()
             assert loop
             loop.create_task(self.sql_server.shutdown())
@@ -767,7 +766,8 @@ class Context:
 
         try:
             return str(s.toSqlString(default_dialect))
-        except:  # pragma: no cover. Have not seen any instance so far, but better be safe than sorry.
+        # Have not seen any instance so far, but better be safe than sorry
+        except:  # noqa: E722; pragma: no cover
             return str(s)
 
     def _get_tables_from_stack(self):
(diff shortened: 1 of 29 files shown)


setup.cfg
```diff
@@ -0,0 +1,21 @@
...
+# https://pycodestyle.pycqa.org/en/latest/intro.html#error-codes
+exclude = __init__.py,versioneer.py
+ignore =
+    E203,  # whitespace before ':'
+    E231,E241,  # Multiple spaces around ","
+    E731,  # Assigning lambda expression
+    #E741,  # Ambiguous variable names
+    W503,  # line break before binary operator
+    W504,  # line break after binary operator
+    F821,  # undefined name
```
charlesbluca, line 14: This exclusion is to resolve failures that come when we try to specify a param type with a class from a module we haven't imported; for example:

https://github.com/dask-contrib/dask-sql/blob/4d5f7dd4c133f7f2c584db2492c8f0ef3cba7e31/dask_sql/physical/rel/convert.py#L35-L37

Are we okay with keeping this exclusion, or would we prefer importing these modules where this failure comes up?

dask_sql/context.py
```diff
@@ -767,7 +766,8 @@ def _to_sql_string(self, s: "org.apache.calcite.sql.SqlNode", default_dialect=No
 
         try:
             return str(s.toSqlString(default_dialect))
-        except:  # pragma: no cover. Have not seen any instance so far, but better be safe than sorry.
+        # Have not seen any instance so far, but better be safe than sorry
+        except:  # noqa: E722; pragma: no cover
```
charlesbluca, line 770: This except block, along with handful others that should pop up in testing, is bare. Is there any way we can specify errors/exceptions for these? Or should we just `noqa` them?

dask_sql/input_utils/hive.py
```diff
@@ -126,7 +126,7 @@ def wrapped_read_function(location, column_information, **kwargs):
             else:  # pragma: no cover
                 # prevent python to optimize it away and make coverage not respect the
                 # pragma
-                dummy = 0
+                dummy = 0  # noqa: F841
```
charlesbluca, line 129: I understand that this is here for sake of coverage, but is there any other way we can retain the `else` block without create an unused variable?

PR #236, opened 2021-09-03
[WIP] Fast path when possible for non numeric aggregation

VibhuJawa:
This PR adds fast path when possible for non numeric aggregation

### DIFF
--- a/dask_sql/physical/rel/logical/aggregate.py
+++ b/dask_sql/physical/rel/logical/aggregate.py
@@ -304,7 +304,7 @@ class LogicalAggregatePlugin(BaseRelPlugin):
                     )
             if isinstance(aggregation_function, AggregationSpecification):
                 dtype = df[input_col].dtype
-                if pd.api.types.is_numeric_dtype(dtype):
+                if pd.api.types.is_numeric_dtype(dtype) or pd.api.types.is_datetime64_any_dtype(dtype):
                     aggregation_function = aggregation_function.numerical_aggregation
                 else:
                     aggregation_function = (
--- a/dask_sql/physical/rel/logical/aggregate.py
+++ b/dask_sql/physical/rel/logical/aggregate.py
@@ -61,6 +61,19 @@ class AggregationSpecification:
             non_numerical_aggregation or numerical_aggregation
         )
 
+    def get_dtype_relevant_aggregation(self, dtype):
+        numeric_agg = self.numerical_aggregation
+
+        if pd.api.types.is_numeric_dtype(dtype):
+            return numeric_agg
+
+        # Todo: Add Categorical and String Checks
+        if numeric_agg in ["min", "max", "first"]:
+            if pd.api.types.is_datetime64_any_dtype(dtype):
+                return numeric_agg
+
+        return self.non_numerical_aggregation
+
 
 class LogicalAggregatePlugin(BaseRelPlugin):
     """
@@ -304,12 +317,9 @@ class LogicalAggregatePlugin(BaseRelPlugin):
                     )
             if isinstance(aggregation_function, AggregationSpecification):
                 dtype = df[input_col].dtype
-                if pd.api.types.is_numeric_dtype(dtype) or pd.api.types.is_datetime64_any_dtype(dtype):
-                    aggregation_function = aggregation_function.numerical_aggregation
-                else:
-                    aggregation_function = (
-                        aggregation_function.non_numerical_aggregation
-                    )
+                aggregation_function = aggregation_function.get_dtype_relevant_aggregation(
+                    dtype
+                )
 
             # Finally, extract the output column name
             output_col = str(agg_call.getValue())


dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -61,6 +61,21 @@ def __init__(self, numerical_aggregation, non_numerical_aggregation=None):
             non_numerical_aggregation or numerical_aggregation
         )
 
+    def get_dtype_relevant_aggregation(self, dtype):
+        numeric_agg = self.numerical_aggregation
+
+        if pd.api.types.is_numeric_dtype(dtype):
+            return numeric_agg
+
+        # Todo: Add Categorical when support comes to dask-sql
```
VibhuJawa, line 70: Have not added Categorical as thats not supported currently with dask-sql . 

`string` seems to work as expected for these three aggregations.

VibhuJawa:
The PR should be ready for an inital review.

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -61,6 +61,21 @@ def __init__(self, numerical_aggregation, non_numerical_aggregation=None):
             non_numerical_aggregation or numerical_aggregation
         )
 
+    def get_dtype_relevant_aggregation(self, dtype):
+        numeric_agg = self.numerical_aggregation
```
ayushdg, line 65: Not a strong preference but might be worth renaming `numeric_aggregation` `numeric_agg` to something that specifies that these aggregations are not specific to `numerical` types only. 
I don't have a good suggestion for alternative names, but I think of them as `supported` or `native` aggregations.

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -61,6 +61,21 @@ def __init__(self, numerical_aggregation, non_numerical_aggregation=None):
...
+
+        if pd.api.types.is_numeric_dtype(dtype):
+            return numeric_agg
+
+        # Todo: Add Categorical when support comes to dask-sql
+        if numeric_agg in ["min", "max", "first"]:
```
ayushdg, line 71: Since `first` is the same aggregation for both numeric and non-numeric, no special check is needed for this aggregation. Though for clarity it might make sense to include it here, as well as include `count` which falls in the same criteria.

ayushdg (changes requested):
The changes look great and have a huge performance gains for these specific cases.  Added a few comments inline.

It would be great to include some test cases to [test_groupby](https://github.com/dask-contrib/dask-sql/blob/main/tests/integration/test_groupby.py) as well that add code coverage for this case.

dask_sql/physical/rel/logical/aggregate.py
VibhuJawa: Changed it to `native` and `custom` with comit https://github.com/dask-contrib/dask-sql/pull/236/commits/86b3132f28e366524fb33cd499bc0b223307d04d . Let me know if you would prefer another name,

dask_sql/physical/rel/logical/aggregate.py
VibhuJawa: Resolved with removing first  (see commit: https://github.com/dask-contrib/dask-sql/pull/236/commits/89822d92dd9436262d4096fcbecbb371207c6bfe) .

Thanks for the review , good catch on this.

VibhuJawa:
> The changes look great and have a huge performance gains for these specific cases. Added a few comments inline.


Thanks , have addressed all of your reviews, I think this is ready for another review. 

> It would be great to include some test cases to [test_groupby](https://github.com/dask-contrib/dask-sql/blob/main/tests/integration/test_groupby.py) as well that add code coverage for this case.

I have included test cases for `StringArray` here, https://github.com/dask-contrib/dask-sql/pull/236/commits/ea48c7a08bef1ec216ebc930e86808566a067b57  in `test_compatibility.py` . 


Adding similar tests was not possible due to not being able to test equality with datetime columns due to a pandas issue with  testing equality b/w string and datetime columns. (Raised the issue here: pd.testing https://github.com/pandas-dev/pandas/issues/43707)  .


For test cases corresponding to `cudf/dask_cudf` , i think we should first let https://github.com/dask-contrib/dask-sql/pull/229/files land to prevent any merge conflicts, Happy to figure something out if you differ on this.

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -48,18 +55,42 @@ class AggregationSpecification:
...
+                return native_agg
+
+            if pd.api.types.is_string_dtype(series.dtype):
+                ## If dask_cudf strings dtype return native aggregation
+                if dask_cudf is not None and isinstance(series, dask_cudf.Series):
+                    return native_agg
+
+                ## If pandas StringDtype native aggregation works
+                ## and with ObjectDtype and Nulls native aggregation can fail
+                if isinstance(series, dd.Series) and isinstance(
```
VibhuJawa, line 88: Essentially we need to take the  native agg code path for `Min/Max` if the `series` is either `dask_cudf` with `StringDtype` or `dask.DataFrame` with `pd.StringDtype` .   We want to take the custom agg code path if the dtype is an `object` .  



This is because `.min/max` on an object dtype series with nulls in pandas fails while it works for `StringDtype`   in pandas and works well for `cuDF` too. 
 

```python
#### Works
pd.Series(['a',None],dtype='string').min()
```
#### Fails
```
pd.Series(['a',None],dtype='object').min()

```

I wonder if we can have a cleaner combined check in `cuDF` . CC: @galipremsagar , @ayushdg  for thoughts ?

PR #238, opened 2021-09-14
Add conda dev environment for Python 3.7, 3.8

charlesbluca:
Meant to address [this issue ](https://github.com/rapidsai/dask-build-environment/pull/4#discussion_r706417850) with the planned dask-sql gpuCI images, this PR adds conda environments for Python 3.7 and 3.8:

- `environment-3.*.yaml` is essentially a copy of `conda.txt`, but specifying Python version; we might be able to trim more from this env if we would like it to represent the bare requirements for dask-sql
- `environment-3.*-dev.yaml` contains all the packages from `conda.txt`, as well as those required for testing and development; this can be used in place of `conda.txt` in the GitHub Actions workflows

I also made some minor changes in the Hive tests so that they are skipped with an informative message if we don't have the required Docker images pulled.

### DIFF
--- a/continuous_integration/environment-3.7-dev.yaml
+++ b/continuous_integration/environment-3.7-dev.yaml
@@ -1,0 +1,37 @@
+name: dask-sql
+channels:
+- conda-forge
+- defaults
+dependencies:
+- black=19.10b0
+- dask-ml>=1.7.0
+- dask>=2.19.0,!=2021.3.0  # dask 2021.3.0 makes dask-ml fail (see https://github.com/dask/dask-ml/issues/803)
+- fastapi>=0.61.1
+- intake>=0.6.0
+- isort=5.7.0
+- jpype1>=1.0.2
+- lightgbm>=3.2.1
+- maven>=3.6.0
+- mlflow>=1.19.0
+- mock>=4.0.3
+- nest-asyncio>=1.4.3
+- openjdk>=8
+- pandas>=1.0.0  # below 1.0, there were no nullable ext. types
+- pip=20.2.4
+- pre-commit>=2.11.1
+- prompt_toolkit>=3.0.8
+- psycopg2>=2.9.1
+- pyarrow>=0.15.1
+- pygments>=2.7.1
+- pyhive>=0.6.4
+- pytest-cov>=2.10.1
+- pytest-xdist
+- pytest>=6.0.1
+- python=3.7
+- scikit-learn>=0.24.2
+- sphinx>=3.2.1
+- tpot>=0.11.7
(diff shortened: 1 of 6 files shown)


rajagurunath:
Hi @charlesbluca,

Nice Work!!  Thanks for the changes/PR, Everything looks bit more organized now, the PR already looks good to me, let me Approve the GitHub workflow! 👍

.github/workflows/test.yml
```diff
@@ -160,7 +150,7 @@ jobs:
       - name: Install dependencies
         shell: bash -l {0}
         run: |
-          mamba install python=3.8 python-blosc lz4 --file conda.txt -c conda-forge
+          mamba install python-blosc lz4 -c conda-forge
```
charlesbluca, line 153: Not sure if we would want to pin `python-blosc` and `lz4` and include them in the `dev` environments

rajagurunath:
> It seems like we could get away with removing the `COPY` line and maybe using the YAML files for env setup there, but I'm not sure if it has a use that's not immediately obvious.

yeah, I am also not sure why we are copying `conda.txt`, But if we use `environment.yaml` file as you suggested, the script `prepare.sh` from dask container [here](https://github.com/dask/dask-docker/blob/main/base/prepare.sh)  can detect and make use of the YAML file I guess ( but not sure whether it will detect conda.txt )

https://github.com/dask/dask-docker/blob/main/base/prepare.sh#L5-L27

.github/workflows/test.yml
rajagurunath: Please forgive my ignorance, I am also not aware of these packages, probably dependency packages for some other packages, Not sure though! 🤔

charlesbluca:
rerun tests

tests/unit/test_utils.py
```diff
@@ -102,6 +104,9 @@ def test_exception_parsing():
     assert str(e) == "Message"
 
 
+@pytest.mark.skipif(
+    sys.platform == "win32", reason="JAVA_HOME will often be incorrect on Windows"
```
charlesbluca, line 108: This and the XGBoost test on Windows are the only two that seem to fail if we don't install our dependencies in the same shell as the JAR creation / running of tests.

IMO we should opt to skip or xfail these tests for now on Windows, as the requirements to get these passing are rather rigid (requiring us to force reinstall several packages and forgo package caching for most tests), and the tests themselves don't seem too critical anyways:

- the `JAVA_HOME` bug on Windows doesn't seem to cause any critical failures elsewhere
- as expressed in its warnings, xgboost/dask isn't officially supported on Windows

PR #240, opened 2021-09-16
Add gpuCI support

charlesbluca:
This PR adds:
- pytest marker `gpu`, which when added to tests causes them to be skipped by default unless `--rungpu` is provided
- tests for the GPU table creation added in #220 and #224 
- the build files used by gpuCI to run on any open PRs

### DIFF
--- a/tests/unit/test_context.py
+++ b/tests/unit/test_context.py
@@ -8,13 +8,21 @@ from pandas.testing import assert_frame_equal
 
 from dask_sql import Context
 
+try:
+    import cudf
+    import dask_cudf
+except ImportError:
+    cudf = None
+    dask_cudf = None
 
-def test_add_remove_tables():
+
+@pytest.mark.parametrize("gpu", [False, pytest.param(True, marks=pytest.mark.gpu)])
+def test_add_remove_tables(gpu):
     c = Context()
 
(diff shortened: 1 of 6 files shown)


tests/unit/test_context.py
```diff
@@ -72,84 +88,129 @@ def test_explain():
...
 
     data_frame = dd.from_pandas(pd.DataFrame({"a": [1, 2, 3]}), npartitions=1)
-    c.create_table("df", data_frame)
+    c.create_table("df", data_frame, gpu=gpu)
 
     result = c.sql("SELECT * FROM df")
-    assert isinstance(result, dd.DataFrame)
-    assert_frame_equal(result.compute(), data_frame.compute())
+    assert isinstance(result, dd.DataFrame if not gpu else dask_cudf.DataFrame)
+    dd.assert_eq(result, data_frame)
```
ayushdg, line 112: Similar to `assert_correct_input` it might be valuable wrapping something that does this as a more common utility for pytests in a followup PR.

ayushdg (approved):
Changes look good to me. Thanks for adding gpuCI to the project. 
Once updates are made upstream such that gpuCI starts running tests against PR's here we might explore codecov and remote the `#pragma no cover` in places they were added since we didn't test earlier. Eg: https://github.com/dask-contrib/dask-sql/pull/224

PR #245, opened 2021-10-05
Do not persist data to memory by default when creating tables

jdye64:
Currently data is persisted into memory when create_table is invoked. Instead this PR will set the default for persist to False which will prevent that from happening.

This Closes: dask-contrib#218

### DIFF
--- a/conftest.py
+++ b/conftest.py
@@ -1,3 +1,12 @@
 import pytest
 
 pytest_plugins = ["distributed.utils_test", "tests.integration.fixtures"]
+
+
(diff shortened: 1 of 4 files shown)


charlesbluca:
Now that #240 is merged, you should revert https://github.com/dask-contrib/dask-sql/pull/245/commits/efa492f505b3b2094692d8264655991a37c8074f, as it retains some bugs from the initial setup of gpuCI

jdye64:
rerun tests

Issue #225, opened 2021-08-25
Row-wise scalar UDFs


brandon-b-miller:
Hello, first off thanks for `dask-sql`!

I noticed (and please correct me if I am wrong) that the current process for running UDFs basically works by forwarding the relevant columns of a table directly to that python function as arguments. That is, if I have

```python
def f(x, y):
    return x + y
```

and I write

```
"SELECT f(col0, col1) FROM my_table"
```
The impl will pass `col0` and `col1` as dask series directly as arguments to `f`, at which point `f` simply runs a binop between those two actual series objects and returns a new series as a result. This implementation works great but has some limitations in the space of functions that a user can define when compared to a "row-wise" UDF. For instance one cannot write functions of this form:

```
def f(x, y):
    if x > 3: # here, type(x) is <class 'dask.dataframe.core.Series'>
        return y
    else:
        return x + 4

# ValueError: The truth value of a Series is ambiguous. Use a.any() or a.all().
```

I was wondering if there were any thoughts/plans around supporting true row-wise ops, where within the partitions each row is individually sent through the UDF as a set of scalars. Perhaps such an implementation could leverage `pd.DataFrame.apply` or similar.

Thanks!

brandon-b-miller:
Thank you both for your detailed replies. 

Allow me to address the performance consideration first as I think it constitutes a stronger motivation. Indeed, pandas does not make many performance promises when it comes to row wise UDFs that behave in this manner. However I believe part of what makes dask great in general is that it focuses on being a really great engine for distributing computation - and it tends to leave what happens within the partitions up to the "backend" library (pandas, cuDF), so long as it implements the right API. From that perspective I do not believe the onus should be on dask to protect the user from workflows that are slow because the pandas `apply` api is slow. `cuDF`'s `apply` function uses numba to jit cuda kernels which are relatively performant for example, even when compared to the logical equivalent using GPU arrays (`x=a+b+c+d+e` might be faster using `apply` than passing the vectors themselves, which will resolve in python to 4 separate binops I believe)

[...]

PR #246, opened 2021-10-05
Support pandas style row udfs

brandon-b-miller:
WIP. Closes https://github.com/dask-contrib/dask-sql/issues/225.

Lets users invoke functions that map onto `xyz.DataFrame.apply`, such as:

```python
  def f(row):
      if row['a'] is pd.NA:
          return 42
      return row['a'] + row['b']
```

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -252,6 +252,7 @@ class Context:
         return_type: type,
         replace: bool = False,
         schema_name: str = None,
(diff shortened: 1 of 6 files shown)


charlesbluca:
Looks like gpuCI is working 😄

brandon-b-miller:
@charlesbluca thanks for the review - I'll address this today and tomorrow :)

PR #247, opened 2021-10-06
Add issue templates

charlesbluca:
This PR adds issue templates roughly copying those used for cuDF - should hopefully make it easier to organize issues moving forward.


ayushdg:
There's a failing gpuCI check unrelated to this PR. That can be addressed separately

charlesbluca:
Yup - currently investigating that on #248, as I can't reproduce it locally

PR #251, opened 2021-10-11
Add support for GPU table creation in dask / location plugins

charlesbluca:
This allows us to stop xfailing some of the GPU tests for table creation


dask_sql/input_utils/location.py
```diff
@@ -23,20 +31,23 @@ def to_dc(
...
     ):
-
         if format == "memory":
             client = default_client()
-            return client.get_dataset(input_item, **kwargs)
+            df = client.get_dataset(input_item, **kwargs)
 
+            for plugin in [DaskInputPlugin(), PandasLikeInputPlugin()]:
+                if plugin.is_correct_input(df, table_name, format, **kwargs):
+                    return plugin.to_dc(df, table_name, format, gpu, **kwargs)
```
charlesbluca, line 40: This seems like the most sensible way to handle published datasets without duplicating code, but I am not sure if there's a cleaner way this could be done.

PR #255, opened 2021-10-13
Use Dask's `sort_values` for first column sorting in `apply_sort`

charlesbluca:
Now that Dask's `sort_values` has support for ascending/descending and null positioning, we can use it to do what `_sort_first_column` was once doing in a (hopefully) more performant way. There are still some failures here that are resolved with https://github.com/dask/dask/pull/8225, so this should be safe to merge by the next Dask release.

### DIFF
--- a/dask_sql/physical/utils/sort.py
+++ b/dask_sql/physical/utils/sort.py
@@ -58,16 +58,13 @@ def apply_sort(
         except ValueError:
             pass
 
-    # Split the first column. We need to handle this one with set_index
-    first_sort_column = sort_columns[0]
-    first_sort_ascending = sort_ascending[0]
-    first_null_first = sort_null_first[0]
-
-    # Only sort by first column first
-    # As sorting is rather expensive, we bether persist here
-    df = df.persist()
-    df = _sort_first_column(
-        df, first_sort_column, first_sort_ascending, first_null_first
+    # Dask doesn't natively support multi-column sorting;
+    # we work around this by initially sorting by the first
+    # column then handling the rest with `map_partitions`
+    df = df.sort_values(
+        by=sort_columns[0],
+        ascending=sort_ascending[0],
+        na_position="first" if sort_null_first[0] else "last",
     )
 
     # sort the remaining columns if given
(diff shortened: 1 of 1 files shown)


rjzamora:
>cc @rjzamora if you have any thoughts on this

I agree that much of the dask_cudf sort_values logic should live in upstream dask.dataframe. However, I will need to think a bit more about the `sort_function` idea to provide any useful feedback.  

**Somewhat-related thoughts**: One thing I like about the dask_cudf logic is that `sort_values` uses `rearrange_by_column` rather than `rearrange_by_divisions`.  This makes the mutli-column sorting code path a bit simpler, because we can use multiple columns to define the new (temporary) "_partitions" column, but then the rest of the logic is the same as a single-column sort. I also like that dask_cudf's set_index is directly based on sort_values (which is not the case in dask.dataframe).

charlesbluca:
Looks like the original fix I had in mind for dask/dask#8255 is actually needed here if we intend to use upstream Dask sorting functions (`rearrange_by_divisions`, `set_partitions_pre`) directly on a dask-cudf dataframe. However, I'm more interested in seeing if we can move most of the logic for sorting upstream, maybe by allowing for a custom sorting function in dask/dask-cudf's `sort_values`. This would allow for the same performance benefits of using the internal sorting logic instead of a `sort_values` call, while also keeping things pretty simple downstream.

In any case, I'm pretty sure using `sort_values` instead of `_sort_first_column` for the initial sort is still beneficial performance-wise, so we can merge this now and make a follow up depending on what direction we decide to move in.

dask_sql/physical/utils/sort.py
```diff
@@ -25,7 +25,7 @@ def apply_sort(
             by=sort_columns,
             ascending=sort_ascending,
             na_position="first" if sort_null_first[0] else "last",
-        )
+        ).persist()
```
charlesbluca, line 28: Noting that I added this persist call after sorting to match up with the other sort paths, which persist after sorting.

dask_sql/physical/utils/sort.py
```diff
@@ -37,34 +37,30 @@ def apply_sort(
         and not any(sort_null_first)
     ):
         try:
-            return df.sort_values(sort_columns, ignore_index=True)
+            return df.sort_values(sort_columns, ignore_index=True).persist()
```
charlesbluca, line 40: Same as above

PR #262, opened 2021-10-18
Simplify single-partition sorting logic

charlesbluca:
With https://github.com/rapidsai/cudf/pull/9455 merged, cuDF now has handling for lists of `ascending` bools _and_ `na_position`, making it consistent with Pandas.

### DIFF
--- a/dask_sql/physical/utils/sort.py
+++ b/dask_sql/physical/utils/sort.py
@@ -19,30 +19,13 @@ def apply_sort(
     sort_null_first: List[bool],
 ) -> dd.DataFrame:
     # if we have a single partition, we can sometimes sort with map_partitions
-    if df.npartitions == 1:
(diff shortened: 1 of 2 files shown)


charlesbluca:
After some more thought here, I actually think it makes sense to merge in changes that depend on nightly cuDF - at the moment, GPU support is still experimental and I feel like it should be expected to require the latest changes on cuDF's end.

Hopefully in the future we can relax this requirement and allow for stable versions (with performance being the only implication), but for now I think this should be good to merge as long as we also make sure to emphasize the need for cuDF nightlies in docs on GPU support.

quasiben:
> After some more thought here, I actually think it makes sense to merge in changes that depend on nightly cuDF - at the moment, GPU support is still experimental and I feel like it should be expected to require the latest changes on cuDF's end.

Agreed I think for cuDF/GPU support we can be quite liberal in our experimentation and communicate that to end users.

PR #276, opened 2021-10-21
Update register_function()

DaceT:
Added an example of how to use the replace parameter to overwrite a function with the same name.


GPUtester:

quasiben:
Thanks @DaceT .  I think this is failing due to a linting issue.  If you install the pre-commit hooks and run black you should be fine

Issue #270, opened 2021-10-20
[ENH] Allow updating UDFs


randerzander:
**Is your feature request related to a problem? Please describe.**
While developing UDF logic, it's common to update or tweak the code then test it live inside a query.

Currently you can only register a function w/ the same name once. If you update the UDF code and re-register you get:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_1172564/3008080374.py in <module>
     22         return c * r
     23 
---> 24 c.register_function(haversine_dist, "haversine_dist", [
     25                     ("source_lat", np.float64),
     26                     ("source_lng", np.float64),

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/context.py in register_function(self, f, name, parameters, return_type, replace, schema_name, row_udf)
    312 
    313         """
--> 314         self._register_callable(
    315             f,
    316             name,

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/context.py in _register_callable(self, f, name, aggregation, parameters, return_type, replace, schema_name, row_udf)
    830 
    831             elif schema.functions[lower_name] != f:
--> 832                 raise ValueError(
    833                     "Registering different functions with the same name is not allowed"
    834                 )

ValueError: Registering different functions with the same name is not allowed
```

**Describe the solution you'd like**
Ideally if you register a function w/ the same name a second time, Dask-SQL would "overwrite" the existing function and use the new code from then on.

**Describe alternatives you've considered**
As a workaround, you can iterate by adding a numbered version to the function name, but that gets tedious.

You could re-create a new Context but that means re-loading all your tables.

randerzander:
Wow, thanks for the suggestion. That works!

randerzander:
Probably linking to the API from the "Custom Functions and Aggregations" page is a good idea.

I can see why register_function doesn't allow replace by default, but it might be nice for the error message to suggest using `replace=True`.

PR #282, opened 2021-10-25
Update register_func() in context.py

DaceT:
Fixes dask-contrib/dask-sql#270


dask_sql/context.py
```diff
@@ -274,8 +274,10 @@ def register_function(
             SELECT f(x)
             FROM df
 
-        Please note that you can always only have one function with the same name;
-        no matter if it is an aggregation or scalar function.
+        Please keep in mind that you can only have one function with the same name,
+        regardless of whether it is an aggregation or a scalar function. However,
+        if you register two functions with the same name, the last one registered
+        will take precedence over the first one.
```
charlesbluca, line 280: Should we specify here that the precedence only applies if `replace=True`?

```
Please keep in mind that you can only have one function with the same name,
regardless of whether it is an aggregation or a scalar function. By default,
attempting to register two functions with the same name will raise an error;
setting `replace=True` will give precedence to the most recently registered
function.
```

dask_sql/context.py
```diff
@@ -298,13 +300,30 @@ def f(x):
...
+
         Args:
             f (:obj:`Callable`): The function to register
             name (:obj:`str`): Under which name should the new function be addressable in SQL
             parameters (:obj:`List[Tuple[str, type]]`): A list ot tuples of parameter name and parameter type.
                 Use `numpy dtypes <https://numpy.org/doc/stable/reference/arrays.dtypes.html>`_ if possible.
             return_type (:obj:`type`): The return type of the function
-            replace (:obj:`bool`): Do not raise an error if the function is already present
+            replace (:obj:`bool`): Raises an error if the function is already present. If you want to overwrite
+            a function, set the value to True; otherwise, the default value is False.
```
charlesbluca, line 326: I think the original arg description (do not raise an error if the function is already present) is still correct here; I think we might be better off here with something like:

```
replace (:obj:`bool`): If `True`, do not raise an error if a function with the same name is already
present; instead, replace the original function. Default is `False`.
```

dask_sql/context.py
```diff
@@ -829,7 +848,7 @@ def _register_callable(
 
             elif schema.functions[lower_name] != f:
                 raise ValueError(
-                    "Registering different functions with the same name is not allowed"
+                    'Registering multiple functions with the same name is only permitted when the parameter "replace=True" is used.'
```
charlesbluca, line 851: Think we should be good without quotes around the keyword argument here:

```suggestion
                    "Registering multiple functions with the same name is only permitted if replace=True"
```

dask_sql/context.py
DaceT: > Should we specify here that the precedence only applies if `replace=True`?
> 
> ```
> Please keep in mind that you can only have one function with the same name,
> regardless of whether it is an aggregation or a scalar function. By default,
> attempting to register two functions with the same name will raise an error;
> setting `replace=True` will give precedence to the most recently registered
> function.
> ```

Yes, we should; thanks for catching that!

charlesbluca (approved):
LGTM, thanks @DaceT!

Issue #278, opened 2021-10-25
[BUG] Row-based functions seem to be broken w/ Pandas tables


randerzander:
I'm trying to run the example row-wise UDF example in [the docs](https://dask-sql.readthedocs.io/en/latest/pages/custom.html#row-wise-pandas-udfs)
```
import pandas as pd
from dask_sql import Context

c = Context()

pdf = pd.DataFrame({'a': [0, 1], 'b': [2, 3]})

def f(row):
    return row['a'] + row['b']

# todo - fix the api
c.create_table('data', pdf)
c.register_function(f, "f", [], None, row_udf=True)
c.sql("SELECT f(a, b) FROM data")
```
Traceback:
```
---------------------------------------------------------------------------
KeyError                                  Traceback (most recent call last)
~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/mappings.py in python_to_sql_type(python_type)
     95     try:
---> 96         return _PYTHON_TO_SQL[python_type]
     97     except KeyError:  # pragma: no cover

KeyError: None

During handling of the above exception, another exception occurred:

NotImplementedError                       Traceback (most recent call last)
/tmp/ipykernel_682686/3931168022.py in <module>
     12 c.create_table('data', pdf)
     13 c.register_function(f, "f", [], None, row_udf=True)
---> 14 c.sql("SELECT f(a, b) FROM data")

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/context.py in sql(self, sql, return_futures, dataframes)
    429                 self.create_table(df_name, df)
    430 
--> 431         rel, select_names, _ = self._get_ral(sql)
    432 
    433         dc = RelConverter.convert(rel, context=self)

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/context.py in _get_ral(self, sql)
    709         """Helper function to turn the sql query into a relational algebra and resulting column names"""
    710         # get the schema of what we currently have registered
--> 711         schemas = self._prepare_schemas()
    712 
    713         # Now create a relational algebra from that

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/context.py in _prepare_schemas(self)
    676             for function_description in schema.function_lists:
    677                 name = function_description.name
--> 678                 sql_return_type = python_to_sql_type(function_description.return_type)
    679                 if function_description.aggregation:
    680                     logger.debug(f"Adding function '{name}' to schema as aggregation.")

[...]

PR #283, opened 2021-10-26
Require UDF return type and update docs

brandon-b-miller:
Closes https://github.com/dask-contrib/dask-sql/issues/278.

Small PR that fixes a docs TODO and adds an error guard.

### DIFF
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -195,6 +195,9 @@ class UDF:
         """
         self.row_udf = row_udf
         self.func = func
+
+        if return_type is None:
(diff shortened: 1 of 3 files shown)


quasiben:
@brandon-b-miller do you have thoughts on @ayushdg concerns over type inference in the above ?  I think this is the last concern and then we can merge

brandon-b-miller:
Updating this shortly.

brandon-b-miller:
Should fail in both cases now :)

quasiben:
Thanks @brandon-b-miller !

PR #289, opened 2021-11-02
Use unique names for null/non-null groupby columns

charlesbluca:
This should unblock #286 by ensuring that all groupby columns computed by `get_groupby_with_nulls_cols` have unique names.

cc @ayushdg


ayushdg (approved):
Thanks @charlesbluca for the fix. Can confirm that it unblocks my usecase and the changes lgtm!

PR #294, opened 2021-11-03
Prevent JVM Segfault

jdye64:
Under certain circumstances importing `dask_cuda` after any `dask_sql` imports will cause an underlying segfault to be triggered in the JVM. In tests thus far adjusting that import order in `context.py` prevents that from happening. Subsequent PRs may be required after further testing.

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -9,8 +9,8 @@ import pandas as pd
 from dask.base import optimize
 from dask.distributed import Client
 
-try
-    import dask_cuda
+try:
+    import dask_cuda # noqa: F401
 except: ImportError: # pragma: no cover
     pass
 
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -8,7 +8,11 @@ import dask.dataframe as dd
 import pandas as pd
 from dask.base import optimize
 from dask.distributed import Client
-import dask_cuda
+
+try
+    import dask_cuda
+except: ImportError: # pragma: no cover
+    pass
 
 from dask_sql import input_utils
 from dask_sql.datacontainer import (
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -11,7 +11,7 @@ from dask.distributed import Client
 
 try:
     import dask_cuda # noqa: F401
-except: ImportError: # pragma: no cover
+except ImportError: # pragma: no cover
     pass
 
 from dask_sql import input_utils
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -9,6 +9,11 @@ import pandas as pd
 from dask.base import optimize
 from dask.distributed import Client
 
+try:
+    import dask_cuda  # noqa: F401
+except ImportError:  # pragma: no cover
+    pass
+
 from dask_sql import input_utils
 from dask_sql.datacontainer import (
     UDF,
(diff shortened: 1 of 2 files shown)


jdye64:
I plan to add a pytest with some of the common occurrences where we were seeing the segfault occur so we can guard against those in the future. I'm just compiling that list offline but will post a pytest in the coming days.

jdye64:
rerun tests

charlesbluca:
Cool, looks like this triggers the segfault! Is there any way we can wrap this up in a GPU-only test?

jdye64:
> Cool, looks like this triggers the segfault! Is there any way we can wrap this up in a GPU-only test?

I think it is actually failing to find `dask_cuda` still

charlesbluca:
Checking the gpuCI run, it looks like dask-cuda is there in the conda list step, so I'm assuming it being imported after `dask_sql.Context` is what's causing the segfault here. However it is failing to find the package in the host tests since we don't install any GPU requirements there.

Would we need to add dask-cuda to the host test requirements, or is there a way to contain the conditions for this segfault (i.e. the dask-cuda import) in a test that can be disabled on host?

jdye64:
Ok good news. So I commented out the fix in [commit](https://github.com/dask-contrib/dask-sql/pull/294/commits/ee5162fb4382133b86badc519fb658a4531cb990) and CI failed. This is actually good. Then I un-commented the fix out again in [commit](https://github.com/dask-contrib/dask-sql/pull/294/commits/1ba9a26022126136138ec27f3760291efc8db3af) and it works. Therefore I am satisfied this at least solves the segfault issue we are interested in.

dask_sql/context.py
```diff
@@ -34,6 +34,13 @@
 from dask_sql.physical.rex import RexConverter, core
 from dask_sql.utils import ParsingException
 
+# try:
+#    import dask_cuda  # noqa: F401
+# except ImportError:  # pragma: no cover
+#    pass
+#
```
ayushdg, line 41: I believe we need to re-add this back prior to jvm startup which happens as a part of line 12.

ayushdg (changes requested):

tests/integration/fixtures.py
```diff
@@ -14,6 +14,16 @@
...
     cudf = None
 
+try:
+    import dask_cudf
+except ImportError:
+    dask_cudf = None
+
+try:
+    from dask_cuda import LocalCUDACluster  # noqa: F401
+except ImportError:
```
charlesbluca, line 24: Nitpick but can these imports be placed in a specific test? Something like

```python
def test_jvm_segfault():
    import dask_cudf
    from dask_cuda import LocalCUDACluster
```

If they need to be imported here to trigger the segfault, then can they be consolidated with the above try-except block with a comment noting why they appear?

```python
try:
    import cudf
    # importing to check for JVM segfault
    import dask_cudf
    from dask_cuda import LocalCUDACluster
except ImportError:
    cudf = None
```

dask_sql/context.py
jdye64: Yep, I had forgotten to add it back after testing that commenting it out actually caused errors

tests/integration/fixtures.py
jdye64: @charlesbluca yes, they needed to be imported there to properly test for the failure. I will combine them however.

dask_sql/context.py
```diff
@@ -34,6 +34,12 @@
 from dask_sql.physical.rex import RexConverter, core
 from dask_sql.utils import ParsingException
 
+try:
+    import dask_cuda  # noqa: F401
+except ImportError:  # pragma: no cover
+    pass
+
+
```
ayushdg, line 42: These need to be moved prior to line 12 (the first dask_sql import) which ends up starting the jvm.

ayushdg (changes requested):

dask_sql/context.py
jdye64: yep, good catch. I got in a hurry and forgot I had changed the ordering for testing.

Issue #296, opened 2021-11-03
[BUG] Selecting based on dates unexpectedly returns no results.


ChrisJar:
**What happened**:
When selecting rows based on a set of dates, dask-sql returns no results even though there are matches

**Minimal Complete Verifiable Example**:

For example:
```python
import dask.dataframe as dd
import pandas as pd
from dask_sql import Context

df = pd.DataFrame({'d_date':["2001-08-01", "2001-08-02", "2001-08-03"], 'val':[1,2,3]})
ddf = dd.from_pandas(df, npartitions=1)

c = Context()
c.create_table("df", ddf)

query = "SELECT val, d_date FROM df WHERE CAST(d_date as date) IN (date '2001-08-01', date '2001-08-03')"
c.sql(query).compute()
```
returns:
```
Empty DataFrame
Columns: [val, d_date]
Index: []
```
This also occurs when the input column is cast to datetime:
```python
import dask.dataframe as dd
import pandas as pd
from dask_sql import Context

df = pd.DataFrame({'d_date':["2001-08-01", "2001-08-02", "2001-08-03"], 'val':[1,2,3]})
df['d_date'] = pd.to_datetime(df['d_date'], format='%Y-%m-%d')
ddf = dd.from_pandas(df, npartitions=1)

c = Context()
c.create_table("df", ddf)

query = "SELECT val, d_date FROM df WHERE CAST(d_date as date) IN (date '2001-08-01', date '2001-08-03')"
c.sql(query).compute()
```
However: this does not occur when using GPU dataframes. For example:
```python
import dask_cudf
import cudf
from dask_sql import Context

df = cudf.DataFrame({'d_date':["2001-08-01", "2001-08-02", "2001-08-03"], 'val':[1,2,3]})
ddf = dask_cudf.from_cudf(df, npartitions=1)

c = Context()
c.create_table("df", ddf)

query = "SELECT val, d_date FROM df WHERE CAST(d_date as date) IN (date '2001-08-01', date '2001-08-03')"
c.sql(query).compute()
```
returns
```
   val      d_date
0    1  2001-08-01
2    3  2001-08-03
```

**Environment**:

- dask-sql version: 0.3.10
- Python version: 3.8
- Operating System: Ubuntu 18.04
- Install method (conda, pip, source): Source

rajagurunath:
Hi @ChrisJar,

Thanks for writing this detailed Description and easy-to-reproduce code snippets, really appreciate it!

[...]

PR #298, opened 2021-11-04
Fix for date format mismatch

rajagurunath:
Initial attempt to solve: https://github.com/dask-contrib/dask-sql/issues/296

Explicitly handled the type conversion for the `date` column and a literal value. And wondering if there is any better way to handle this datetime casting. Happy to know everyone's thoughts on this!

### DIFF
--- a/dask_sql/mappings.py
+++ b/dask_sql/mappings.py
@@ -163,7 +163,8 @@ def sql_to_python_value(sql_type: str, literal_value: Any) -> Any:
         dt = datetime.fromtimestamp(
             int(literal_value.getTimeInMillis()) / 1000, timezone.utc
         )
-
+        if sql_type == "DATE":
+            return dt.date()
         return dt
 
     elif sql_type.startswith("DECIMAL("):
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -196,14 +196,16 @@ class CastOperation(Operation):
         if not is_frame(operand):
             return operand
 
-        output_type = str(rex.getType())
-        output_type = sql_to_python_type(output_type.upper())
-
-        return_column = cast_column_to_type(operand, output_type)
+        sql_output_type = str(rex.getType())
+        python_output_type = sql_to_python_type(sql_output_type.upper())
+        return_column = cast_column_to_type(operand, python_output_type)
 
         if return_column is None:
             return operand
         else:
+            # handle datetime type specially
+            if sql_output_type == "DATE":
+                return_column = return_column.dt.date
             return return_column
 
 
--- a/tests/integration/test_select.py
+++ b/tests/integration/test_select.py
@@ -115,3 +115,13 @@ def test_timezones(c, datetime_table):
     result_df = result_df.compute()
 
     assert_frame_equal(result_df, datetime_table)
+
+
+def test_date_casting(c, datetime_table):
+    # check date casting
+    query = "SELECT cast(timezone as date) as date1,cast(utc_timezone as date) as date2 FROM datetime_table "
+    result_df = c.sql(query).compute().astype(str)
+    expected_df = pd.DataFrame(
+        {"date1": ["2014-08-01"] * 3, "date2": ["2014-08-01"] * 3}
+    )
+    assert_frame_equal(result_df, expected_df, check_dtype=False)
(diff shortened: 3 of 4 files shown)


quasiben:
This looks like a reasonable solution.  Any ideas why cuDF works while Pandas fails ?

rajagurunath:
Hi @quasiben,

Thanks for reviewing the PR and putting me in the right direction (I missed the part about why it works for CUDF)

I tried digging deep into this and found the following, Both pandas and cudf seem to work for the local timezone. but pandas doesn't seem to work with UTC timezone.


while executing the query, we are converting the literal timestamp value to the datetime value using the following function.

https://github.com/dask-contrib/dask-sql/blob/61cbca951bb4b62639ffcba43034f1dac392e24e/dask_sql/mappings.py#L160-L165

example using the above function:
image

Since we are only accepting UTC timezone currently, we may need to do the casting explicitly, what do you think? 

And One more point in the favour of handling date casting explicitly is the current version of code includes `hh:MM: SS` even after casting the `datetime` column into a `date` column format (for both cudf and pandas version), i.e following test case seems to fail in the current version of the dask-sql (without date casting). 
https://github.com/dask-contrib/dask-sql/blob/61cbca951bb4b62639ffcba43034f1dac392e24e/tests/integration/test_select.py#L120-L127

Open for other ideas/Suggestions from everyone!

quasiben:
Thank for digging into this @rajagurunath !  I didn't know dask-sql sets everything as UTC if TZ is not defined.  
1) We should make this more clear in documentation
2) If the user casts a column to `date` we _could_ also set the column in UTC:
> ddf['d_date'] = ddf.d_date.dt.tz_localize('UTC')

I suspect this is slow than what you have.

dask_sql/mappings.py
```diff
@@ -163,7 +163,8 @@ def sql_to_python_value(sql_type: str, literal_value: Any) -> Any:
         dt = datetime.fromtimestamp(
             int(literal_value.getTimeInMillis()) / 1000, timezone.utc
         )
-
+        if sql_type == "DATE":
```
charlesbluca, line 166: Would you mind adding a comment here linking to https://github.com/dask-contrib/dask-sql/issues/296?

```suggestion
        # address datetime format mismatch
        # https://github.com/dask-contrib/dask-sql/issues/296
        if sql_type == "DATE":
```

dask_sql/physical/rex/core/call.py
```diff
@@ -196,14 +196,16 @@ def cast(self, operand, rex=None) -> SeriesOrScalar:
...
-
-        return_column = cast_column_to_type(operand, output_type)
+        sql_output_type = str(rex.getType())
+        python_output_type = sql_to_python_type(sql_output_type.upper())
+        return_column = cast_column_to_type(operand, python_output_type)
 
         if return_column is None:
             return operand
         else:
+            # handle datetime type specially
```
charlesbluca, line 206: Same as above - just adding a link to https://github.com/dask-contrib/dask-sql/issues/296

```suggestion
            # address datetime format mismatch
            # https://github.com/dask-contrib/dask-sql/issues/296
```

tests/integration/test_select.py
```diff
@@ -115,3 +115,13 @@ def test_timezones(c, datetime_table):
     result_df = result_df.compute()
 
     assert_frame_equal(result_df, datetime_table)
+
+
+def test_date_casting(c, datetime_table):
```
charlesbluca, line 120: Thinks we don't actually need `datetime_table` here? Since we only use the context here

```suggestion
def test_date_casting(c):
```

charlesbluca (commented):
Thanks for the work here @rajagurunath 😄 just a few comments:

ChrisJar:
@rajagurunath Thanks for working on this! One thing I noticed was that when I perform a date comparison with this PR:
```python
import dask.dataframe as dd
import pandas as pd
from dask_sql import Context

df = pd.DataFrame({'d_date':["2001-08-01", "2001-08-02", "2001-08-03"], 'val':[1,2,3]})
ddf = dd.from_pandas(df, npartitions=1)

c = Context()
c.create_table("df", ddf)

query = "SELECT val, d_date FROM df WHERE CAST(d_date as date) >= DATE '2001-08-02'"
c.sql(query).compute()
```
I get:
```
TypeError: '>=' not supported between instances of 'str' and 'datetime.date'
```
When I run the same code without this PR I get:
```
TypeError('Invalid comparison between dtype=datetime64[ns] and datetime')
```
Do you know what might be happening here?

charlesbluca:
Working on a PR to supersede this one in #343

PR #308, opened 2021-11-11
Update nightly recipe / setup for `2021.11.0` release

charlesbluca:
As part of the new release, we made some modifications to the conda recipe using [grayskull](https://github.com/conda-incubator/grayskull), so that it now generally matches up with the setup.py. This PR reflects some of those changes on the nightly recipe, and also makes some updates to the setup.py so that it is consistent with the desired conda recipe.


continuous_integration/recipe/meta.yaml
```diff
@@ -11,8 +11,12 @@ source:
   git_url: ../..
 
 build:
-  noarch: python
   number: {{ GIT_DESCRIBE_NUMBER }}
+  skip: true   # [py2k]
+  entry_points:
+    - dask-sql-server = dask_sql.server.app:main
+    - dask-sql = dask_sql.cmd:main
+  noarch: python
```
charlesbluca, line 19: Should we still build nightlies `noarch` even though the stable releases don't do this?

setup.py
```diff
@@ -61,7 +61,7 @@ def run(self):
 
 setup(
     name="dask_sql",
-    description="Dask SQL",
+    description="SQL query layer for Dask",
```
charlesbluca, line 64: Just noting that this change maps to the conda recipe's `summary`.

setup.py
```diff
@@ -74,7 +74,7 @@ def run(self):
     python_requires=">=3.6",
     setup_requires=["setuptools_scm"] + sphinx_requirements,
     install_requires=[
-        "dask[dataframe,distributed]==2021.10.0",
+        "dask[dataframe,distributed]>=2021.10.0",
```
charlesbluca, line 77: We'll need this if we want to use unpinned versions of Dask with the nightlies for certain features (`dask-sql-server` in particular)

continuous_integration/recipe/meta.yaml
jakirkham: No this should be dropped. It doesn't work with the selector below

continuous_integration/recipe/meta.yaml
```diff
@@ -21,30 +25,35 @@ requirements:
...
     - tzlocal >=2.1
-    - prompt_toolkit >=3.0.8
-    - pygments >=2.7.3
-    - nest-asyncio >=1.0.0
-    - tabulate >=0.8.9
+    - prompt-toolkit
+    - pygments
+    - nest-asyncio
+    - tabulate
+    - importlib-metadata  # [py<38]
```
jakirkham, line 43: This is the selector that makes `noarch` not an option

jakirkham (commented):
Given the recipe changes, `noarch` needs to be dropped

continuous_integration/recipe/meta.yaml
jakirkham: Will need to use [variants]( https://docs.conda.io/projects/conda-build/en/latest/resources/variants.html ) to construct a build matrix for the Python versions

No easy answers for OS other than including more jobs on other OSes

Am assuming we are the primary users of nightlies. So would use that as a guiding principle until someone else asks for something more

charlesbluca: Thanks for the link - I'll play around with variants for now, and check in with RAPIDS folk to see if we have any specific needs here.

continuous_integration/recipe/conda_build_config.yaml
```diff
@@ -0,0 +1,4 @@
+py_version:
```
jakirkham, line 1:
```suggestion
python:
```

continuous_integration/recipe/meta.yaml
```diff
@@ -17,15 +16,15 @@ build:
...
+  string: py{{ py_version.replace(".", "") }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
   script: {{ PYTHON }} -m pip install . --no-deps -vv
 
 requirements:
   build:
     - maven >=3.6.0
   host:
     - pip
-    - python
+    - python {{ py_version }}
```
jakirkham, line 26:
```suggestion
    - python
```

jakirkham (commented):
This should just work if the variant name share the package name

continuous_integration/recipe/meta.yaml
```diff
@@ -11,40 +10,48 @@ source:
...
 
 build:
-  noarch: python
   number: {{ GIT_DESCRIBE_NUMBER }}
-  string: py{{ py_version }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+  skip: true   # [py2k]
+  entry_points:
+    - dask-sql-server = dask_sql.server.app:main
+    - dask-sql = dask_sql.cmd:main
+  string: py{{ py_version.replace(".", "") }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```
charlesbluca, line 18: Forgot about this - is there a clean way to get a Python version for the package string from the build variant?

continuous_integration/recipe/meta.yaml
```diff
@@ -1,6 +1,5 @@
 {% set name = "dask-sql" %}
 {% set version = environ.get('GIT_DESCRIBE_TAG', '0.0.0.dev') + environ.get('VERSION_SUFFIX', '') %}
-{% set py_version=environ.get('CONDA_PY', 36) %}
```
charlesbluca, line 3: If not will probably need to add this back in and modify

charlesbluca (commented):
Missed something

continuous_integration/recipe/meta.yaml
jakirkham: Can just replace with `python` instead of `py_version`. Though maybe you have already done this :)

continuous_integration/recipe/meta.yaml
```diff
@@ -11,40 +10,48 @@ source:
...
 
 build:
-  noarch: python
   number: {{ GIT_DESCRIBE_NUMBER }}
-  string: py{{ py_version }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
+  skip: true   # [py2k]
+  entry_points:
+    - dask-sql-server = dask_sql.server.app:main
+    - dask-sql = dask_sql.cmd:main
+  string: py{{ python }}{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```
jakirkham, line 18: Sorry will need to get rid of the `.` as well

```suggestion
  string: py{{ python.replace(".", "") }}{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
```

charlesbluca:
Maybe we need to do the `replace` in a `set` template at the top of the file?

jakirkham:
Shouldn't be needed

charlesbluca:
Think the issue might be that we need to use a replace filter here? If these are jinja2 templates that is

jakirkham:
Yeah saw the latest commit. That seems like what is needed

continuous_integration/recipe/meta.yaml
```diff
@@ -11,40 +10,48 @@ source:
...
     - pip check
+    - dask-sql-server --help
+    - dask-sql --help
   requires:
     - pip
 
 about:
-  home: https://github.com/dask-contrib/dask-sql
-  summary: SQL Query Layer for dask
+  home: http://github.com/dask-contrib/dask-sql/
```
jakirkham, line 54: Should this be `https`?

```suggestion
  home: https://github.com/dask-contrib/dask-sql/
```

setup.py
```diff
@@ -61,7 +61,7 @@ def run(self):
 
 setup(
     name="dask_sql",
-    description="Dask SQL",
+    description="SQL query layer for Dask",
     url="http://github.com/dask-contrib/dask-sql/",
```
jakirkham, line 65: Also wondering if this should be `https`

```suggestion
    url="https://github.com/dask-contrib/dask-sql/",
```

jakirkham (commented):
Couple minor suggestions

charlesbluca:
Looks like the conda build job takes roughly twice as long, which isn't as bad as I was expecting. I think this should be good to merge in pending tests, and we can follow up with a different option if other OSes are desired (could we do MacOS / Windows builds on gpuCI?)

If timing ever becomes an issue, I wonder if it's possible / safe to only build the package for one Python version on PRs

jakirkham (approved):

jakirkham:
Thanks Charles! 😄

jakirkham:
Yeah to your point above about build time. We could separate out the matrix onto multiple CI jobs. This is effectively what conda-forge does. It contains a bunch of [variants]( https://github.com/conda-forge/dask-sql-feedstock/tree/305f80de4c87b78f845630b1ac86fc9bc9198bda/.ci_support ) and builds each one on a separate CI job. It's certainly something we could pursue if it becomes an issue.

Big picture, think it really all depends how long we maintain the Python 3.7 build here. Guessing we will want to add 3.9 & 3.10, which would increase the build matrix and may motivating dropping older Python versions. Dask (and others) may also choose to drop 3.7 when adding 3.10, which would push us in the same direction. Once 3.7 is dropped we don't really need the `importlib-metadata` dependency and can return to `noarch` (assuming we don't add any other Python version specific dependencies). At which point we don't really need to worry about more sophisticated builds.

Issue #279, opened 2021-10-25
[ENH] Support passing scalar args to UDFs


randerzander:
**Is your feature request related to a problem? Please describe.**
I'd like to pass non-Dataframe and Series arguments to UDFs.

**Describe the solution you'd like**
```
import pandas as pd

pdf = pd.DataFrame({'val': [0, 1, 2]})

def func(x, A, B):
  return x['val'] + A - B

pdf.apply(func, axis=1, args=(1, 2))
# result
0   -1
1    0
2    1
dtype: int64

# attempting to call from Dask-SQL:
c.create_table('pdf', pdf)
c.register_function(func, 'func', [
    ("val", np.int32),
    ("A", np.int32),
    ("B", np.int32)],
    np.int32,
    row_udf=True, replace=True)
c.sql("select func(val, 1, 2) from pdf")
```
Trace:
```
---------------------------------------------------------------------------
AttributeError                            Traceback (most recent call last)
/tmp/ipykernel_862152/1486820537.py in <module>
----> 1 c.sql("select func(val, 1, 2) from pdf")

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/context.py in sql(self, sql, return_futures, dataframes)
    431         rel, select_names, _ = self._get_ral(sql)
    432 
--> 433         dc = RelConverter.convert(rel, context=self)
    434 
    435         if dc is None:

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/physical/rel/convert.py in convert(cls, rel, context)
     58             f"Processing REL {rel} using {plugin_instance.__class__.__name__}..."
     59         )
---> 60         df = plugin_instance.convert(rel, context=context)
     61         logger.debug(f"Processed REL {rel} into {LoggableDataFrame(df)}")
     62         return df

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/physical/rel/logical/project.py in convert(self, rel, context)
     53             else:
     54                 random_name = new_temporary_column(df)
---> 55                 new_columns[random_name] = RexConverter.convert(
     56                     expr, dc, context=context
     57                 )

~/conda/envs/dsql/lib/python3.8/site-packages/dask_sql/physical/rex/convert.py in convert(cls, rex, dc, context)
     64         )
     65 
---> 66         df = plugin_instance.convert(rex, dc, context=context)
     67         logger.debug(f"Processed REX {rex} into {LoggableDataFrame(df)}")
     68         return df

[...]

PR #311, opened 2021-11-11
Row UDF scalar arguments

brandon-b-miller:
Closes https://github.com/dask-contrib/dask-sql/issues/279

### DIFF
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -212,7 +212,7 @@ class UDF:
                     column_args.append(operand)
                 else:
                     scalar_args.append(operand)
-                df = column_args[0].to_frame()
+            df = column_args[0].to_frame()
             for col in column_args[1:]:
                 df[col.name] = col
             result = df.apply(
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -205,10 +205,19 @@ class UDF:
 
     def __call__(self, *args, **kwargs):
         if self.row_udf:
-            df = args[0].to_frame()
-            for operand in args[1:]:
-                df[operand.name] = operand
-            result = df.apply(self.func, axis=1, meta=self.meta).astype(self.meta[1])
+            column_args = []
+            scalar_args = []
+            for operand in args:
+                if isinstance(operand, dd.Series):
+                    column_args.append(operand)
+                else:
+                    scalar_args.append(operand)
+                df = column_args[0].to_frame()
+                for col in column_args[1:]:
+                    df[col.name] = operand
+            result = df.apply(
+                self.func, axis=1, args=tuple(scalar_args), meta=self.meta
+            ).astype(self.meta[1])
         else:
             result = self.func(*args, **kwargs)
         return result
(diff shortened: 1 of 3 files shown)


brandon-b-miller:
This is ready for review - I thought some code might need to be written to disambiguate when one has a column with a name that could also be a constant, like having a column with name `42` and also passing 42 as a constant parameter. But it turns out we can just write `\"42\"` in the query to get the column with that name instead.

dask_sql/datacontainer.py
```diff
@@ -205,10 +205,19 @@ def __init__(self, func, row_udf: bool, return_type=None):
...
+            column_args = []
+            scalar_args = []
+            for operand in args:
+                if isinstance(operand, dd.Series):
+                    column_args.append(operand)
+                else:
+                    scalar_args.append(operand)
+                df = column_args[0].to_frame()
+                for col in column_args[1:]:
+                    df[col.name] = operand
```
ayushdg, line 217:
```suggestion
                    df[col.name] = col
```

ayushdg, line 217: I'm guessing this section was meant to be outside the loop.

ayushdg (changes requested):
Thanks a lot for the changes @brandon-b-miller !
Couple of minor comments in addition to the changes suggested:
Would it be possible to also update the docstring to include an example with scalar values?

Also I was able to catch the issues below while testing this with udfs that accept multiple column args, since in the single arg case the output would be correct. Might be worth adding that case to the tests as well.

brandon-b-miller:
Updated! :)

tests/integration/test_function.py
```diff
@@ -63,6 +65,64 @@ def f(row):
...
+
+    statement = f"SELECT F(a, {k}) as a from df"
+
+    return_df = c.sql(statement)
+    return_df = return_df.compute()
+    expectation = op(df[["a"]], k).astype(retty)
+    assert_frame_equal(return_df.reset_index(drop=True), expectation)
+
+
+# Test row UDFs with one args
```
charlesbluca, line 92:
```suggestion
# Test row UDFs with two args
```

tests/integration/test_function.py
```diff
@@ -63,6 +65,64 @@ def f(row):
     assert_frame_equal(return_df.reset_index(drop=True), expectation)
 
 
+# Test row UDFs with one args
```
charlesbluca, line 68:
```suggestion
# Test row UDFs with one arg
```

charlesbluca (commented):
Thanks for the work here @brandon-b-miller 🙂 just small nitpicks, the code itself LGTM

dask_sql/datacontainer.py
```diff
@@ -205,10 +205,19 @@ def __init__(self, func, row_udf: bool, return_type=None):
...
-                df[operand.name] = operand
-            result = df.apply(self.func, axis=1, meta=self.meta).astype(self.meta[1])
+            column_args = []
+            scalar_args = []
+            for operand in args:
+                if isinstance(operand, dd.Series):
+                    column_args.append(operand)
+                else:
+                    scalar_args.append(operand)
+                df = column_args[0].to_frame()
```
ayushdg, line 215:
```suggestion
            df = column_args[0].to_frame()
```

ayushdg (approved):
One small suggestion otherwise lgtm!

charlesbluca (approved):
LGTM; thanks for the work @brandon-b-miller 😄

Issue #310, opened 2021-11-11
[BUG] dask-sql-server + Superset not working for GPU tables


charlesbluca:
**What happened**:
It is relatively trivial to connect dask-sql's Presto server to Apache Superset for some basic visualization. Things seem to be working fine for Pandas-backed tables, but when attempting to run any queries with cuDF-backed tables we run into a `TypeError`:

```python
Traceback (most recent call last):
  File "/raid/charlesb/dev/rapids/compose/etc/conda/cuda_11.2/envs/rapids/lib/python3.8/site-packages/uvicorn/protocols/http/h11_impl.py", line 373, in run_asgi
    result = await app(self.scope, self.receive, self.send)
  File "/raid/charlesb/dev/rapids/compose/etc/conda/cuda_11.2/envs/rapids/lib/python3.8/site-packages/uvicorn/middleware/proxy_headers.py", line 75, in __call__
    return await self.app(scope, receive, send)
  File "/raid/charlesb/dev/rapids/compose/etc/conda/cuda_11.2/envs/rapids/lib/python3.8/site-packages/fastapi/applications.py", line 208, in __call__
    await super().__call__(scope, receive, send)
  File "/raid/charlesb/dev/rapids/compose/etc/conda/cuda_11.2/envs/rapids/lib/python3.8/site-packages/starlette/applications.py", line 112, in __call__
    await self.middleware_stack(scope, receive, send)
  File "/raid/charlesb/dev/rapids/compose/etc/conda/cuda_11.2/envs/rapids/lib/python3.8/site-packages/starlette/middleware/errors.py", line 181, in __call__
    raise exc
  File "/raid/charlesb/dev/rapids/compose/etc/conda/cuda_11.2/envs/rapids/lib/python3.8/site-packages/starlette/middleware/errors.py", line 159, in __call__
    await self.app(scope, receive, _send)
  File "/raid/charlesb/dev/rapids/compose/etc/conda/cuda_11.2/envs/rapids/lib/python3.8/site-packages/starlette/exceptions.py", line 82, in __call__
    raise exc
  File "/raid/charlesb/dev/rapids/compose/etc/conda/cuda_11.2/envs/rapids/lib/python3.8/site-packages/starlette/exceptions.py", line 71, in __call__
    await self.app(scope, receive, sender)
[...]

PR #312, opened 2021-11-12
Add handling for cuDF-backed tables in `dask-sql-server`

charlesbluca:
This should unblock failures when serving cuDF-backed tables using `dask-sql-server`, but is not really ideal as it forces a host synchronization whenever we need to return a query result; a more ideal long term solution would be to explore server options that allow us to return results per-column instead of per-row (i.e. iterating over columns instead of rows, which can't be done in cuDF).

cc @bryevdv this should resolve the failures we discussed and allow for some deeper exploration of Superset

xref #310

### DIFF
--- a/dask_sql/server/responses.py
+++ b/dask_sql/server/responses.py
@@ -77,6 +77,8 @@ class DataResults(QueryResults):
 
     @staticmethod
     def get_data_description(df):
+        if hasattr(df, "to_pandas"):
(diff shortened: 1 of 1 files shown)


randerzander (approved):
In my poking at this months back, this is exactly what I ended up trying that worked.

I looked into the Presto client docs a bit, but couldn't find much in the way of supported result formats.

PR #316, opened 2021-11-13
Ignore case for queries in the parser configuration

### LINKED ISSUE #315: [ENH] Support user control over case sensitivity
Some SQL query generators make it difficult to control the upper/lower cases of columns, tables, etc in emitted SQL.

In my use case, I'm trying to avoid modifying a query that uses uppercase column names, where my table was defined w/ lower case column names. It'd be useful to allow a configuration of case sensitivity used in the SQL parsing step. It [looks like](https://stackoverflow.com/questions/55385516/is-it-possible-to-configure-apache-calcite-identifiers-to-be-case-insensitive) this is doable in Calcite.

Test:
```
from dask_sql import Context
import pandas as pd

c = Context()
df = pd.DataFrame({'id': [0, 1]})

c.create_table('test', df)
c.sql("select ID from test")
```
Trace:
```
ParsingException                          Traceback (most recent call last)
/tmp/ipykernel_574321/308482520.py in <module>
      4 df = cudf.DataFrame({'id': [0, 1]})
      5 c.create_table('test', df)
----> 6 c.sql("select ID from test")

~/conda/envs/dsql-dask-main/lib/python3.8/site-packages/dask_sql/context.py in sql(self, sql, return_futures, dataframes)
    434                 self.create_table(df_name, df)
    435 
--> 436         rel, select_names, _ = self._get_ral(sql)
    437 
    438         dc = RelConverter.convert(rel, context=self)

~/conda/envs/dsql-dask-main/lib/python3.8/site-packages/dask_sql/context.py in _get_ral(self, sql)
    772             # if debug is not set.
    773             # Instead, we raise a nice exception
--> 774             raise ParsingException(sql, str(e.message())) from None
    775 
    776         # Internal, temporary results of calcite are sometimes

ParsingException: Can not parse the given SQL: org.apache.calcite.runtime.CalciteContextException: From line 1, column 8 to line 1, column 9: Column 'ID' not found in any table; did you mean 'id'?

The problem is probably somewhere here:

	select ID from test
	       ^^
```

jdye64:
This PR modifies the parser configuration to ensure that case does not matter across queries. This closes #315


jdye64:
One of the tests timed out. Running again. rerun tests

ayushdg:
Do you think that this could use the config options in #286 to expose this as something that can be exposed from the python layer to the java layer. Config options doesn't do this today but other parts of the code such as `create_model` or `create_experiment` allow this. The only difference in those cases is that they are implemented as a custom plugin. I'm not sure how the options could be piped to an existing plugin

jdye64:
> Do you think that this could use the config options in #286 to expose this as something that can be exposed from the python layer to the java layer. Config options doesn't do this today but other parts of the code such as `create_model` or `create_experiment` allow this. The only difference in those cases is that they are implemented as a custom plugin. I'm not sure how the options could be piped to an existing plugin

Sure, I just made the needed changes for that to happen. I just hard coded it to `False` for now but once #286 merges I will make another update to use that higher ConfigurationContainer abstraction.

dask_sql/context.py
```diff
@@ -801,7 +801,18 @@ def _get_ral(self, sql):
         schemas = self._prepare_schemas()
 
         # Now create a relational algebra from that
-        generator_builder = RelationalAlgebraGeneratorBuilder(self.schema_name)
+        ignore_case = self.schema[self.schema_name].config.get_config_by_prefix(
+            "dask.sql.identifier.ignore_case"
```
jdye64, line 805: I am following existing logic here using magic strings but think we should likely move all of the configuration keys to `mappings.py` and add further documentation in a subsequent PR. I am happy to make that PR as well once this one merges. Would prefer to not make that change in this PR since that change effects other codebase areas as well.

jdye64:
rerun tests

dask_sql/context.py
```diff
@@ -800,8 +800,19 @@ def _get_ral(self, sql):
...
-        generator_builder = RelationalAlgebraGeneratorBuilder(self.schema_name)
+        # True if the SQL query should be case sensitive and False otherwise
+        case_sensitive = self.schema[self.schema_name].config.get_config_by_prefix(
+            "dask.sql.identifier.case.sensitive"
+        )
+        case_sensitive = (
+            False
+            if case_sensitive == {}
+            else case_sensitive["dask.sql.identifier.case.sensitive"]
+        )
```
charlesbluca, line 811: Minor simplification:

```suggestion
        case_sensitive = self.schema[self.schema_name].config.get_config_by_prefix(
            "dask.sql.identifier.case.sensitive"
        ).get("dask.sql.identifier.case.sensitive", False)
```

tests/integration/test_select.py
```diff
@@ -80,6 +80,7 @@ def test_select_of_select(c, df):
 
 
 def test_select_of_select_with_casing(c, df):
+    c.set_config(("dask.sql.identifier.case.sensitive", True))
```
charlesbluca, line 83: Was the default behavior of dask-sql before to have case sensitivity be set true? In that case, should that be the option we default to if it's left unspecified? Right now it looks like we would default to false.

tests/integration/test_compatibility.py
```diff
@@ -934,3 +935,16 @@ def test_integration_1():
...
+    c = Context()
+    c.set_config(("dask.sql.identifier.case.sensitive", False))
+    df = pd.DataFrame({"id": [0, 1]})
+
+    c.create_table("test", df)
+
+    try:
+        c.sql("select ID from test")
+    except ParsingException as pe:
+        assert False, f"Queries should be case in-sensitve but raised exception {pe}"
```
charlesbluca, line 950:
```suggestion
        assert False, f"Queries should be case insensitive but raised exception {pe}"
```

dask_sql/context.py
charlesbluca: Also note that depending on the answer to https://github.com/dask-contrib/dask-sql/pull/316#discussion_r764347926 we might want to change the default in the `get` to True.

tests/integration/test_select.py
jdye64: Yeah, probably a good idea

jdye64:
@charlesbluca not sure what is up with CI but seems like everything passed??

charlesbluca:
Looks like gpuCI got hung up on the Docker image pull and timed out; there was an issue opened about this recently in Dask: https://github.com/dask/dask/issues/8465

rerun tests

tests/integration/test_select.py
charlesbluca, line 83: With the default option changed, we should be safe to remove these lines:

```suggestion
```

tests/integration/test_select.py
```diff
@@ -95,6 +96,8 @@ def test_select_of_select_with_casing(c, df):
     expected_df = pd.DataFrame(
         {"AAA": df["a"] + df["b"], "aaa": 2 * df["b"], "aAa": df["a"] - 1}
     )
+
+    c.drop_config("dask.sql.identifier.case.sensitive")
```
charlesbluca, line 100:
```suggestion
```

charlesbluca (commented):
Thanks for the work here @jdye64 😄 just one last comment:

jdye64:
rerun tests

charlesbluca:
rerun tests

charlesbluca (approved):
Thanks for the work here @jdye64! Think the gpuCI failures yesterday may have been related to the AWS outage

PR #273, opened 2021-10-20
Remove null-splitting from `_perform_aggregation`

charlesbluca:
dask-sql currently uses a workaround `get_groupby_with_nulls_cols` to split a dataframe by nulls/non-nulls before performing a groupby. This seems unnecessary, as we can use `groupby(dropna=False)` to keep nulls in the groupby result, and then later move them to the front of the dataframe (depending on how strictly we want to follow PostgreSQL). For comparison, here are the task graphs for a basic SQL operation:

```sql
SELECT
     user_id, SUM(b) AS "S"
FROM user_table_1
GROUP BY user_id
```

Before this PR:

```
HighLevelGraph with 20 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7fbc344ae590>
 0. from_pandas-b5125674539d8b5d1f754260e384e1fa
 1. assign-e29cf71863faee69539b6f78b8dbc8f3
 2. getitem-aa5bcd5e86836ebad1df2a90117e16f6
 3. fillna-94de4188d246caf9e1b4bf4f1debce03
 4. isnull-76ff5f3bc798aa3da45b6b16dae758fc
 5. inv-8c38b3dd8368fbb685b718f17180f00b
 6. aggregate-agg-838eda3f7eb4dbc264011a4f8e14a10b
 7. rename-c2e63c54b4d4fff51825029f94563613
 8. reset_index-3a34f3d8a1856ef12b299845da6afa8c
 9. rename-79dfdb294050405020a476146c8b5845
 10. getitem-be78c3b9d2c681e09cccdfdf783dbd1f
 11. eq-77e1f642c7c2a8e643f4a22b61af4e11
 12. inv-07fa50e82ef9e1e810819771a8e1ce2d
 13. getitem-1cc4911cff51b0537d276b6a9de6ebf7
 14. where-7cc4db419a8a2172e17accd935b851f4
 15. assign-8ac7e56123d06d1204acb6609e9541b6
 16. getitem-0ab1e73633488c2092c98ca87c31333c
 17. getitem-2a55368cca98a96e0d82198cd6bd7802
 18. assign-837d2902d726cbdd2f356860ab530f04
 19. getitem-f7c62a09776201bb03dae99feccd8138
```

And after:

```
HighLevelGraph with 16 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7f9c4455d730>
 0. from_pandas-b5125674539d8b5d1f754260e384e1fa
[...]

PR #326, opened 2021-11-22
Remove null-splitting from `_perform_aggregation`

charlesbluca:
Duplicate of #273, which was merged in before some groupby `split_out` tests could be resolved


tests/integration/test_groupby.py
```diff
@@ -397,10 +397,48 @@ def test_groupby_split_every(c, gpu, split_every, expected_keys):
...
-    dd.assert_eq(df.compute().sort_values("user_id"), expected_df, check_index=False)
+    dd.assert_eq(df, expected_df, check_index=False)
+
+    c.drop_config("dask.groupby.aggregate.split_every")
+    c.drop_table("split_every_input")
+
+
+@pytest.mark.gpu
+@pytest.mark.parametrize("split_every,expected_keys", [(2, 91), (3, 85), (4, 81)])
+def test_groupby_split_every_gpu(c, split_every, expected_keys):
```
charlesbluca, line 413: Now that we are able to access dask-cudf's groupby agg codepath, the task graph size varies depending on if we use GPU or not. There could be a clean way to parametrize this all into one test, but I opted for this for now to make the difference more clear.

In general, is there a way we can verify that `split_every` is working that doesn't depend on checking the task graph? Asking because now upstream changes in dask-cudf can break this test.

PR #328, opened 2021-11-23
Resolve failures in nightly package builds

charlesbluca:
Looks like something in `conda-build` is breaking down with `conda==4.11`? Not sure if this is related to using Boa, but checking to see if restricting `conda<4.11` resolves this issue


charlesbluca:
Looks like this is already being looked into:

https://github.com/conda/conda-build/issues/4333

PR #334, opened 2021-11-29
Circumvent deep copy of context in `PredictModelPlugin`

### LINKED ISSUE #332: [BUG]Predict fails with multi GPU cuML models
#### [BUG]Predict fails with multi GPU cuML models

We currently deep copy context when predicting with models (see [line](https://github.com/dask-contrib/dask-sql/blob/0c0578799bdf21879bce4e66dd9d4413b63791d5/dask_sql/physical/rel/custom/predict.py#L91)). This behavior is not supported for some multi-gpu cuML models due to lacking `__deepcopy__ ` attribute. 

**DASK-SQL MRE**

```python
cuml = pytest.importorskip("cuml", reason="cuml not installed")

model_query = """
        model_class = 'cuml.dask.linear_model.LinearRegression',
        target_column = 'target'
    ) AS (
        SELECT x, y, x*y AS target
        FROM timeseries
    )
    """
c.sql(model_query)

predict_sql = """
    SELECT * FROM PREDICT(
        MODEL my_model,
        SELECT x, y FROM timeseries
    )
    """
  c.sql(predict_sql)

```

**cuML MRE**

```python
from dask_cuda import LocalCUDACluster
from cuml.dask.linear_model import LinearRegression
import copy
import dask
import cudf

cluster = LocalCUDACluster()
client = Client(cluster)

df = dask.datasets.timeseries(freq="1D")
df = df.map_partitions(cudf.from_pandas).reset_index(drop=True)
y = df['x']*df['y']
model = LinearRegression()
model = model.fit(df[['x','y']], y)

copy.deepcopy(model)
``` 

```python
Function:  _get_model_attr
args:      (LinearRegressionMG(), '__deepcopy__')
kwargs:    {}
Exception: 'AttributeError("Attribute __deepcopy__ does not exist on model <class \'cuml.linear_model.linear_regression_mg.LinearRegressionMG\'>")'
```

## Proposed Solution
IMO creating a deep copy of context seems really expensive honestly so long term this does not seem like a good solution so the first pass at this will be to try to remove that copy . 

If that fails we can look into enabling deep copy for these models.

See related [PR](https://github.com/dask-contrib/dask-sql/pull/330)

charlesbluca:
Closes #332

It looks like the relatively expensive deep copy of the context done in `PredictModelPlugin` should be circumvented by simply adding the temporary table we create to the original context and dropping it before returning; this also unblocks usage with multi-GPU cuML models.

cc @VibhuJawa

### DIFF
--- a/dask_sql/physical/rel/custom/predict.py
+++ b/dask_sql/physical/rel/custom/predict.py
@@ -1,4 +1,3 @@
-import copy
 import logging
 import uuid
 from typing import TYPE_CHECKING
@@ -88,8 +87,7 @@ class PredictModelPlugin(BaseRelPlugin):
             else:  # pragma: no cover
                 continue
 
-        tmp_context = copy.deepcopy(context)
-        tmp_context.create_table(temporary_table, predicted_df)
+        context.create_table(temporary_table, predicted_df)
 
         sql_ns = org.apache.calcite.sql
         pos = sql.getParserPosition()
@@ -112,8 +110,9 @@ class PredictModelPlugin(BaseRelPlugin):
             None,  # hints
         )
 
-        sql_outer_query = tmp_context._to_sql_string(outer_select)
-        df = tmp_context.sql(sql_outer_query)
+        sql_outer_query = context._to_sql_string(outer_select)
+        df = context.sql(sql_outer_query)
+        context.drop_table(temporary_table)
 
         cc = ColumnContainer(df.columns)
         dc = DataContainer(df, cc)
--- a/tests/integration/test_model.py
+++ b/tests/integration/test_model.py
@@ -26,8 +26,11 @@ def check_trained_model(c, model_name=None):
         )
         """
 
+    tables_before = c.schema["root"].tables.keys()
     result_df = c.sql(sql).compute()
 
+    # assert that there are no additional tables in context from prediction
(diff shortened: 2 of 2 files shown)


dask_sql/physical/rel/custom/predict.py
```diff
@@ -112,8 +110,9 @@ def convert(
             None,  # hints
         )
 
-        sql_outer_query = tmp_context._to_sql_string(outer_select)
-        df = tmp_context.sql(sql_outer_query)
```
VibhuJawa, line 116: Seems reasonable to me but can we should a test to ensure we have not broken anything. Like are we sure that  `sql_outer_query ` does not add anything . 

Maybe we add a test to verify that the registered tables still line up and nothing temporary is left around.

VibhuJawa (changes requested):
Thanks for the work on this @charlesbluca . Looks reasonable to me .  Have requested some testing to ensure we dont miss anything.

dask_sql/physical/rel/custom/predict.py
charlesbluca: Good idea, I will add that in now

VibhuJawa (approved):
Thanks for the quick work on this @charlesbluca . LGTM.

PR #340, opened 2021-12-03
[WIP] Attempt to resolve gpuCI breakage

charlesbluca:
Not exactly sure what's causing the gpuCI failures yet, but I do notice the following popping up in our `pip install` logs:

```
ERROR: After October 2020 you may experience errors when installing or updating packages. This is because pip will change the way that it resolves dependency conflicts.

We recommend you use --use-feature=2020-resolver to test your packages with the new resolver before it becomes the default.

dask-cudf 22.2.0a0+103.g83adc4b6cf requires cupy-cuda115, which is not installed.
```

I am going to try out the new `pip` resolver to see what happens there.

### DIFF
--- a/continuous_integration/gpuci/build.sh
+++ b/continuous_integration/gpuci/build.sh
@@ -41,13 +41,13 @@ gpuci_logger "Activate conda env"
 conda activate dask_sql
 
 gpuci_logger "Install dask"
-python -m pip install git+https://github.com/dask/dask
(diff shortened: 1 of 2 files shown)


charlesbluca:
Resolved, looks like we had mismatched versions of cudf / libcudf due to some implicit changes in cuDF's numpy dependency from https://github.com/rapidsai/cudf/pull/9687

PR #343, opened 2021-12-09
Fix TIMESTAMP / DATE scalars, add support for DATE column casting

### LINKED ISSUE #296: [BUG] Selecting based on dates unexpectedly returns no results.

charlesbluca:
This is a continuation of #298 that also aims to implement `CAST(... AS DATE)` properly, as up until now that would work identically to `CAST(... AS TIMESTAMP)`:

- Casts DATE / TIMESTAMP scalars to `np.datetime64` instead of `datetime.datetime` for compatibility with Pandas; for whatever reason `datetime.datetimes` work in cuDF, which should probably be documented somewhere
- Casts DATE scalars to `np.dtype("datetime64[D]")` and TIMESTAMP scalars to `np.dtype("datetime64[ns]")`, so that DATE filters work as expected
- Adds special support for `CAST (... AS DATE)` using the datetime column's `dt.date` attribute; this currently doesn't work in dask-cuDF as it is dependent on the datetime accessor of cuDF, which doesn't have a `date` attribute
  - In general, this is a hacky workaround for truncating timestamps that is required because both Pandas and cuDF don't allow truncation through `astype("datetime64[D]")`; in Pandas this only applies for timezoned datetimes, and in cuDF this applies for all datetimes

Closes #296 
Closes #298


beckernick:
@charlesbluca is https://github.com/rapidsai/cudf/issues/7880 in the critical path for this PR, or are there relatively straightforward workarounds?

charlesbluca:
Need to look through the code again to make sure this is accurate, but IIRC another path forward here would be to add support for casting pandas timezoned datetime / cudf datetime columns to `datetime64[D]` - if that were the case, we would be able to truncate the timestamps to dates with a single `astype` call and avoid the use of the `DatetimeAccessor`.

charlesbluca:
@ChrisJar I've checked and this _should_ resolve #296 - if you get a chance, could you verify?

dask_sql/physical/rex/core/call.py
```diff
@@ -224,14 +224,20 @@ def cast(self, operand, rex=None) -> SeriesOrScalar:
...
-            return operand
-        else:
-            return return_column
+            return_column = operand
+
+        # TODO: ideally we don't want to directly access the datetime dates,
+        # but Pandas can't truncate timezone datetimes and cuDF can't
+        # truncate datetimes
+        if output_type == "DATE":
+            return return_column.dt.strftime("%Y-%m-%d").astype(python_type)
```
charlesbluca, line 238: We can't cast a pandas/cudf column to a `datetime64[D]` column directly with `astype`, and we can't use `col.dt.date` to extract the dates as Python datetimes because this isn't supported by cuDF - seems like the only workaround other than that is to cast the column into a truncated "date" string column, and then cast that back to `datetime64[ns]`.

cc @shwina is there possibly a better way we could be truncating the time (hours, minutes, seconds) from a `datetime64[ns]` column?

charlesbluca: Opting to use the solution described [here](https://github.com/rapidsai/cudf/issues/7880#issuecomment-1030124233):

```python
In [42]: s
Out[42]:
0   2001-01-01 12:12:12
1   2001-01-02 01:02:03
dtype: datetime64[ns]

In [43]: s.dt.floor("D")
Out[43]:
0   2001-01-01
1   2001-01-02
dtype: datetime64[ns]
```

PR #345, opened 2021-12-13
Bump dask dependency to `>=2021.11.1,<=2021.11.2`

charlesbluca:
In preparation for our next release which will be packaged alongside RAPIDS 21.12, this PR bumps our Dask dependency to match up with that of dask-cudf. Note that we should be safe to remove the upper bound on this dependency once the release is cut and published.

cc @randerzander @jakirkham


jakirkham:
cc @quasiben

jakirkham:
Thanks Charles! 😄

PR #347, opened 2021-12-14
Update Calcite to 1.29.0 and log4j to 2.17.0 to address CVE-2021-44228

### LINKED ISSUE #346: log4j Vulnerability
Due to the recently discovered log4j vulnerability I think we need to restrict log4j versions to 2.15.0.  I _think_ this done in the [pom.xml](https://github.com/dask-contrib/dask-sql/blob/main/planner/pom.xml) file

xref https://www.cisa.gov/uscert/ncas/current-activity/2021/12/13/cisa-creates-webpage-apache-log4j-vulnerability-cve-2021-44228

@jdye64 is this something you're familiar with ?

jdye64:
This closes #346


jdye64:
Please note the Calcite jars for 1.29.0 have not yet synced with the maven repository I am just opening this PR to get out in front of it. Once they sync we can run the build again and I can make any needed changes. The build will fail as is now since it will not be able to find them.

planner/pom.xml
```diff
@@ -14,10 +14,11 @@
 		<project.reporting.outputEncoding>${sourceEncoding}</project.reporting.outputEncoding>
 
 		<java.version>1.8</java.version>
+		<log4j.version>2.16.0</log4j.version>
```
charlesbluca, line 17: Just making sure this isn't a typo and that we intend to bump to `2.16.0`?

jdye64: It is not a typo. The guidance had changed after making the PR and its title to use 2.16.0 instead of 2.15.0 I will just update the PR title as well to prevent confusion.

jdye64:
Rerun tests

jdye64:
Good news is Calcite 1.29.0 is now released. The bad news is the update broke some of our bindings. I will work on fixing those.

jdye64:
Alright, all breaking changes have been fixed. However seeing some Java stackoverflow errors on the cost based optimization routines since the upgrade.

jdye64:
Rerun tests

tests/integration/test_rex.py
```diff
@@ -7,6 +7,7 @@
 from pandas.testing import assert_frame_equal
 
 
+@pytest.mark.xfail
```
charlesbluca, line 10: We should probably give a reason here for clarity moving forward:

```suggestion
@pytest.mark.xfail(
    reason="Bumping to Calcite 1.29.0 to address CVE-2021-44228 caused a stack overflow in this test"
)
```

charlesbluca (approved):
Thanks for sorting this out @jdye64 😄 just one last comment then this should be good to merge in:

PR #351, opened 2022-01-03
Feature/jdbc

PeterLappo:
Hi,
I've added some system information to the prestodb JDBC driver so that you can see schema, table and column data in tools like dbeaver which are Java tools that use the presto JDBC driver. I'm presuming but haven't tested it that it will also work with presto ODBC drivers.
Here are a couple of dbeaver screen shots.
dbeaver1
dbeaver2

Thanks


dask_sql/server/app.py
```diff
@@ -74,6 +75,8 @@ async def query(request: Request):
     """
     try:
         sql = (await request.body()).decode().strip()
+        # required for JDBC Driver
+        sql = adjust_for_presto_sql(sql)
```
charlesbluca, line 79: If this is mandatory for compatibility with JDBC driver and is always occurring, perhaps we should just perform the adjustment directly here?

tests/integration/test_jdbc.py
```diff
@@ -0,0 +1,155 @@
...
+            "type": "varchar",
+            "typeSignature": {"rawType": "varchar", "arguments": []},
+        }
+    ]
+    assert len(result["data"]) == 3
+    assert result["data"] == [
+        ["", "root"],
+        ["", "a_schema"],
+        ["", "system_jdbc"],
+    ]
```
charlesbluca, line 72: Wondering if there is a way to place these assertions in a helper function like `assert_resp` to simplify the tests; okay if this isn't possible

tests/integration/test_server.py
```diff
@@ -3,14 +3,17 @@
...
 # needed for the testclient
 pytest.importorskip("requests")
 
 
 @pytest.fixture(scope="module")
 def app_client():
-    _init_app(app)
+    c = Context()
+    c.sql("SELECT 1 + 1").compute()
+    _init_app(app, c)
```
charlesbluca, line 16: Any reason why we're manually creating a context for the standard server tests?

charlesbluca (commented):
Thanks for the work here @PeterLappo 😄 a couple comments:

Also looks like this PR is failing style checks, this should be resolved by running `pre-commit run --all-files` on your branch and pushing the resulting changes

dask_sql/server/presto_jdbc.py
```diff
@@ -0,0 +1,175 @@
...
+
+    logger.info("jdbc meta data ready")
+
+
+def create_catalog_row(catalog: str = ""):
+    return {"TABLE_CAT": catalog}
+
+
+def create_schema_row(catalog: str = "", schema: str = ""):
+    return {"TABLE_CATALOG": catalog, "TABLE_SCHEM": schema}
```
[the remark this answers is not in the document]
PeterLappo: This is what the presto driver expects. I thought it odd too

dask_sql/server/presto_jdbc.py
```diff
@@ -0,0 +1,175 @@
...
+
+    :param sql: str
+    :return: str
+    """
+    if sql.find("system.jdbc") > 0:
+        # Catalog not supported and the '.' will cause queries to fail so replace with '_'
+        sql = sql.replace("system.jdbc", "system_jdbc")
+        logger.info(f"presto sql={sql}")
+
+    return sql
```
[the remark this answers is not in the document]
PeterLappo: Fair, will update...

dask_sql/server/app.py
```diff
@@ -179,6 +182,7 @@ def run_server(
 
     """
     _init_app(app, context=context, client=client)
+    create_meta_data(context)
```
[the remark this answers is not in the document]
PeterLappo: I did consider a flag, but makes sense, and will disable by default

tests/integration/test_jdbc.py
```diff
@@ -0,0 +1,155 @@
...
+            break
+
+        next_url = result["nextUri"]
+
+        counter += 1
+        assert counter <= 100
+
+        sleep(0.1)
+
+    return result
```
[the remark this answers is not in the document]
PeterLappo: I come from a Java world where the convention is to put "private" code at the end but actually coped from test_server.py which has it at the end. Shall I re-use from test_server.py? @charlesbluca

tests/integration/test_server.py
PeterLappo: The test fails if you run `pytest tests/integration/test_server.py` in isolation. And as I used this module as starter for test_jdbc.py I discovered after some messing about that tests/integration/test_compatibility.py  has to run first to ensure test_server.py and test_jdbc.py could run. It means both will now run independently.

PeterLappo:
@charlesbluca Updated as recommended. What is the next step to merge to main?

GPUtester:

charlesbluca:
add to allowlist

dask_sql/server/presto_jdbc.py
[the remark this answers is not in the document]
charlesbluca: Ah okay thanks for the clarification - if Trino is the more active fork of Presto, perhaps it would be worthwhile to explore if it is possible to use here instead of PrestoDB, though there is no push to do that before merging this in

tests/integration/test_jdbc.py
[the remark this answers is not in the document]
charlesbluca: Ah I didn't realize that this was also the case in test_server.py; in that case my preference would be to move the util functions up to the top in both files - my reasoning here being to match up with other testing files (such as [test_compatibility.py](https://github.com/dask-contrib/dask-sql/blob/main/tests/integration/test_compatibility.py)) that place their utility functions at the top of the file.

I also note that since we are duplicating functions like `get_result_or_error`, we might want to add a testing utility module as a central location for functions like this, but that is out of the scope of this PR.

tests/integration/test_server.py
charlesbluca: That makes sense, thanks for the explanation

dask_sql/server/presto_jdbc.py
```diff
@@ -0,0 +1,150 @@
...
+    if c is None:
+        logger.warn("Context None: jdbc meta data not created")
+        return
+    catalog = ""
+    system_schema = "system_jdbc"
+    c.create_schema(system_schema)
+
+    # if we supply a catalogue dbeaver (for instance) tries to use it
+    # but dask_sql presto interface doesn't support catalog
+    # so better for now not to supply a catalog
```
charlesbluca, line 36: If we leave the commented code, maybe we should format this as a `TODO` to indicate that future work can be done to enable catalogs:

```suggestion
    # TODO: add support for catalogs in presto interface
    # see https://github.com/dask-contrib/dask-sql/pull/351
```

dask_sql/server/app.py
```diff
@@ -74,6 +75,16 @@ async def query(request: Request):
...
         sql = (await request.body()).decode().strip()
+        # required for Prestodb JDBC driver compatibility
+        # The driver sends a Catalog called system which is not supported by dask_sql
+        # and the '.' will cause queries to fail so we replace with '_'
+        # In parallel to this create_meta_data(context) creates a schema called
+        # system_jdbc which contains tables compatible with the Prestodb Driver
+        # Note1: Trinodb driver (a fork of Prestodb) unfortunately doesn't work with dask_sql
+        # as it has different http headers
+        # Note2: For a JDBC request to succeed create_meta_data(context) needs to be enabled
+        # and called after all user tables have been created in order to create the JDBC metadata
```
charlesbluca, line 86: Just trying to cut down this explanation; feel free to let me know if I've missed something here

```suggestion
        # required for PrestoDB JDBC driver compatibility
        # replaces queries to unsupported `system` catalog with queries to `system_jdbc`
        # schema created by `create_meta_data(context)` when `jdbc_metadata=True`
        # TODO: explore Trino which should make JDBC compatibility easier but requires
        # changing response headers (see https://github.com/dask-contrib/dask-sql/pull/351)
```

charlesbluca (commented):
Thanks for addressing my concerns @PeterLappo 🙂 would you mind merging the latest changes? Believe the last test failures were due to some changes in upstream libraries which should be resolved now

PeterLappo:
@charlesbluca have merged with main. Does CI pipeline run automatically?

charlesbluca:
Thanks!

> Does CI pipeline run automatically?

It normally does, but since you haven't pushed any commits to this repo your PR requires approval for CI to run - I've approved and this should be good to merge if tests pass

charlesbluca (approved):

PR #355, opened 2022-01-06
Fix issue reporting in scheduled upstream testing

charlesbluca:
It looks like something was going wrong in the conditionals of the upstream testing workflow that caused test failures to not trigger an issue; I've copied over some of the code from dask-ml's working [upstream testing workflow](https://github.com/dask/dask-ml/blob/main/.github/workflows/upstream.yml) so that this hopefully isn't an issue moving forward.


charlesbluca:
rerun tests

PR #359, opened 2022-01-10
Remove Join Condition Push CBO Rule since it was causing infinite cos…

jdye64:
…t based optimization for certain queries


charlesbluca:
cc @randerzander

charlesbluca (approved):
Approving now but will hold off on merging until we get other confirmation that this resolves the infinite CBO

randerzander:
This is resolved! Thanks all!

PR #365, opened 2022-01-10
Switch tests from `pd.testing.assert_frame_equal` to `dd.assert_eq`

charlesbluca:
This should help to generally clean up some of the tests by allowing us to remove a lot of the unnecessary `compute` and `to_pandas` calls


charlesbluca:
Something important to note is that by default, `assert_eq` uses a local, synchronous scheduler to compute input Dask dataframes. In addition to causing the observed independent test clusters above (which arise from trying to compute dataframes with the local scheduler dependent on data stored in the independent cluster), this mean that all `assert_eq` tests would need to have a conditional kwarg passed through to ensure that the independent cluster is used when it is active.

Because this solution is rather cumbersome, I'm going to explore upstream work to make it so Dask allows `assert_eq`'s scheduler to be controlled through its config module.

cc @ayushdg

charlesbluca:
Hoping that https://github.com/dask/dask/pull/8811 should unblock this

tests/integration/test_rex.py
```diff
@@ -77,49 +76,29 @@ def test_literal_null(c):
...
-    result_df = result_df.compute()
+def test_random(c):
+    query = 'SELECT RAND(0) AS "0", RAND_INTEGER(0, 10) AS "1"'
 
-    # As the seed is fixed, this should always give the same results
-    expected_df = pd.DataFrame({"0": [0.26183678695392976], "1": [8]})
-    expected_df["1"] = expected_df["1"].astype("Int32")
-    assert_frame_equal(result_df, expected_df)
+    result_df = c.sql(query)
+    expected_df = c.sql(query)
```
ayushdg, line 89: I'm guessing the idea here is to verify that multiple executions with the same seed give the same result. Could we rename `expected_df` to something else since I was a little confused why `expected_df` was re-running the same query initially.

ayushdg (approved):
Thanks a lot for these changes! This should significantly simplify tests moving forward and make adding gpu tests even easier.

PR #366, opened 2022-01-11
[REVIEW] Fix unary conditional join operations

jdye64:


tests/integration/test_join.py
```diff
@@ -183,3 +183,23 @@ def test_join_literal(c):
...
+        SELECT
+                a
+            FROM df1
+            INNER JOIN df2 ON
+            (
+                a = c
+                AND b IS NOT NULL
+            )
+    """
+    ddf = c.sql(query)
```
charlesbluca, line 205: Can we compare this dataframe to the expected pandas result?

charlesbluca, line 205: Minor nitpick:

```suggestion
    ddf = c.sql(
        """
    SELECT a FROM df1
    INNER JOIN df2 ON
    (
        a = c
        AND b IS NOT NULL
    )
    """
    )
```

tests/integration/test_join.py
```diff
@@ -183,3 +183,23 @@ def test_join_literal(c):
     df_expected = pd.DataFrame({"user_id": [], "b": [], "user_id0": [], "c": []})
 
     assert_frame_equal(df.reset_index(), df_expected.reset_index(), check_dtype=False)
+
+
+def test_conditional_join(c):
+    df1 = pd.DataFrame({"a": [1, 2, 2], "b": ["x", "y", "z"]})
```
charlesbluca, line 189: Should we add some nulls here to ensure the conditional is working?

charlesbluca (commented):
Thanks for digging into this @jdye64 🙂 couple comments around the test:

tests/integration/test_select.py
```diff
@@ -116,3 +116,11 @@ def test_timezones(c, datetime_table):
...
 
     assert_frame_equal(result_df, datetime_table)
+
+
+def test_multi_case_when(c):
+    df = pd.DataFrame({"a": [1, 6, 7, 8, 9]})
+    c.create_table("df", df)
+
+    query = "SELECT CASE WHEN a BETWEEN 6 AND 8 THEN 1 ELSE 0 END FROM df"
+    c.sql(query).compute()
```
charlesbluca, line 126: Could we also check if this gives the expected result?

```suggestion
    actual_df = c.sql(
        """
    SELECT
        CASE WHEN a BETWEEN 6 AND 8 THEN 1 ELSE 0 END AS C
    FROM df
    """
    ).compute()
    expected_df = pd.DataFrame({"C": [0, 1, 1, 1, 0]})

    assert_frame_equal(actual_df, expected_df)
```

charlesbluca (approved):
Thanks for the work here @jdye64! LGTM

randerzander:
I believe this is intended to fix a StackOverflow condition when running [q14](https://github.com/rapidsai/gpu-bdb/blob/999818b7450fb86de75415c3ff6232a2d730ed7e/gpu_bdb/queries/q14/gpu_bdb_query_14_dask_sql.py).

Unfortunately even when I include this fix, I still get a hang and the DAG is never fully generated. I suspect this may be another CBO infinite loop?

randerzander:
With @jdye64 's latest changes here, the query works again as expected.

This is good to merge from my perspective.

charlesbluca:
Failures seem to be related to CodeCov and not actual tests failing - going to merge this in. Thanks again @jdye64!

PR #368, opened 2022-01-13
Add support for `gpu` kwarg in `Context.sql` and `explain`

charlesbluca:
While working on #365, I noticed that `Context.sql` and `explain` both allow dataframes to be input directly, prompting the corresponding `create_table` calls before executing.

This PR adds the `gpu` kwarg to these methods, to be passed to the underlying `create_table` calls.


ayushdg (approved):
Changes look good to me and should be good to merge once CI passes.

Issue #350, opened 2021-12-21
⚠️ Nightly upstream-dev CI failed ⚠️


charlesbluca:
Looks like the Dask upstream testing has started failing:

https://github.com/dask-contrib/dask-sql/actions/runs/1604667145

This should've triggered an issue to be opened automatically, but didn't; I can look into that later, but it seems to have something to do with these conditionals:

https://github.com/dask-contrib/dask-sql/blob/2e5cfcb59701ea7d35b146a32690a0315b7f82d8/.github/workflows/test-upstream.yml#L102-L106

https://github.com/dask-contrib/dask-sql/blob/2e5cfcb59701ea7d35b146a32690a0315b7f82d8/.github/workflows/test-upstream.yml#L116-L119

cc @quasiben @galipremsagar

ayushdg:
Test failures seem to be ml related. I can look into this unless someone already is

galipremsagar:
@ayushdg You can look into it. AFAIK no one else is looking into it.

ayushdg:
Tracked this down to failures in dask/dask-ml#895 post the merge of dask/dask#8452.

ayushdg:
This should now be fixed since dask/dask-ml#898 merged. @charlesbluca Would it be possible to trigger a run with upstream dask and dask-ml to see if the issue is resolved on our end?

charlesbluca:
Currently our upstream testing only uses latest Dask, but I can modify it to also use latest dask-ml and then run a test manually.

charlesbluca:
Tested with the latest dask / dask-ml in this [run](https://github.com/dask-contrib/dask-sql/actions/runs/1660208403) and looks like things are working again 🎉 thanks for the work here @ayushdg!

PR #369, opened 2022-01-13
Remove max version restriction for Dask/Distributed

charlesbluca:
In advance of our next release, removes the max version restriction for Dask/Distributed so that we can check for potential compatibility issues (specifically those raised in https://github.com/dask-contrib/dask-sql/issues/350 since dask-ml hasn't cut a release since @ayushdg's fix)


ayushdg (approved):
Thanks for these changes. I suspect things will work now since dask hasn't cut a release since the changes that broke dask-ml merged in. But I believe the dask community is planning a release sometime this/next week and I suspect we'll see some failures seen in #350 until a dask-ml release.

PR #383, opened 2022-01-26
Drop support for python 3.7, add testing for python 3.10

charlesbluca:
This PR follows Dask's dropping of python 3.7 support and migration to supporting 3.10 by doing the same here.


continuous_integration/environment-3.10-jdk11-dev.yaml
```diff
@@ -0,0 +1,43 @@
...
+- fs>=2.4.11
+- intake>=0.6.0
+- isort=5.7.0
+- jpype1>=1.0.2
+- lightgbm>=3.2.1
+- maven>=3.6.0
+- mlflow>=1.19.0
+- mock>=4.0.3
+- nest-asyncio>=1.4.3
+- numpy<1.22  # numba only supports up to numpy 1.21
```
charlesbluca, line 22: We shouldn't have to make this pinning - the `defaults`-hosted Numba package that our Conda environments pick up doesn't have the expected max version constraint on numpy, which makes it possible to install a conflicting 1.22+ version.

Opened up https://github.com/numba/numba/issues/7794 describing this issue.

charlesbluca: Resolved this by removing the `defaults` channel

Issue #385, opened 2022-01-28
[BUG] `select from table limit` reads the full dataset and persists in memory. 


ayushdg:
**What happened**:
When performing a `SELECT * FROM table LIMIT 10`, from a table read in via parquet, I notice the full dataset being read and persisted on query execution.

**What you expected to happen**:
Nothing to happen at query execution and when the user does decide to persist/compute the result only the relevant subset of data is read in. 

**Minimal Complete Verifiable Example**:

```python
from dask_cuda import LocalCUDACluster
from distributed import Client, wait
import cudf
import dask_cudf
from dask_sql import Context
import dask

write_data = False

if __name__ == "__main__":
    cluster = LocalCUDACluster()
    client = Client(cluster)
    c = Context()
    
    if write_data:
        dask.datasets.timseries(start="2022-01-01", end="2024-01-01").to_parquet("test_data.parquet")


    ddf = dask_cudf.read_parquet("test_data.parquet")
    c.create_table("test", ddf, persist=False)

    # This results in the whole dataset persisted in memory and even though `len(res)==5` all the data is in memory
    res = c.sql("SELECT * from test LIMIT 10")
```

**Anything else we need to know?**:

**Environment**:

- dask-sql version: 2022.1.0
- Python version: 3.8
- Operating System: ubuntu 18.04
- Install method (conda, pip, source): conda

charlesbluca:
Looks like this is happening because of a direct call to `persist()` in `_apply_offset`:

https://github.com/dask-contrib/dask-sql/blob/b571a600a639f0dcabd9d0162d01d078472e3e51/dask_sql/physical/rel/logical/limit.py#L61

Which is used by the `DaskLimitPlugin` to handle offsets and limits on select statements.

ayushdg:
Thanks for triaging. So there's a couple of things I'd like to discuss.
- Explore if not persisting is an option. 
- Irrespective of the persist, do we need to read in the full data or can we read in a subset similar to doing ddf.head()

charlesbluca:
> do we need to read in the full data or can we read in a subset similar to doing ddf.head()

It looks like this is what's being done immediate after the persist; we check if the requested data is located in the first partition, and if so, simply use `df.head()`:

https://github.com/dask-contrib/dask-sql/blob/b571a600a639f0dcabd9d0162d01d078472e3e51/dask_sql/physical/rel/logical/limit.py#L62-L66

[...]

PR #387, opened 2022-01-28
Don't persist dataframes before applying offset / limit

charlesbluca:
Should address the persisting of entire dataframes in `SELECT ... LIMIT` statements observed in #385


ayushdg (approved):
Thanks for the quick fix. Changes LGTM! 

I'll wait for @randerzander to confirm that this resolves what he was seeing.

randerzander:
> I'll wait for @randerzander to confirm that this resolves what he was seeing.

It does =)

Thanks all!

PR #389, opened 2022-01-31
Enable ability for user to pass in a list of CBO rules that should be…

jdye64:
… disabled for a context


planner/src/main/java/com/dask/sql/application/RelationalAlgebraGeneratorBuilder.java
```diff
@@ -14,11 +15,13 @@
 	private final String rootSchemaName;
 	private final List<DaskSchema> schemas;
 	private final boolean case_sensitive; // False if case should be ignored when comparing SQLNode(s)
+	private final ArrayList<RelOptRule> disabled_rules; // CBO rules that should be disabled for this context
 
-	public RelationalAlgebraGeneratorBuilder(final String rootSchemaName, final boolean case_sensitive) {
+	public RelationalAlgebraGeneratorBuilder(final String rootSchemaName, final boolean case_sensitive, final ArrayList<RelOptRule> disabled_rules) {
```
ayushdg, line 20: This method is expecting an `ArrayList` of `RelOptRule` objects but what it would get from python is actually an arrayList of strings. Do you think it makes sense to add a utility that accepts the strings and creates RelOptRule objects from them?

planner/src/main/java/com/dask/sql/application/DaskPlanner.java
```diff
@@ -86,4 +103,12 @@ public DaskPlanner() {
...
         return defaultContext;
     }
+
+    public ArrayList<RelOptRule> getAllRules() {
+        return this.ALL_RULES;
+    }
+
+    public ArrayList<RelOptRule> getDisabledRules() {
+        return this.disabledRules;
+    }
```
ayushdg, line 113: Does it make sense to also maintain a list of enabledRules in this class, or is it up to the python layer to compute that with these two bits of information.

planner/src/main/java/com/dask/sql/application/RelationalAlgebraGeneratorBuilder.java
jdye64: Completely. I was going to expose the `RelOptRule` inside of `java.py` but I agree it makes more since to just let Java handle that conversion and hide that mess from Python. I will add another commit doing that.

planner/src/main/java/com/dask/sql/application/DaskPlanner.java
jdye64: Yep, that would be really easy to do so I will add it to the Java side. Also I need to actually expose these functions to the python side since I had forgotten to do that in the past commit.

randerzander (changes requested):
Given the intent here is to allow users to turn off specific CBO rules which may be problematic, we'll need a place where they can see how to configure them, and what constants/enums they need to supply.

I think this is as good a time as any to start the beginnings of a docs page about Dask-SQL's use of Calcite's CBO.

planner/src/main/java/com/dask/sql/application/DaskPlanner.java
jdye64: @ayushdg would it be helpful if the return value here was just a list of strings instead of a list of `RelOptRule`?

planner/src/main/java/com/dask/sql/application/RelationalAlgebraGenerator.java
```diff
@@ -1,10 +1,13 @@
 package com.dask.sql.application;
 
 import java.sql.SQLException;
+import java.util.ArrayList;
 import java.util.List;
 
 import com.dask.sql.schema.DaskSchema;
 
+import org.apache.calcite.plan.RelOptRule;
+import org.apache.calcite.plan.RelOptRuleCall;
```
charlesbluca, line 10: Maybe I'm missing something, but could we remove this import? Looks like it's not being used anywhere in this file

jdye64: `RelOptRuleCall` can be removed but `RelOptRule` is used in the constructor. I'll make that change now

charlesbluca (approved):
LGTM, thanks for the work @jdye64! Could we maybe open an issue to track the follow up work that needs to happen here?

PR #394, opened 2022-02-08
Update docs theme, use sphinx-tabs for CPU/GPU examples

charlesbluca:
This PR bumps the `dask-sphinx-theme` to be more in line with Dask / Distributed's docs, and adds the `sphinx-tabs` extension so that code-blocks can be tabbed to show their GPU equivalent (when possible)


charlesbluca:
@VibhuJawa if you get a chance could you check out `machine_learning.rst` and let me know if there's any other blocks in there that can be run on GPU? Based my examples off the GPU model tests but not sure if I missed anything 🙂

VibhuJawa:
> @VibhuJawa if you get a chance could you check out `machine_learning.rst` and let me know if there's any other blocks in there that can be run on GPU? Based my examples off the GPU model tests but not sure if I missed anything 🙂

Looking at it right now. We need to add to do two things: 

1.  Add explanation around `wrap_fit` and `wrap_predict` . 
2.  Use the same `cuML` and `sklearn` model to avoid confusion 

Will make a PR with the suggested changes to this PR in a bit.

charlesbluca:
That sounds good to me - I can make TODOs for those examples so that we can remember to add GPU sections there in a follow up.

It's not immediately obvious what code changes would be made to example 6 to account for GPUs - or is the idea that a model created with a cuML class should have no issues being exported?

VibhuJawa:
> It's not immediately obvious what code changes would be made to example 6 to account for GPUs - or is the idea that a model created with a cuML class should have no issues being exported?

Depends on the model, i think most models work but some models like FIL have issues . We would have to test it to know.  Anyways, having this under CI will give us confidence of putting stuff under CI. 

> That sounds good to me - I can make TODOs for those examples so that we can remember to add GPU sections there in a follow up.


Yup. Raised issue to track this,  https://github.com/dask-contrib/dask-sql/issues/397 .  Lets add stuff there. :-)

charlesbluca:
Going to go ahead and merge this in, as the gpuCI failures are unrelated

PR #402, opened 2022-02-14
Generalize dtype check in `python_to_sql_type`

charlesbluca:
Chatting with @DaceT exposed that `python_to_sql_type` only has handling for pandas / numpy types, which can cause issues when attempting to read in a table with a cuDF dtype, like so:

```python
import cudf

from dask_sql import Context

c = Context()
cdf = cudf.datasets.timeseries()

c.create_table("tble", cdf)
result = c.sql("""SELECT timestamp FROM cdf""") 
result.compute()
```

```
TypeError: unhashable type: 'CategoricalDtype'
```

A pretty simple workaround for this is generalizing the column dtype check so that instead of checking for a `np.dtype`, we check if the Python type is "dtype-like".

Interested in if there's a trivial util function like `is_dataframe_like` for dtypes?

### DIFF
--- a/dask_sql/mappings.py
+++ b/dask_sql/mappings.py
@@ -88,11 +88,11 @@ _SQL_TO_PYTHON_FRAMES = {
 def python_to_sql_type(python_type):
     """Mapping between python and SQL types."""
 
-    if hasattr(python_type, "type"):
-        python_type = python_type.type
-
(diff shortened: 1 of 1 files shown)


charlesbluca:
Things are a little more complicated than I initially assumed here - opened #423 to track this issue

Issue #401, opened 2022-02-14
[DOC] API docs not rendering


ayushdg:
## Report incorrect documentation

**Location of incorrect documentation**
The api documentations [page](https://dask-sql.readthedocs.io/en/latest/pages/api.html) returns a blank page on the docs site.

Might be similar to #215 

---

charlesbluca:
Attempting to build the docs, it looks like we fail when trying to load in a non-existent function:

```
Running Sphinx v4.4.0
loading pickled environment... done
building [mo]: targets for 0 po files that are out of date
building [html]: targets for 15 source files that are out of date
updating environment: 0 added, 15 changed, 0 removed
reading sources... [100%] sql/select                                                                                                                                                                                                                                                                                  
WARNING: autodoc: failed to import function 'fsql' from module 'dask_sql.integrations.fugue'; the following exception was raised:
Traceback (most recent call last):
  File "/datasets/charlesb/miniconda3/envs/dask-sql-docs/lib/python3.9/site-packages/sphinx/util/inspect.py", line 448, in safe_getattr
    return getattr(obj, name, *defargs)
AttributeError: module 'dask_sql.integrations.fugue' has no attribute 'fsql'

The above exception was the direct cause of the following exception:

Traceback (most recent call last):
  File "/datasets/charlesb/miniconda3/envs/dask-sql-docs/lib/python3.9/site-packages/sphinx/ext/autodoc/importer.py", line 110, in import_object
    obj = attrgetter(obj, mangled_name)
  File "/datasets/charlesb/miniconda3/envs/dask-sql-docs/lib/python3.9/site-packages/sphinx/ext/autodoc/__init__.py", line 332, in get_attr
    return autodoc_attrgetter(self.env.app, obj, name, *defargs)
  File "/datasets/charlesb/miniconda3/envs/dask-sql-docs/lib/python3.9/site-packages/sphinx/ext/autodoc/__init__.py", line 2828, in autodoc_attrgetter
    return safe_getattr(obj, name, *defargs)
  File "/datasets/charlesb/miniconda3/envs/dask-sql-docs/lib/python3.9/site-packages/sphinx/util/inspect.py", line 464, in safe_getattr
    raise AttributeError(name) from exc
AttributeError: fsql
```

[...]

PR #406, opened 2022-02-17
Use `build_ext`/`install_lib` subclasses to build external java

### LINKED ISSUE #400: [ENH] Changing `java` command to build option

Currently Dask-SQL has a non-standard build option, `java`, which is used to build the Java components. Though this is presumably something we would want to do during the package build. This would also make it easier to use with other Python build tools familiar with standard Python build options (like `build_ext`).

Would be nice to have the Java build step implemented as a `build_ext` subclass with an option to build the Java component. This would allow for a standard command to be used, `build_ext`, which other Python build tools are familiar with.

Another [command]( https://github.com/pypa/setuptools/tree/main/setuptools/command ) could be used other than `build_ext` if there's an appropriate one (though `build_ext` sounds right as it is used to compile non-Python code).

One could parse args manually in `setup.py` and use them. This is a bit hacky, but could be simple to implement.

**Additional context**

Came from looking at how to build the Java components using `pip`.

charlesbluca:
This PR switches out the `MavenCommand` building our external Java for subclasses of `build_ext` and `install_lib`; this means that it will be run implicitly in commands like `pip install .`, which should also resolve issues with RTD not successfully building the package in #401

Closes #400

cc @jakirkham

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -37,7 +37,7 @@ jobs:
           pip list
           conda list
       - name: Build jar
-        run: python setup.py java
+        run: python setup.py build_ext
       - name: Build and publish package
         env:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -36,12 +36,10 @@ jobs:
           which python
           pip list
           conda list
-      - name: Build jar
-        run: python setup.py build_ext
-      - name: Build and publish package
+      - name: Build package
+        run: python setup.py sdist bdist_wheel
+      - name: Publish package
         env:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
-        run: |
-          python setup.py sdist bdist_wheel
-          twine upload dist/*
+        run: twine upload dist/*
--- a/README.md
+++ b/README.md
@@ -121,7 +121,7 @@ After that, you can install the package in development mode
 
     pip install -e ".[dev]"
 
-To compile the Java classes (at the beginning or after changes), run
+To recompile the Java classes after changes have been made to the source contained in `planner/`, run
 
     python setup.py build_ext
 
--- a/README.md
+++ b/README.md
@@ -123,7 +123,7 @@ After that, you can install the package in development mode
 
 To compile the Java classes (at the beginning or after changes), run
 
-    python setup.py java
+    python setup.py build_ext
 
 This repository uses [pre-commit](https://pre-commit.com/) hooks. To install them, call
 
--- a/continuous_integration/gpuci/build.sh
+++ b/continuous_integration/gpuci/build.sh
@@ -48,7 +48,7 @@ python -m pip install git+https://github.com/dask/distributed
 
 gpuci_logger "Install dask-sql"
 pip install -e ".[dev]"
-python setup.py java
+python setup.py build_ext
 
 gpuci_logger "Check Python version"
 python --version
--- a/continuous_integration/gpuci/build.sh
+++ b/continuous_integration/gpuci/build.sh
@@ -48,7 +48,6 @@ python -m pip install git+https://github.com/dask/distributed
 
 gpuci_logger "Install dask-sql"
 pip install -e ".[dev]"
-python setup.py build_ext
 
 gpuci_logger "Check Python version"
 python --version
--- a/continuous_integration/recipe/meta.yaml
+++ b/continuous_integration/recipe/meta.yaml
@@ -16,8 +16,7 @@ build:
     - dask-sql-server = dask_sql.server.app:main
     - dask-sql = dask_sql.cmd:main
   string: py_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
-  script_env:
-    - PYTHON
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
 
 requirements:
   build:
--- a/continuous_integration/recipe/meta.yaml
+++ b/continuous_integration/recipe/meta.yaml
@@ -16,7 +16,7 @@ build:
     - dask-sql-server = dask_sql.server.app:main
     - dask-sql = dask_sql.cmd:main
   string: py_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
-  script: {{ PYTHON }} -m pip install . --no-deps -vv
+  script: {{ PYTHON }} -m pip install . --global-option=build_ext --no-deps -vv
 
 requirements:
   build:
--- a/continuous_integration/recipe/meta.yaml
+++ b/continuous_integration/recipe/meta.yaml
@@ -16,7 +16,7 @@ build:
     - dask-sql-server = dask_sql.server.app:main
     - dask-sql = dask_sql.cmd:main
   string: py_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
-  script: {{ PYTHON }} -m pip install . --global-option=build_ext --no-deps -vv
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
 
 requirements:
   build:
--- a/setup.py
+++ b/setup.py
@@ -4,12 +4,10 @@ import subprocess
 import sys
 
 from setuptools import find_packages, setup
+from setuptools.command.build_ext import build_ext as build_ext_orig
+from setuptools.extension import Extension
 
 import versioneer
-
-cmdclass = versioneer.get_cmdclass()
-build_ext_orig = cmdclass["build_ext"]
-build_py_orig = cmdclass["build_py"]
 
 
 class build_ext(build_ext_orig):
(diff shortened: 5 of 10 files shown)


setup.py
```diff
@@ -1,44 +1,28 @@
...
-    description = "run the mvn install command"
-    user_options = []
-
-    def initialize_options(self):
-        """No options"""
-        pass
-
-    def finalize_options(self):
-        """No options"""
-        pass
```
jakirkham, line 24: If we want to have an option to enable/disable here, we could leverage these to provide that for us. No strong opinion on whether or not we do this

jakirkham (commented):
Thanks Charles! 😄

Had one comment below

setup.py
charlesbluca: Yeah I think something like that would be nice if we ever want to install the Python bindings alone

.github/workflows/release.yml
```diff
@@ -37,7 +37,7 @@ jobs:
           pip list
           conda list
       - name: Build jar
-        run: python setup.py java
+        run: python setup.py build_ext
```
jakirkham, line 40: This could be fused with the step below

continuous_integration/gpuci/build.sh
```diff
@@ -48,7 +48,7 @@ python -m pip install git+https://github.com/dask/distributed
 
 gpuci_logger "Install dask-sql"
 pip install -e ".[dev]"
-python setup.py java
+python setup.py build_ext
```
jakirkham, line 51: This should already happen in `pip install`. So this can be dropped

```suggestion
```

jakirkham (commented):
Had a couple small suggestions about consolidations. These can probably be implemented in a few places, but only highlighted a couple

.github/workflows/release.yml
charlesbluca: Good catch! Yeah I wanted to go through and consolidate all these instances where we're installing the Java twice

continuous_integration/recipe/meta.yaml
```diff
@@ -16,8 +16,7 @@ build:
     - dask-sql-server = dask_sql.server.app:main
     - dask-sql = dask_sql.cmd:main
   string: py_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
-  script_env:
-    - PYTHON
+  script: {{ PYTHON }} -m pip install . --no-deps -vv
```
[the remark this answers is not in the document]
jakirkham: Right would suggest this

```suggestion
  script: {{ PYTHON }} -m pip install . --global-option=build_ext --no-deps -vv
```

setup.py
```diff
@@ -1,44 +1,28 @@
...
             )
         command = [maven_command, "clean", "install", "-f", "pom.xml"]
-        self.announce(f"Running command: {' '.join(command)}", level=distutils.log.INFO)
 
         subprocess.check_call(command, cwd="planner")
 
-        # Copy the artifact. We could also make maven do that,
-        # but in this way we have full control from python
         os.makedirs("dask_sql/jar", exist_ok=True)
         shutil.copy("planner/target/DaskSQL.jar", "dask_sql/jar/DaskSQL.jar")
```
jakirkham, line 27: Think we still need to call the base class's `run`

```suggestion
        shutil.copy("planner/target/DaskSQL.jar", "dask_sql/jar/DaskSQL.jar")

        super().run()
```

continuous_integration/recipe/meta.yaml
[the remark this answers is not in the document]
charlesbluca: I'm pretty sure we're running the install - I see a `setup.py install` being called in the RTD builds, it's just excluding the `build_ext` step:

https://readthedocs.org/projects/dask-sql/builds/16131690/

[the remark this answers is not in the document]
charlesbluca: One potential option (though I imagine it's not ideal) could be to alias the install command with something like

```properties
[aliases]
install = build_ext install
```

in the `setup.cfg`, which would make it so `setup.py install` would implicitly run `build_ext`

setup.py
```diff
@@ -1,47 +1,33 @@
...
-    user_options = []
-
-    def initialize_options(self):
-        """No options"""
-        pass
-
-    def finalize_options(self):
-        """No options"""
-        pass
+class build_ext(build_ext_orig):
```
jakirkham, line 12: Was just thinking since we don't have any `Extension`s specified, this might be causing `build_ext` to not run. We could either add an empty `Extension` or move this to some other step like `build`

charlesbluca: Down to try out either of these approaches

> some other step like `build`

Is there a generic `build` step? Didn't see anything in [the setuptools commands](https://github.com/pypa/setuptools/tree/main/setuptools/command)

jakirkham: I think [`build` comes from `distutils`]( https://github.com/pypa/setuptools/blob/6193a69d8608b6ca7e0e5ac1223d63abb55eee3b/setuptools/_distutils/command/build.py#L16 ), but `setuptools` may also alias this somewhere. Alternatively we could try using `build_py`.

Though maybe as a first step we can try adding an empty `Extension`?

charlesbluca: Tried this out in https://github.com/dask-contrib/dask-sql/pull/406/commits/3e5c3a95b7dbe9304e0182bcd07bf57fdd12f3f0, looks like that successfully triggers the Java build but the JAR doesn't end up getting copied successfully to its expected location

jakirkham: That's good to hear 🙂

Think we need to add it to [`MANIFEST.in`]( https://github.com/dask-contrib/dask-sql/blob/ef0fa16b23cac489298c492035dcf744e7b7e351/MANIFEST.in )

jakirkham: Another option might be to override the `get_outputs` method of this `build_ext` subclass to specify those file paths

charlesbluca:
It turns out we may also need `build_ext` extended as well to handle the editable install, which does not end up calling `install_lib`.

I've opted to include both commands, and factor out as much shared logic between the two as possible into `install_java_libraries` 🙂

setup.py
```diff
@@ -5,29 +5,55 @@
...
+        super().build()
 
-        subprocess.check_call(command, cwd="planner")
+        # copy java source to build directory
+        self.copy_tree("planner", os.path.join(self.build_dir, "planner"))
 
-        os.makedirs("dask_sql/jar", exist_ok=True)
-        shutil.copy("planner/target/DaskSQL.jar", "dask_sql/jar/DaskSQL.jar")
+        # build java in build directory
+        install_java_libraries(self.build_dir)
```
jakirkham, line 53: Does this result in building the JAR twice given `mvn` is called in this function?

charlesbluca: Now that I've removed the empty extension, `build_ext` should only be called on editable installs where `install_lib` is not called - so I think we should be safe 🙂 

Can verify that running both

```
pip install -e . --no-deps -vv
pip install . --no-deps -vv
```

I only see the jar get built once - are there any other commands we should look out for?

charlesbluca:
Hmm looks like this somehow broke the RTD builds, will try to recreate locally

setup.py
```diff
@@ -55,6 +55,9 @@ def build(self):
         # remove java source as it doesn't need to be packaged
         shutil.rmtree(os.path.join(self.build_dir, "planner"))
 
+        # copy jar to source directory so dask-sql can be imported from there
+        self.copy_tree(os.path.join(self.build_dir, "dask_sql/jar"), "dask_sql/jar")
```
jakirkham, line 59: Maybe we can check the [`READTHEDOCS` environment variable]( https://docs.readthedocs.io/en/stable/builds.html#build-environment ) before doing this?

charlesbluca: Great idea! Was considering setting some sort of RTD flag but didn't realize there was one in place 🙂

charlesbluca: I'm pretty sure this should be okay as an RTD-only fix - the niche case it would cover is allowing users to import from within the source repo if they use `python setup.py install` (this wouldn't work for pip installs since those copy the source directory to a temp folder), and I would imagine the majority of users / developers working directly within the source repo would probably want an editable install where this isn't an issue.

If anyone opens an issue expressing difficulty with that specific scenario, it shouldn't be an issue to make it a general fix

setup.py
```diff
@@ -1,46 +1,63 @@
...
-        subprocess.check_call(command, cwd="planner")
+        # remove java source as it doesn't need to be packaged
+        shutil.rmtree(os.path.join(self.build_dir, "planner"))
 
-        # Copy the artifact. We could also make maven do that,
-        # but in this way we have full control from python
-        os.makedirs("dask_sql/jar", exist_ok=True)
-        shutil.copy("planner/target/DaskSQL.jar", "dask_sql/jar/DaskSQL.jar")
+        # copy jar to source directory for RTD builds to API docs build correctly
+        if os.environ.get("READTHEDOCS", False):
```
jakirkham, line 59: Worth noting this will likely be a `str` (IOW `"True"`). This should still work as that is a non-empty `str`. Just thought it was worth pointing out

jakirkham:
```suggestion
        if os.environ.get("READTHEDOCS", "False") == "True":
```

.github/workflows/release.yml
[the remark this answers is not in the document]
charlesbluca: Yup, we no longer need to build the jar separately here - will update the workflow steps

README.md
```diff
@@ -123,7 +123,7 @@ After that, you can install the package in development mode
 
 To compile the Java classes (at the beginning or after changes), run
 
-    python setup.py java
+    python setup.py build_ext
```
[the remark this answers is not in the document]
charlesbluca: I think it could be useful to keep this command in the docs for development purposes, as there are situations where a developer may only want to recompile the Java classes - I do think that the preceding line could be edited to make that more clear though

jakirkham (approved):
Looks good. Thanks Charles! 😀

PR #408, opened 2022-02-23
Drop `common` column from result of cross join, remove from corresponding `meta`

charlesbluca:
Follow up to #300

In that PR, I assumed we could make the results of cross join operation consistent with the `meta` we construct for it by simply adding the `common` column we join on to the `meta`. I now understand that this doesn't work, as we don't necessarily know where the `common` column will be in the result, and the `meta` must have the same column ordering as its dataframe.

This PR resolves this by taking my original approach and just dropping the `common` column from the resulting dataframe altogether.


charlesbluca:
Think the failures here should be unblocked by https://github.com/dask-contrib/dask-sql/pull/420

PR #409, opened 2022-02-23
Fix use of row UDFs at intermediate query stages

brandon-b-miller:
Row UDFs in `dask-sql` are analogous to the functions expected by `pandas.DataFrame.apply` an expect a row of data containing all the scalars in a single row. The UDF accesses these scalars from the row via the corresponding column labels within the function:

```python
def f(row):
    x = row['a'] * 2
    y = row['b']
    return x - y
```

This works fine if the dataframe that calls the `apply` from dask in the end actually contains columns named `a` and `b`. This is however likely only the case for simple queries and fails if any columns are aliased during more complex operations, such as joins. 

This PR proposes to solve the problem by retaining the names of the original variables the function was registered with and reapplying them to the data later, just before calling `apply`.

### DIFF
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -215,6 +215,7 @@ class UDF:
             df = column_args[0].to_frame()
             for col in column_args[1:]:
                 df[col.name] = col
+            df.columns = self._names
             result = df.apply(
                 self.func, axis=1, args=tuple(scalar_args), meta=self.meta
             ).astype(self.meta[1])
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -215,8 +215,8 @@ class UDF:
                 else:
                     scalar_args.append(operand)
 
-            df = column_args[0].to_frame()
-            for name, col in zip(self.names, column_args):
+            df = column_args[0].to_frame(self.names[0])
+            for name, col in zip(self.names[1:], column_args[1:]):
                 df[name] = col
 
             result = df.apply(
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -183,7 +183,7 @@ class DataContainer:
 
 
 class UDF:
-    def __init__(self, func, row_udf: bool, return_type=None):
+    def __init__(self, func, row_udf: bool, params, return_type=None):
         """
         Helper class that handles different types of UDFs and manages
         how they should be mapped to dask operations. Two versions of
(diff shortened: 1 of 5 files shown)


dask_sql/datacontainer.py
```diff
@@ -212,9 +214,11 @@ def __call__(self, *args, **kwargs):
                     column_args.append(operand)
                 else:
                     scalar_args.append(operand)
+
             df = column_args[0].to_frame()
-            for col in column_args[1:]:
-                df[col.name] = col
+            for name, col in zip(self.names, column_args):
```
charlesbluca, line 219: If we pass the first parameter column name to `to_frame`, we lose a layer off the resulting HLG and don't have to deal with a superfluous column:

```
            df = column_args[0].to_frame(self.names[0])
            for name, col in zip(self.names[1:], column_args[1:]):
```

tests/integration/test_function.py
```diff
@@ -40,6 +41,27 @@ def f(row):
...
+        f, "f", [("x", np.int64), ("y", np.int64)], np.int64, row_udf=True
+    )
+
+    return_df = c.sql(f"SELECT F({colname_x},{colname_y}) FROM df_wide")
+
+    return_df = return_df.compute()
+    expect = df_wide[colname_x] + df_wide[colname_y]
+    got = return_df[return_df.columns[0]]
+
+    assert_series_equal(expect, got, check_names=False)
```
charlesbluca, line 62: Can we use `dd.assert_eq` here instead? In general hoping to move to that as the standard frame assertion function with https://github.com/dask-contrib/dask-sql/pull/365:

```suggestion
    expect = df_wide[colname_x] + df_wide[colname_y]
    got = return_df.iloc[:, 0]

    dd.assert_eq(expect, got, check_names=False)
```

charlesbluca (changes requested):
Looks good 🙂 just a couple suggestions:

dask_sql/datacontainer.py
```diff
@@ -212,9 +214,11 @@ def __call__(self, *args, **kwargs):
                     column_args.append(operand)
                 else:
                     scalar_args.append(operand)
+
             df = column_args[0].to_frame()
```
charlesbluca, line 218:
```suggestion
            df = column_args[0].to_frame(self.names[0])
```

dask_sql/datacontainer.py
charlesbluca, line 219:
```suggestion
            for name, col in zip(self.names[1:], column_args[1:]):
```

charlesbluca (changes requested):
Some suggestions for my first review:

dask_sql/datacontainer.py
brandon-b-miller: thanks -this was masking another bug as well.

PR #419, opened 2022-03-08
Optimize fast codepath for applying limits

charlesbluca:
Often, we will specify a `LIMIT` on the result of a complex query, which can potentially leave us with a Dask dataframe with several empty partitions (in my case, I encountered this with cross-join operations). In these cases, checking the length of the dataframe's first partition doesn't really give us a good idea of if we can apply a `LIMIT` on the dataframe with minimal persists using `head()`, as the first partition actually containing data could be the Nth partition of the dataframe.

This PR resolves this by expanding the partition length check to grab the length of the first _nonempty_ partition; this new check ideally shouldn't result in any significant extra persists, as the only additional dataframes we are loading into memory are empty.

cc @ayushdg @randerzander

### DIFF
--- a/dask_sql/physical/rel/logical/limit.py
+++ b/dask_sql/physical/rel/logical/limit.py
@@ -59,9 +59,13 @@ class DaskLimitPlugin(BaseRelPlugin):
         function, which is not possible with normal "map_partitions".
         """
         if not offset:
-            # We do a (hopefully) very quick check: if the first partition
-            # is already enough, we will just use this
-            first_partition_length = len(df.partitions[0])
+            # get length of first nonempty partition
+            for partition in df.partitions:
+                first_partition_length = len(partition)
+                if first_partition_length:
+                    break
+
+            # if this partition contains the requested data, return
             if first_partition_length >= end:
                 return df.head(end, compute=False)
 
--- a/dask_sql/physical/rel/logical/limit.py
+++ b/dask_sql/physical/rel/logical/limit.py
@@ -59,15 +59,16 @@ class DaskLimitPlugin(BaseRelPlugin):
         function, which is not possible with normal "map_partitions".
         """
         if not offset:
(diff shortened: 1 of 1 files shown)


charlesbluca:
Also, I'm kinda scratching my head as to why we're doing this check in the first place - perhaps I need to look into Dask's implementation of `head()`, but I would assume that if we don't have to apply an `OFFSET` that `head()` would always be the optimal way to grab the requested data in terms of persists?

dask_sql/physical/rel/logical/limit.py
```diff
@@ -59,11 +59,16 @@ def _apply_offset(self, df: dd.DataFrame, offset: int, end: int) -> dd.DataFrame
...
-            # is already enough, we will just use this
-            first_partition_length = len(df.partitions[0])
-            if first_partition_length >= end:
-                return df.head(end, compute=False)
+            # compute the number of partitions that need to be persisted in head
+            npartitions = 0
+            nrows = 0
+            for partition in df.partitions:
+                npartitions += 1
+                nrows += len(partition)
```
ayushdg, line 67: I might be wrong here but iiuc `len(partition)` is an eager operation, assuming we have a join followed by a `limit`, would dask compute the join `n` number of times until it found a non empty partition in this case?

charlesbluca: Hmm that's a good point, and perhaps why there was originally a call to `persist` preceding this check that we removed in https://github.com/dask-contrib/dask-sql/pull/387 (to avoid a duplicate computation in the case that `head()` doesn't suffice).

I wonder how bad it would be if we simply made a call to `df.head(end, npartitions=-1, compute=False)` here? Obviously that would trigger computations on partitions that might not be needed, but it would also push all the persisting to the `compute()` call instead of doing it here.

charlesbluca:
Superceded by #696

PR #431, opened 2022-03-14
Allow UDFs to be decorated at call time

brandon-b-miller:
This PR allows `UDF` to be initialized with a decorator which is then used at runtime to wrap the function, and wraps all UDFs with `make_pickable_without_dask_sql`. This avoids subtle issues where the particulars of the function being applied cause the workers to kick off work that interferes with complex workloads.


brandon-b-miller:
After some thought and discussion this approach probably isn't worth doing in favor of more global solutions to the JVM import issue. Closing this.

PR #433, opened 2022-03-15
Add basic predicate-pushdown optimization

rjzamora:
This PR adds a predicate-pushdown optimization for `WHERE` statements when the table originates from `dask.dataframe.read_parquet`.  Note that the logic used here is very similar to [dask#8810](https://github.com/dask/dask/pull/8810).

**TODO**:

- [x] Add test coverage
- [x] Add a mechanism to disable the new predicate pushdown optimization?
- [x] Decide if these changes belong in Dask-SQL, or if we should take the time to get them in dask/dask [**Update**: I think it makes a lot of sense to merge the optimization here first to make it easier to iterate/improve where it matters most.]

**Example**:

```python
import os
import pandas as pd
import dask.dataframe as dd
from dask import config
from dask.utils_test import hlg_layer
from dask_sql import Context

path = "./tmpdir"
if not os.path.isdir(path):
   df = pd.DataFrame({"a": [1, 2, 3] * 5, "b": range(15), "c": ["A"]*15})
   ddf0 = dd.from_pandas(df, npartitions=3)
   ddf0.to_parquet(path)
   
ddf = dd.read_parquet(path)
c = Context()
c.create_table("my_table", ddf)

df = c.sql("SELECT a FROM my_table WHERE (b > 5 AND b < 10) OR a = 1")
hlg_layer(df.dask, "read-parquet").creation_info['kwargs']['filters']
```
**Output**: `[[('b', '>', 5), ('b', '<', 10)], [('a', '==', 1)]]`


rjzamora:
cc @randerzander

tests/integration/test_filter.py
```diff
@@ -122,3 +123,49 @@ def test_filter_year(c):
...
+
+    # Check computed result is correct.
+    # Note that we must sort by index to ensure
+    # ordering can be compared directly
+    return_df = return_df.compute()
+    df = parquet_ddf.compute()
+    expected_df = df_func(df)
+    assert_frame_equal(
+        return_df.sort_index(), expected_df.sort_index(),
+    )
```
charlesbluca, line 171: Would it be possible to do this assertion using `assert_eq`? Or do we need a serial dataframe for `df_func`?

```suggestion
    expected_df = df_func(parquet_ddf)
    assert_frame_equal(
        return_df.sort_index(), expected_df.sort_index(),
    )
```

For context, trying to move over to `assert_eq` wherever possible in https://github.com/dask-contrib/dask-sql/pull/365 🙂

charlesbluca (commented):
Thanks for the work here @rjzamora 🙂 I personally don't see anything wrong with maintaining a downstream implementation of predicate pushdown while we wait for decisions to be made upstream, especially if it'll help provide some insights on what could be done better there.

> Add a mechanism to disable the new predicate pushdown optimization?

Would it make sense to enable / disable through Dask config, and have a sensible default in the [YAML config file](https://github.com/dask-contrib/dask-sql/blob/main/dask_sql/sql.yaml)?

dask_sql/physical/rel/logical/filter.py
```diff
@@ -31,7 +32,7 @@ def filter_or_scalar(df: dd.DataFrame, filter_condition: Union[np.bool_, dd.Seri
 
     # In SQL, a NULL in a boolean is False on filtering
     filter_condition = filter_condition.fillna(False)
-    return df[filter_condition]
+    return attempt_predicate_pushdown(df[filter_condition])
```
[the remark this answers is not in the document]
charlesbluca: Ha just saw this - [in my review](https://github.com/dask-contrib/dask-sql/pull/433#pullrequestreview-911971136) I pointed out that it might make sense to have this controlled through Dask's config module, with a default set in our own config YAML

dask_sql/physical/rel/logical/optimize.py
```diff
@@ -0,0 +1,315 @@
...
+    # Check that we have a single IO layer with `filters` support
+    io_layer = []
+    for k, v in dsk.layers.items():
+        if isinstance(v.layer, DataFrameIOLayer):
+            io_layer.append(k)
+            if (
+                "filters" not in v.creation_info.get("kwargs", {})
+                or v.creation_info["kwargs"]["filters"] is not None
+            ):
+                # No filters support, or filters is already set
```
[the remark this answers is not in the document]
rjzamora: Right - We can probably do this, but I think the logic could get a bit tricky. So, in the interest of avoiding logic bugs, I am just assuming that the user-defined `filters` argument is already correct/sufficient. Maybe we could just address this in follow-up work if resetting the filters turns out to be important?

tests/integration/test_filter.py
rjzamora: Yep - `assert_eq` makes sense, but I was just going with the pattern used by other tests :)

rjzamora:
>I personally don't see anything wrong with maintaining a downstream implementation of predicate pushdown while we wait for decisions to be made upstream

Right - I think we will have the most success iterating and improving on predicate-pushdown in the short term if it lives in dask-sql for now.  However, just in case others felt differently, I updated [dask#8810](https://github.com/dask/dask/pull/8810) to be almost identical to this PR (but the optimization is turned off by default). Therefore, we could also try to get this same logic merged in dask, and simply "use" `attempt_predicate_pushdown` in dask-sql (or set the appropriate `dask.config` options).

dask_sql/physical/rel/logical/optimize.py
```diff
@@ -0,0 +1,349 @@
...
+                    "Predicate pushdown optimization skipped. The IO "
+                    "layer does not support a `filters` argument, or "
+                    "`filters` was already populated."
+                )
+                return ddf
+    if len(io_layer) != 1:
+        # Not a single IO layer
+        logger.warning(
+            f"Predicate pushdown optimization skipped. {len(io_layer)} "
+            f"IO layers detected, but only one IO layer is allowed."
```
ayushdg, line 62: Is there a way to improve this message? In cases where my table was created directly from a pandas/Dask DataFrame, the warning here can be a bit confusing `0 IO layers detected, but only one is allowed`

dask_sql/physical/rel/logical/optimize.py
```diff
@@ -0,0 +1,349 @@
...
+
+    # We were able to extract a DNF filter expression.
+    # Check that we have a single IO layer with `filters` support
+    io_layer = []
+    for k, v in ddf.dask.layers.items():
+        if isinstance(v, DataFrameIOLayer):
+            io_layer.append(k)
+            if (
+                "filters" not in v.creation_info.get("kwargs", {})
+                or v.creation_info["kwargs"]["filters"] is not None
```
ayushdg, line 49: In cases where this was created from  a `dd.read_csv` call, `v.creation_info` returns a noneType

ayushdg (changes requested):
Thanks @rjzamora for working a feature that's exteremly valuable for sql.
Adding my initial review while things are still a wip. 

I also found cases where the predicate filters don't get generated as expected?
```python
df = c.sql("SELECT a FROM my_table WHERE (my_table.b > 3 AND my_table.b < 10 OR my_table.a = 1) AND (my_table.c = 'A')")
```

dask_sql/physical/rel/logical/optimize.py
rjzamora: Ah, oops - I just introduced this bug today when I moved this IO-layer check to happen first.

dask_sql/physical/rel/logical/optimize.py
rjzamora: Perhaps we don't really need any warning here, since this is likely to happen a lot (pretty much any time the root IO layer is not read_parquet")?

rjzamora:
>I also found cases where the predicate filters don't get generated as expected?

>df = c.sql("SELECT a FROM my_table WHERE (my_table.b > 3 AND my_table.b < 10 OR my_table.a = 1) AND (my_table.c = 'A')")

Hmm - Good point.  I'm not sure how to generate a filter expression for logical operations that don't directly map to disjunctive normal form :/. I guess we would need a utility to parse the final filter expressoin and covert it to DNF.

rjzamora:
@ayushdg - I think I addressed the critical problems here.

Regarding your comment about `in` statements being supported by pyarrow. The current PR covers an sql "IN" statement, but indeed uses the expanded form generated by dask_sql.  Are you saying that we should try to find a way to use "in" directly?

rjzamora:
@ayushdg - Do you have any futher concerns here? Once we are confident that this optimization is delivering what dask-sql needs, we should try to move it upstream. However, I di think dask-sql is the best place for it to live at first.

dask_sql/physical/rel/logical/filter.py
```diff
@@ -1,11 +1,13 @@
 import logging
 from typing import TYPE_CHECKING, Union
 
+import dask.config as dask_config
 import dask.dataframe as dd
 import numpy as np
 
 from dask_sql.datacontainer import DataContainer
 from dask_sql.physical.rel.base import BaseRelPlugin
+from dask_sql.physical.rel.logical.optimize import attempt_predicate_pushdown
```
charlesbluca, line 10: This is a minor nitpick, but since optimize.py doesn't actually contain any plugins to convert from SQL -> Dask, and at least for now I assume it is only going to be used within the scope of filter operations, I think it would make sense to relocate it to `dask_sql.physical.utils.filter` (though we could go for `optimize` if we expect these utils to be used in other places later on).

As an example, consider the relationship between:

- https://github.com/dask-contrib/dask-sql/blob/main/dask_sql/physical/rel/logical/sort.py
- https://github.com/dask-contrib/dask-sql/blob/main/dask_sql/physical/utils/sort.py

I'll also note that I'm really not sure if this is the best way to be doing things, but keeping it consistent for now should make it easier to refactor if/when the time comes 🙂

tests/integration/fixtures.py
```diff
@@ -112,6 +112,30 @@ def datetime_table():
...
+            "d": [
+                pd.Timestamp("2013-08-01 23:00:00"),
+                pd.Timestamp("2014-09-01 23:00:00"),
+                pd.Timestamp("2015-10-01 23:00:00"),
+            ]
+            * 5,
+            "index": range(15),
+        },
+    )
+    dd.from_pandas(df, npartitions=3).to_parquet(os.path.join(tmpdir, "parquet"))
```
charlesbluca, line 133: Not exactly sure why, but looks like this `from_pandas` call is resulting in killed workers in the independent cluster testing 😕 I can look into this locally and follow up with more info

tests/integration/test_filter.py
```diff
@@ -1,7 +1,7 @@
 import dask.dataframe as dd
 import pandas as pd
 import pytest
-from pandas.testing import assert_frame_equal
```
charlesbluca, line 4: I know I said earlier to switch over to `assert_eq` here, but after learning about the default behavior of `assert_eq` in https://github.com/dask-contrib/dask-sql/pull/365#issuecomment-1068282514, I think we should probably hold off on this switch until https://github.com/dask/dask/pull/8811 is out in a Dask release and #365 is merged in 😓 would you mind switching these back to `assert_frame_equal`?

charlesbluca (commented):
Some additional thoughts; I'm going to be looking into the test failures to see if I can find a clear cause:

ayushdg (approved):
> @ayushdg - Do you have any futher concerns here? Once we are confident that this optimization is delivering what dask-sql needs, we should try to move it upstream. However, I di think dask-sql is the best place for it to live at first.

I don't have further concerns here. I was able to confirm that filters were being passed down even in slightly more complex cases where two tables were being joined and each table was filtered on a few predicates 😄

tests/integration/fixtures.py
rjzamora: Would it help to use pandas directly for the write here?

charlesbluca: No we should be good with this method - the failures would likely arise with other fixtures anyways. Hopefully this should get resolved with https://github.com/dask-contrib/dask-sql/pull/437

rjzamora: Okay, sounds good. I'll assume this is good to go then - Let me know if I should address anything else :)

charlesbluca (approved):
Things look good here to me 🙂 ideally would like to merge this in once CI is resolved and we can verify that everything passes

tests/integration/fixtures.py
```diff
@@ -114,6 +127,30 @@ def datetime_table():
     )
 
 
+@pytest.fixture(params=[pytest.param(None, marks=PARQUET_MARK)])
```
charlesbluca, line 130:
```suggestion
@pytest.fixture()
```

tests/integration/fixtures.py
```diff
@@ -188,6 +225,19 @@ def c(
...
+    # in case the required dependencies are
+    # not available.
+    from dask_sql.context import Context
+
+    c = Context()
+    c.create_table("parquet_ddf", parquet_ddf)
+
+    yield c
+
+
```
charlesbluca, line 240:
```suggestion
```

.github/docker-compose.yaml
```diff
@@ -11,5 +11,7 @@ services:
         container_name: dask-worker
         image: daskdev/dask:latest
         command: dask-worker dask-scheduler:8786
+        environment:
+            EXTRA_CONDA_PACKAGES: "pyarrow>1.0.0"
```
charlesbluca, line 15: Would probably add a comment here briefly summarizing the need - we'll probably consult this later when documenting cluster requirements:

```suggestion
            EXTRA_CONDA_PACKAGES: "pyarrow>1.0.0"  # required for parquet read/writes
```

charlesbluca (commented):
Just a few comments related to the fixture dependency check, which I think we can remove now that things have been addressed with the `EXTRA_CONDA_PACKAGES`:

tests/integration/fixtures.py
```diff
@@ -23,6 +23,19 @@
...
+    try:
+        import fastparquet as pq_dep
+    except ImportError:
+        pq_dep = None
+
+PARQUET_MARK = pytest.mark.skipif(
+    pq_dep is None, reason="Parquet dependencies not installed."
+)
+
+
```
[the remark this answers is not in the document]
rjzamora: Yes - We can remove this if pyarrow is required. Although 0.15.1 is a bit "old" at this point :)

charlesbluca (approved):

PR #437, opened 2022-03-22
Resolve independent cluster test failures

charlesbluca:
Fixes breakage we've been seeing in independent cluster testing due to:

- mismatched pandas versions (resolved with https://github.com/dask/dask-docker/pull/209)
- the bump to distributed 2022.3.0

In particular, one big change we've made is that we now use a single session-wide client to perform all computations or random client operations - backed either by a local cluster or an independent cluster for our independent cluster testing. I've updated the handful of tests that manage client lifecycle stuff to avoid this causing larger failures, but it would be nice at some point to explore better scope options for the `client` fixture.

cc @ayushdg


ayushdg (approved):
Thanks for the quick fix! This should be good in the short term for unblocking CI

PR #451, opened 2022-04-07
Add Java source code to source distribution

charlesbluca:
This should resolve the build issues we've been seeing in https://github.com/conda-forge/dask-sql-feedstock/pull/36 due to the Java source not being included in the tars on PyPI

cc @ayushdg

### DIFF
--- a/MANIFEST.in
+++ b/MANIFEST.in
@@ -1,4 +1,5 @@
 recursive-include dask_sql *.yaml
+recursive-include planner *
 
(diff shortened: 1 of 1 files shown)


ayushdg (approved):

charlesbluca:
We were able to resolve https://github.com/conda-forge/dask-sql-feedstock/pull/36 independently, but by merging this we shouldn't run into that issue in the future

PR #453, opened 2022-04-08
Bump `httpclient` dependency

charlesbluca:
Seemed like we were able to build from source with an updated `httpclient` version? I'm interested in how this runs in CI


ayushdg (approved):
Changes lgtm! 
@jdye64 It would be great if you could confirm that adding httpclient as an explicit dependency instead of getting it from avatica-core works in these cases.

galipremsagar (approved):

PR #482, opened 2022-04-20
Register table names with the original case in addition to lower case

### LINKED ISSUE #481: [BUG] dask-sql creates all tables as lower case.
The dask `create_table` method adds to the list of tables with [lower case](https://github.com/dask-contrib/dask-sql/blob/main/dask_sql/context.py#L218) which means that any queries that are executed must use the lower case name.

**What happened**:

All tables are registered as lower case, so I am unable to include any queries with upper case table names - which requires me to convert all queries to `lower()`.  

**What you expected to happen**:

I would expect that the case would be preserved when adding to schemas, or at least like in the case of #177 there would be an original and lower case verison.

See the following code snippet that  creates a dask dataframe, and then creates two tables, one with `lower_case` name, and another with `UPPER_CASE`. 

```python
from dask_sql import Context
c = Context()

ddf = dd.from_pandas(pd.DataFrame({"key": ["value"]}), npartitions=1)

# create table with lower case
c.create_table("lower_case", ddf)
print(c.schema["root"].tables) 
c.sql("SELECT * FROM lower_case")

# Create table with upper case
c.create_table("UPPER_CASE", ddf)
print(c.schema["root"].tables) 
c.sql("SELECT * FROM UPPER_CASE")
```
The second call to print the tables lists both as lower case:
```
{'lower_case': <dask_sql.datacontainer.DataContainer object at 0x2d8432970>, 'upper_case': <dask_sql.datacontainer.DataContainer object at 0x2d49d9670>}
```

And the second select fails to find the table with UPPER CASE NAME
```
ParsingException: Can not parse the given SQL: From line 1, column 15 to line 1, column 22: Object 'MY_TABLE' not found within 'root'; did you mean 'my_table'?
```

**Anything else we need to know?**:

Perhaps there is a reason why everything was made `lower()` - but I can't seem to find it in the [docs](https://dask-sql.readthedocs.io/en/latest/search.html?q=lower).

**Environment**:

- dask-sql version: `2022.4.1`
- Python version: `3.9`
- Operating System: `mac-osx arm64`
- Install method (conda, pip, source): `conda`

brightsparc:
Closes #481

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -215,8 +215,12 @@ class Context:
             gpu=gpu,
             **kwargs,
         )
+        # Register table names with the original case in addition to lower case name.
+        # see: https://github.com/dask-contrib/dask-sql/issues/481
+        self.schema[schema_name].tables[table_name] = dc
         self.schema[schema_name].tables[table_name.lower()] = dc
         if statistics:
+            self.schema[schema_name].statistics[table_name] = statistics
             self.schema[schema_name].statistics[table_name.lower()] = statistics
 
     def register_dask_table(self, df: dd.DataFrame, name: str, *args, **kwargs):


GPUtester:

quasiben:
ok to test

Issue #483, opened 2022-04-20
[DF] Wrap Dask-SQL python exception in Rust with pyo3


jdye64:
**Is your feature request related to a problem? Please describe.**
Currently when DataFusion encounters invalid SQL an `Result` with the `Err` field is generated on the Rust side. However to mimic existing functionality it would be better if we wrapped that result in an actual Python exception that can be caught on the Python side and handled appropriately.

**Describe the solution you'd like**
Wrap `Result` instances with `Err` field populated, an error, in a pyo3 `PyException` that can be sent back as an exception to the Python layer.

**Describe alternatives you've considered**
None

**Additional context**
None

Issue #468, opened 2022-04-12
[DF] Port Select logic to Datafusion


jdye64:
**Is your feature request related to a problem? Please describe.**
Update all existing "Select" logic to replace Apache Calcite data structures with Apache Datafusion data structures.

**Describe the solution you'd like**
Replace instances of Calcite [RelNode](https://calcite.apache.org/javadocAggregate/org/apache/calcite/rel/RelNode.html) with correlating instances of Datafusion [LogicalPlan](https://docs.rs/datafusion/latest/datafusion/logical_plan/plan/enum.LogicalPlan.html). This includes any base classes, functions, members that might also rely on that logic.

**Describe alternatives you've considered**
None

**Additional context**
Success criteria is for all pytests in test_select.py to be passing

PR #471, opened 2022-04-12
Datafusion aggregate

jdye64:
Adds the support for Aggregate and Group By in queries. 

This closes #464 
This closes #472
This closes #478

PR #489, opened 2022-04-20
Basic DataFusion Select Functionality

jdye64:
Basic SELECT functionality working, PyTests that are not working due to other issues are referenced in the `@pytest.mark.skip(...)`

This closes: #468 
This closes: #490 
This closes: #496 
This closes: #483


jdye64:
It will be much easier to review this PR after [471](https://github.com/dask-contrib/dask-sql/pull/471) is merged. Feel free to wait until after that for a review.

dask_planner/src/sql/types.rs
```diff
@@ -1,107 +1,312 @@
...
+                            }
+                            // Default to Nanosecond which is common if not present
+                            None => TimeUnit::Nanosecond,
+                        };
+                        (unit, tz)
+                    }
+                    // Default to Nanosecond and None for tz which is common if not present
+                    None => (TimeUnit::Nanosecond, None),
+                };
+                DataType::Timestamp(unit, tz)
```
ayushdg, line 116: Do we want to pass the same timezone even when the type is a `timestamp` type without a timezone?

dask_sql/mappings.py
```diff
@@ -128,10 +123,10 @@ def sql_to_python_value(sql_type: str, literal_value: Any) -> Any:
...
 
-    elif sql_type.startswith("INTERVAL"):
+    elif sql_type == SqlTypeName.INTERVAL:
         # check for finer granular interval types, e.g., INTERVAL MONTH, INTERVAL YEAR
         try:
-            interval_type = sql_type.split()[1].lower()
+            interval_type = str(sql_type).split()[1].lower()
 
             if interval_type in {"year", "quarter", "month"}:
                 # if sql_type is INTERVAL YEAR, Calcite will covert to months
```
ayushdg, line 132: Some of these earlier assumptions about calcite may not hold true for datafusion and will need updating as and when mappings are encountered

dask_sql/mappings.py
```diff
@@ -270,9 +261,7 @@ def cast_column_type(
     """
     current_type = df[column_name].dtype
 
-    logger.debug(
-        f"Column {column_name} has type {current_type}, expecting {expected_type}..."
-    )
+    print(f"Column {column_name} has type {current_type}, expecting {expected_type}...")
```
ayushdg, line 264: Debug print

dask_sql/physical/rel/logical/project.py
```diff
@@ -78,7 +81,10 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
         # Make sure the order is correct
         cc = cc.limit_to(column_names)
 
-        cc = self.fix_column_to_row_type(cc, column_names)
+        cc = self.fix_column_to_row_type(cc, rel.getRowType())
         dc = DataContainer(df, cc)
-        dc = self.fix_dtype_to_row_type(dc, rel.table())
+        dc = self.fix_dtype_to_row_type(dc, rel.getRowType())
+
+        print(f"After Project: {dc.df.head(10)}")
```
ayushdg, line 88: Same as above

dask_sql/mappings.py
jdye64: Completely agree. Even though all the current tests are passing I am sure there are some missing edge cases. I was thinking about opening a issue just for that alone and was hoping that you might be able to help with that one since you have done a lot of work in the date/time area. Could I open an issue and assign it to you for a subsequent PR?

dask_planner/src/sql/types.rs
jdye64: I honestly don't know? You would know much better. Which way is the correct way?

ayushdg: A default timestamp (without timezone) based on postresql is interpreted as the timezone in utc irrespective of the offset provided. So in this case the tz should always be utc.

jdye64: That might have been handled by the library but either way I added a check just to be sure.

continuous_integration/recipe/meta.yaml
```diff
@@ -30,6 +30,7 @@ requirements:
     - setuptools-rust>=1.1.2
   run:
     - python
+    - setuptools-rust>=1.1.2
```
ayushdg, line 33: Based on the discussion [here](https://github.com/dask-contrib/dask-sql/pull/471#discussion_r853217325) is this needed?

continuous_integration/recipe/run_test.py
```diff
@@ -13,19 +13,21 @@
...
-        my_data.name,
-        SUM(my_data.x) AS "S"
-    FROM
-        my_data
-    GROUP BY
-        my_data.name
-"""
-)
-expect = pd.DataFrame({"name": ["Alice", "Bob", "Chris"], "S": [14850, 14950, 15050]})
+# This needs to be temprarily disabled since this query requires features that are not yet implemented
```
ayushdg, line 17: Out of curiosity what functionality is missing for this to work?

dask_planner/src/expression.rs
```diff
@@ -49,130 +58,69 @@ impl From<ScalarValue> for PyScalarValue {
...
+                                            proj.input.schema().index_of_column(col).unwrap();
+                                        // match &exprs[plan.get_index(col)] {
+                                        match &exprs[col_index] {
+                                            Expr::AggregateFunction { args, .. } => {
+                                                match &args[0] {
+                                                    Expr::Column(col) => {
+                                                        println!(
+                                                            "AGGREGATE COLUMN IS {}",
+                                                            col.name
+                                                        );
```
ayushdg, line 101: If the debug prints are useful as work progresses, it might be worth adding a `TODO` comment to clean up later

dask_sql/context.py
```diff
@@ -805,7 +807,11 @@ def _get_ral(self, sql):
                 f"Multiple 'Statements' encountered for SQL {sql}. Please share this with the dev team!"
             )
 
-        nonOptimizedRel = self.context.logical_relational_algebra(sqlTree[0])
+        try:
+            nonOptimizedRel = self.context.logical_relational_algebra(sqlTree[0])
+        except DFParsingException as pe:
+            raise ParsingException(sql, str(pe)) from None
```
ayushdg, line 813: If one doesn't exist already we should add a test case checking for this.

tests/unit/test_mapping.py
```diff
@@ -3,28 +3,32 @@
...
 
 def test_sql_to_python():
-    assert sql_to_python_value("CHAR(5)", "test 123") == "test 123"
-    assert type(sql_to_python_value("BIGINT", 653)) == np.int64
-    assert sql_to_python_value("BIGINT", 653) == 653
-    assert sql_to_python_value("INTERVAL", 4) == timedelta(milliseconds=4)
+    assert sql_to_python_value(SqlTypeName.VARCHAR, "test 123") == "test 123"
+    assert type(sql_to_python_value(SqlTypeName.BIGINT, 653)) == np.int64
+    assert sql_to_python_value(SqlTypeName.BIGINT, 653) == 653
+    assert sql_to_python_value(SqlTypeName.INTERVAL, 4) == timedelta(microseconds=4000)
```
ayushdg, line 24: Is the default datafusion unit in microseconds?

.github/workflows/test.yml
```diff
@@ -162,7 +162,6 @@ jobs:
       - name: Install dependencies and nothing else
         run: |
           conda install setuptools-rust
-          conda install pyarrow>=4.0.0
```
ayushdg, line 165: is the pyarrow requirement no longer needed?

ayushdg (changes requested):
Few minor questions but generally things look good to me

continuous_integration/recipe/meta.yaml
jdye64: No, it is not needed. I had based this PR off the `datafusion-aggregate` one and probably did that before removing this in the other PR. It can be removed and I will remove it.

dask_planner/src/expression.rs
jdye64: Yep, my plan is to not merge with upstream master during all of these commits. What that will give us is basically another shot to review before we merge into main as all of these type changes will stick out like red flags and can be changed then but at the same time can be left in for now to assist with current development work.

Long story short. 100% plan on removing them during that final PR

dask_sql/context.py
jdye64: Several already exist one coming top of mind is the `test_wrong_input` pytest in `test_select.py`

tests/unit/test_mapping.py
jdye64: Yes, however that can be changed and I should look into finding out how I can make it default to milliseconds instead

.github/workflows/test.yml
jdye64: Thankfully no. I found a portion of the DataFusion library where those mappings were already present so I just wrote a Rust function that I exposed to Python that allowed me to use that instead of including PyArrow in the Python layer since DataFusion already includes Arrow.

Previously I didn't know that portion of the DataFusion library existed so that was the reason for the original inclusion of PyArrow.

continuous_integration/recipe/run_test.py
jdye64: The `SUM()` operation has not yet been implemented as part of the `RexCall -> call.py` port

tests/unit/test_mapping.py
ayushdg: Actually, even if it needs a few changes on the python side my personal preference is to staying with this default and evaluating if we need to move to millis if it causes problems down the line

ayushdg (approved):
Changes lgtm! Thanks a lot @jdye64

PR #494, opened 2022-04-21
Bump DataFusion version

andygrove:
This PR updates the DataFusion dependency to pick up the work in https://github.com/apache/arrow-datafusion/pull/2294 which moved the logical plan into the `datafusion_expr` crate and also introduced a `TableSource` trait which is used in the logical plan instead of the `TableProvider` trait, which is for physical plans.

I also removed some downcast code that is no longer necessary now that `TableScan` contains the real table name and not the aliased name, which is what led to this downcast code being added as a workaround.


GPUtester:

charlesbluca:
add to allowlist

charlesbluca:
@andygrove feel free to ping me if you push additional commits as it will require workflow approval on each commit you make

andygrove:
thanks @charlesbluca .. yes I did need to push another commit after testing this locally

Issue #486, opened 2022-04-20
[DF] Handle SQL CAST operations


jdye64:
**Is your feature request related to a problem? Please describe.**
The current DataFusion implementation lacks the ability to properly handle CAST operations in a SQL string, they are ignored. The logic to handle a DataFusion [CAST](https://docs.rs/datafusion/latest/datafusion/logical_plan/enum.Expr.html#variant.Cast) needs to the added to the Dask-SQL Python code base.

**Describe the solution you'd like**
Properly RexConvert a SQL string containing a CAST operation

**Describe alternatives you've considered**
None

**Additional context**
None

jdye64:
resolved via #498

PR #498, opened 2022-04-24
Allow for Cast parsing and logicalplan

jdye64:
Closes: #486


jdye64:
FYI If reviewing I would wait until https://github.com/dask-contrib/dask-sql/pull/489 is merged before reviewing this PR

jdye64:
@galipremsagar, @charlesbluca, or @ayushdg do you see a reason why the `df.cast()` operation would be returning values that don't match the existing dataframe? 

[this test](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/tests/integration/test_select.py#L157) is failing and I have traced the issue to [mapping.py](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/dask_sql/mappings.py#L295) where the cast occurs. You can see in the run [here](https://github.com/dask-contrib/dask-sql/runs/6158189708?check_suite_focus=true#step:8:96) that the results are slightly different. Almost seems like a precision thing??

ayushdg:
> @galipremsagar, @charlesbluca, or @ayushdg do you see a reason why the `df.cast()` operation would be returning values that don't match the existing dataframe?
> 
> [this test](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/tests/integration/test_select.py#L157) is failing and I have traced the issue to [mapping.py](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/dask_sql/mappings.py#L295) where the cast occurs. You can see in the run [here](https://github.com/dask-contrib/dask-sql/runs/6158189708?check_suite_focus=true#step:8:96) that the results are slightly different. Almost seems like a precision thing??



> @galipremsagar, @charlesbluca, or @ayushdg do you see a reason why the `df.cast()` operation would be returning values that don't match the existing dataframe?
> 
> [this test](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/tests/integration/test_select.py#L157) is failing and I have traced the issue to [mapping.py](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/dask_sql/mappings.py#L295) where the cast occurs. You can see in the run [here](https://github.com/dask-contrib/dask-sql/runs/6158189708?check_suite_focus=true#step:8:96) that the results are slightly different. Almost seems like a precision thing??

I got a chance to debug this and the core issue seems to be that the new bindings interprets the first column (given as a timestamp with Berlin/Europe timezone) as a `SqlTypeName.DATE` type and as a consequence casts that column to UTC. 

The results for column1 have a mismatch because of this where the expected df was expecting the output to have the Berlin time but it gets UTC time instead. 
Is there a way of figuring out why the input sqlTypes are being inferred as DATE and not timestamp?

ayushdg:
> @galipremsagar, @charlesbluca, or @ayushdg do you see a reason why the `df.cast()` operation would be returning values that don't match the existing dataframe?
> 
> [this test](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/tests/integration/test_select.py#L157) is failing and I have traced the issue to [mapping.py](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/dask_sql/mappings.py#L295) where the cast occurs. You can see in the run [here](https://github.com/dask-contrib/dask-sql/runs/6158189708?check_suite_focus=true#step:8:96) that the results are slightly different. Almost seems like a precision thing??



> @galipremsagar, @charlesbluca, or @ayushdg do you see a reason why the `df.cast()` operation would be returning values that don't match the existing dataframe?
> 
> [this test](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/tests/integration/test_select.py#L157) is failing and I have traced the issue to [mapping.py](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/dask_sql/mappings.py#L295) where the cast occurs. You can see in the run [here](https://github.com/dask-contrib/dask-sql/runs/6158189708?check_suite_focus=true#step:8:96) that the results are slightly different. Almost seems like a precision thing??

I got a chance to debug this and the core issue seems to be that the new bindings interprets the first column (given as a timestamp with Berlin/Europe timezone) as a `SqlTypeName.DATE` type and as a consequence casts that column to UTC. 

The results for column1 have a mismatch because of this where the expected df was expecting the output to have the Berlin time but it gets UTC time instead. 
Is there a way of figuring out why the input sqlTypes are being inferred as DATE and not timestamp?

jdye64:
@ayushdg thanks for taking a look either way. Yeah, the value from [debug print statement] (https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/dask_sql/mappings.py#L294) is `Need to cast from datetime64[ns, Europe/Berlin] to datetime64[ns]` which seems right? Any other thoughts?

ayushdg:
> @ayushdg thanks for taking a look either way. Yeah, the value from [debug print statement](https://github.com/dask-contrib/dask-sql/blob/0b9558d55f56172d3d3617c5dcd351e50433a231/dask_sql/mappings.py#L294) is `Need to cast from datetime64[ns, Europe/Berlin] to datetime64[ns]` which seems right? Any other thoughts?

Yup. I looked into this a bit more. The date type usually gets rid of the time (hours minutes etc) and only keeps the date. In the current logic this is implemented here: https://github.com/dask-contrib/dask-sql/blob/031c04c6e8163f54ad26b40c84e205dcec82cec7/dask_sql/physical/rex/core/call.py#L238. 
In the new codepath I didn't see that codepath getting hit and even after casting we just end up with the timestamp that has the date and time.

ayushdg:
rerun tests

tests/integration/test_filter.py
```diff
@@ -72,12 +68,11 @@ def test_string_filter(c, string_table):
     )
 
 
-@pytest.mark.skip(reason="WIP DataFusion")
 @pytest.mark.parametrize(
     "input_table",
     [
         "datetime_table",
-        pytest.param("gpu_datetime_table", marks=pytest.mark.gpu),
+        # pytest.param("gpu_datetime_table", marks=pytest.mark.gpu),
```
ayushdg, line 75:
```suggestion
        pytest.param("gpu_datetime_table", marks=(pytest.mark.gpu, pytest.mark.skip(reason="WIP DataFusion")),
```

tests/integration/test_filter.py
```diff
@@ -96,7 +91,6 @@ def test_filter_cast_date(c, input_table, request):
     assert_eq(return_df, expected_df)
 
 
-@pytest.mark.skip(reason="WIP DataFusion")
 @pytest.mark.parametrize(
     "input_table",
     [
```
ayushdg, line 96: gpuCI is failing on this test and would probably need to same changes as above.

ayushdg (changes requested):
Minor changes around pytest but otherwise lgtm!

tests/integration/test_filter.py
jdye64: ahh ok will make those as skip for now and figure out what is going on. Thanks for the review

jdye64: Oh cool. Didn't know I could do that. Just made the change and thanks for the suggestion.

ayushdg (approved):

PR #502, opened 2022-04-29
Support CASE WHEN and BETWEEN

jdye64:
Closes: #487


jdye64:
Better to review this after [PR](https://github.com/dask-contrib/dask-sql/pull/498) is merged

ayushdg:
The diff in GitHub haven't updated because of the force push. Would it be possible to merge your branch with the datafusion-sql-planner once again to ensure GitHub recalculates the diff properly?

jdye64:
Yep, sorry force pushed on purpose because the history was getting long and this is just a dev branch.

dask_planner/src/expression.rs
```diff
@@ -281,9 +305,44 @@ impl PyExpr {
                 operands.push(py_ex);
                 Ok(operands)
             }
-            _ => Err(PyErr::new::<pyo3::exceptions::PyTypeError, _>(
-                "unknown Expr type encountered",
-            )),
+            Expr::Case {
```
andygrove, line 308: We should be able to delegate to DataFusion to get the Arrow type for the expression, and then just map that to the dask type.

jdye64:
FYI - I'm waiting on https://github.com/dask-contrib/dask-sql/pull/509 to merge before continuing on and fixing the failing issues here.

dask_planner/src/expression.rs
andygrove: Actually, never mind. I misunderstood what was happening here and thought this was determining the data type for an expression.

jdye64: No worries

dask_sql/physical/rex/core/call.py
```diff
@@ -235,7 +235,7 @@ def cast(self, operand, rex=None) -> SeriesOrScalar:
         # TODO: ideally we don't want to directly access the datetimes,
         # but Pandas can't truncate timezone datetimes and cuDF can't
         # truncate datetimes
-        if output_type == "DATE":
+        if output_type == "DATE" or output_type == "TIMESTAMP":
```
ayushdg, line 238: Is there a specific test that prompted the change? The logic exists only for the `date` type because it truncates all time info and only keeps the date.

dask_sql/physical/rex/core/call.py
```diff
@@ -765,6 +765,18 @@ def date_part(self, what, df: SeriesOrScalar):
             raise NotImplementedError(f"Extraction of {what} is not (yet) implemented.")
 
 
+class BetweenOperation(Operation):
```
ayushdg, line 768: Did calcite automatically convert a between operation into a combination of other ops?

dask_planner/src/expression.rs
```diff
@@ -203,15 +225,44 @@ impl PyExpr {
...
+                };
+
                 Ok(operands)
             }
-            _ => Err(PyErr::new::<pyo3::exceptions::PyTypeError, _>(
-                "unknown Expr type encountered",
-            )),
+            Expr::Between {
+                expr,
+                negated: _,
```
ayushdg, line 254: We should raise a followup issue to handle `not between` with test cases that probably make use of the `negated` attribute

tests/integration/test_select.py
```diff
@@ -126,7 +124,7 @@ def test_timezones(c, datetime_table):
     "input_table",
     [
         "long_table",
-        pytest.param("gpu_long_table", marks=pytest.mark.gpu),
+        # pytest.param("gpu_long_table", marks=pytest.mark.gpu),
```
ayushdg, line 127: is the gpu test failing in this case?

tests/integration/test_select.py
```diff
@@ -196,22 +194,31 @@ def test_timestamp_casting(c, input_table, request):
...
+    expected_df["timezone"] = (
+        expected_df["timezone"].astype("<M8[ns]").dt.floor("D").astype("<M8[ns]")
+    )
+    expected_df["no_timezone"] = (
+        expected_df["no_timezone"].astype("<M8[ns]").dt.floor("D").astype("<M8[ns]")
+    )
+    expected_df["utc_timezone"] = (
+        expected_df["utc_timezone"].astype("<M8[ns]").dt.floor("D").astype("<M8[ns]")
+    )
+
```
ayushdg, line 207: Since we're working with timestamp throughout we shouldn't need to floor to day here

tests/integration/test_select.py
```diff
@@ -196,22 +194,31 @@ def test_timestamp_casting(c, input_table, request):
...
-@pytest.mark.skip(reason="WIP DataFusion")
 def test_multi_case_when(c):
     df = pd.DataFrame({"a": [1, 6, 7, 8, 9]})
     c.create_table("df", df)
 
     actual_df = c.sql(
         """
     SELECT
-        CASE WHEN a BETWEEN 6 AND 8 THEN 1 ELSE 0 END AS C
+        CASE WHEN a BETWEEN 6 AND 8 THEN 1 ELSE 0 END AS "C"
```
ayushdg, line 218: does datafusion not handle this case where the renamed column is not in `"`

ayushdg (changes requested):
Thanks for the changes @jdye64 . I have some minor comments but overall things look good

tests/integration/test_select.py
jdye64: No, and in fact it seems like we were doing this wrong all along and that in order for case to be honored, at least by default, on postgres the name must be in `"`. Otherwise the name is always converted to lowercase in postgres.

tests/integration/test_select.py
jdye64: hmm forgot this was commented out. Let me uncomment and see after the next commit.

dask_sql/physical/rex/core/call.py
jdye64: accidentally left there from some experimentation. Removing it.

dask_sql/physical/rex/core/call.py
jdye64: Yes, previously we were directly receiving the `<>` operator from Calcite but DataFusion just returns a `Between` variant. so that is why this is needed now.

dask_planner/src/expression.rs
jdye64: Done, https://github.com/dask-contrib/dask-sql/issues/519

ayushdg (approved):

ayushdg (commented):
On further testing I discovered that `case when` without the `else` branch don't work because we're looking for an operand in the `else` case that doesn't exist. This didn't happen earlier since calcite automatically added the else operand with a null literal. 
I'll leave it upto you to decide if that change should be handled in this pr or in a followup but either way we should add a couple of test cases to check for this.

dask_planner/src/expression.rs
```diff
@@ -147,53 +175,24 @@ impl PyExpr {
...
-                let mut operands: Vec<PyExpr> = Vec::new();
-                let left_desc: Expr = *left.clone();
-                let py_left: PyExpr = PyExpr::from(left_desc, self.input_plan.clone());
-                operands.push(py_left);
-                let right_desc: Expr = *right.clone();
-                let py_right: PyExpr = PyExpr::from(right_desc, self.input_plan.clone());
-                operands.push(py_right);
-                Ok(operands)
-            }
+            Expr::BinaryExpr { left, op: _, right } => Ok(vec![
```
andygrove, line 192: minor nit: we can just ignore attributes we are not interested in

```suggestion
            Expr::BinaryExpr { left, right, .. } => Ok(vec![
```

dask_planner/src/expression.rs
```diff
@@ -224,13 +252,13 @@ impl PyExpr {
...
-            Expr::Cast {
-                expr: _,
-                data_type: _,
-            } => Ok(String::from("cast")),
-            _ => Err(PyErr::new::<pyo3::exceptions::PyTypeError, _>(
-                "Catch all triggered ....",
-            )),
+            Expr::Cast { .. } => Ok(String::from("cast")),
+            Expr::Between { .. } => Ok(String::from("between")),
+            Expr::Case { .. } => Ok(String::from("case")),
```
andygrove, line 257: minor nit: original code is absolutely fine but it is more typical to just use `.to_string()` or `.to_owned()`.

```suggestion
            Expr::Cast { .. } => Ok("cast".to_string()),
            Expr::Between { .. } => Ok("between".to_string()),
            Expr::Case { .. } => Ok("case".to_string()),
```

andygrove (approved):
LGTM. Left a couple of minor style nits but these can be addressed later.

ayushdg (approved):

PR #504, opened 2022-05-04
Minor code cleanup in row_type()

andygrove:
I was getting up to speed with the `dask-sql` code and took the opportunity to make the `row_type()` method more functional and removed some unnecessary clones and pass fields and schemas by reference instead.

I also removed the `unwrap()` and have it returning a `PyErr` instead, although there are no tests for this but the existing test still pass.

### DIFF
--- a/dask_planner/src/sql/logical.rs
+++ b/dask_planner/src/sql/logical.rs
@@ -12,7 +12,9 @@ pub use datafusion_expr::LogicalPlan;
 
 use datafusion::common::Result;
 use datafusion::prelude::Column;
+use pyo3::ffi::Py_FatalError;
 
+use crate::sql::exceptions::py_type_err;
 use pyo3::prelude::*;
 
 #[pyclass(name = "LogicalPlan", module = "dask_planner", subclass)]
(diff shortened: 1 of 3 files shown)


dask_planner/src/sql/logical.rs
```diff
@@ -140,19 +143,15 @@ impl PyLogicalPlan {
...
-            );
-        }
-        RelDataType::new(false, rel_fields)
+    pub fn row_type(&self) -> PyResult<RelDataType> {
+        let schema = self.original_plan.schema();
+        let mut rel_fields: Vec<RelDataTypeField> = schema
+            .fields()
+            .iter()
+            .map(|f| RelDataTypeField::from(f, schema.as_ref()))
+            .collect::<Result<Vec<_>>>()
```
andygrove, line 152: The iterator is over items of type `Result` so `collect()` would normally return `Vec<Result<_>>` but we can use this turbofish notation to transform that into `Result<Vec<_>>` instead.

dask_planner/src/sql/logical.rs
```diff
@@ -140,19 +143,15 @@ impl PyLogicalPlan {
...
-            rel_fields.push(
-                RelDataTypeField::from(
-                    fields[i].clone(),
-                    self.original_plan.schema().as_ref().clone(),
-                )
-                .unwrap(),
-            );
-        }
-        RelDataType::new(false, rel_fields)
+    pub fn row_type(&self) -> PyResult<RelDataType> {
```
jdye64, line 146: PyO3 will automatically wrap any return type from an annotated function with `PyResult<T>` but maybe it is best to explicitly lay that out. I like that better too.

dask_planner/src/sql/logical.rs
jdye64: much neater code here. Really like the turbofish use.

jdye64 (approved):
LGTM. Thanks for the clean up!

ayushdg (approved):
Thanks for the changes, things look much cleaner.

Issue #499, opened 2022-04-25
[DF] Bump rustc version to 1.60


jdye64:
The rustc version needs to be bumped to 1.60 for CI

PR #508, opened 2022-05-05
Bump rust version

jdye64:
Closes: #499


ayushdg (approved):
Changes lgtm! We should make a follow-up issue to update the gpuci environment as well, which in happy to take on.

PR #509, opened 2022-05-06
Improve code for getting column name from expression

andygrove:
This improves `PyExpr::_column_name` to delegate to DataFusion to build a name for an expression. This code does not appear to be widely used in the tests though so it is hard to tell if this change meets all requirements.

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -63,7 +63,7 @@ impl PyExpr {
         }
     }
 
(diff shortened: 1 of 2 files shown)


jdye64 (changes requested):
Small return type request that was my fault to begin with.

jdye64 (approved):
Very happy to have this change!

charlesbluca (approved):

ayushdg (approved):

PR #514, opened 2022-05-10
Use `from_map` for offset / limit computation

charlesbluca:
Following up on https://github.com/dask/dask/pull/8911, this PR switches over to `from_map` for the computation of limit/offsets instead of `from_delayed`.

cc @rjzamora


dask_sql/physical/rel/logical/limit.py
```diff
@@ -80,17 +79,17 @@ def _apply_offset(self, df: dd.DataFrame, offset: int, end: int) -> dd.DataFrame
         # its pandas representation at this point and we can calculate the cumsum
         # (which is not possible on the dask object). Recalculating it should not cost
         # us much, as we assume the number of partitions is rather small.
-        def select_from_to(df, partition_index, partition_borders):
-            partition_borders = partition_borders.cumsum().to_dict()
+        def select_from_to(index, partition, partition_borders):
+            partition_borders = partition_borders.compute().cumsum().to_dict()
```
charlesbluca, line 83: One difference between `from_map` and `from_delayed` is that Dask objects would already be computed by the time the function is called - is this an intended change?

ayushdg:
Given that the function is performed on all partitions of an existing dataframe instead of creating a new one, this might be a place where `map_partitions` might be a good candidate

rjzamora:
@charlesbluca and I discussed this PR briefly offline, but just to make a note here: The ideal appraoch is probably to use `map_partitions` (like Ayush suggested) with a [`BlockwiseDep`](https://github.com/dask/dask/blob/6685666dd4e3d6542fe2cb2cfbb5a4eb004f76ce/dask/blockwise.py#L27) argument for any arguments that vary between partitions.

charlesbluca:
Closing in favor of #517

PR #515, opened 2022-05-10
Add support for Expr::Sort in expr_to_field

andygrove:
This improves the `expr_to_field` function (used to get column names) to support `Expr::Sort`. It also updates the DataFusion version.

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -471,8 +471,15 @@ impl PyExpr {
 
 /// Create a [DFField] representing an [Expr], given an input [LogicalPlan] to resolve against
 pub fn expr_to_field(expr: &Expr, input_plan: &LogicalPlan) -> Result<DFField> {
-    // TODO this is not the implementation that we really want and will be improved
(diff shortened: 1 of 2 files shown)


dask_planner/src/expression.rs
```diff
@@ -471,8 +471,15 @@ impl PyExpr {
...
-    let fields = exprlist_to_fields(&[expr.clone()], &input_plan.schema())?;
-    Ok(fields[0].clone())
+    match expr {
+        Expr::Sort { expr, .. } => {
+            // DataFusion does not support create_name for sort expressions (since they never
+            // appear in projections) so we just delegate to the contained expression instead
+            expr_to_field(expr, input_plan)
+        }
+        _ => {
+            let fields = exprlist_to_fields(&[expr.clone()], &input_plan)?;
```
andygrove, line 481: Note that with the latest DataFusion code, we now pass a `LogicalPlan` here instead of a schema.

ayushdg (approved):
Thanks for the changes. Did confirm that this works with the approach being used in #505

PR #516, opened 2022-05-10
Reduce crate dependencies

andygrove:
This is a just a minor cleanup of the declared dependencies:

- We do not use `sqlparser` directory so that is removed
- We do not need to depend on `datafusion_expr` since that crate is re-exported in the `datafusion` crate


jdye64 (approved):
100% go

ayushdg (approved):

PR #517, opened 2022-05-11
Use `map_partitions` to compute LIMIT / OFFSET

charlesbluca:
As pointed out on https://github.com/dask-contrib/dask-sql/pull/514, it is possible to compute LIMIT / OFFSETs with `map_partitions` by taking advantage of `BlockIndex` to keep track of each partition's index - this allows us to remove our only uses of `from_delayed` and also seems to reduce the graph size for LIMIT operations pretty significantly; using the graphs from `test_limit` as an example:

Without this PR:

```
HighLevelGraph with 17 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7f9bd0c1d3d0>
 0. from_pandas-7bc6f23913d424ecb44791cc9a3e85e7
 1. blocks-3ef2903a6af6a27edefdfb7c353fc8a2
 2. finalize-da039d13-697f-47c5-af84-5bf412301874
 3. blocks-7a24b6ecfb5ef5a925abe755dbff7fa2
 4. finalize-570d98d5-8c4f-460e-a4ac-d41bc498f257
 5. lambda-51f32144a2d9625e9357b6f2b8eb8f7d
 6. finalize-1b6b2bbd-b2d7-4c28-ba0b-2a3b0d9c2d84
 7. select_from_to-d1af5e72-a001-47e1-ba61-ecd3bba31fb6
 8. finalize-caad46ad-7206-4db0-9a1c-0420f18c66a2
 9. select_from_to-31f6f207-e3c2-44a8-a8d5-78fbd31639c2
 10. finalize-9ffb943a-ecff-4a00-975a-6aeb82bf8c9f
 11. blocks-31465118a7efa7948d10fb6c06af1846
 12. finalize-74bbf2d0-f086-4a61-97ab-05659e7742de
 13. select_from_to-05f5c569-7c79-41a8-b1dd-34d92b43ef51
 14. from-delayed-b27defa85cf4b1aaee609033747f3fc6
 15. getitem-87ce225eb6a27a23eedb6903b6d15a75
 16. rename-c0afa93d468187440fa2b74c9e9d626e
```

With this PR:

```
HighLevelGraph with 6 layers.
<dask.highlevelgraph.HighLevelGraph object at 0x7fcf9e48e160>
 0. from_pandas-7bc6f23913d424ecb44791cc9a3e85e7
 1. lambda-dab2ebc3132687137906110aef7019da
 2. finalize-2aebf1f8-9258-48db-8ce6-c69eb0a10452
 3. limit_partition_func-91d758a891974a9adc19793c9b719376
 4. getitem-b0e73c4215d983aaf703d13f080d0508
 5. rename-c3f7863f4c23ac5b4ce9eb41ff91846b
```


dask_sql/physical/rel/logical/limit.py
```diff
@@ -101,8 +88,8 @@ def select_from_to(df, partition_index, partition_borders):
 
             return df.iloc[from_index:to_index]
 
-        # (b) Now we just need to apply the function on every partition
-        # We do this via the delayed interface, which seems the easiest one.
-        return map_on_partition_index(
-            df, select_from_to, partition_borders, meta=df._meta
+        return df.map_partitions(
+            limit_partition_func,
+            BlockIndex(numblocks=(df.npartitions,)),
```
ayushdg, line 93: iiuc, the `partition_index` attribute associates a partition with a number/value. Is there a reason why `BlockIndex` is preferred in this situation?

ayushdg (approved):
Thanks for this change, small question but overall changes lgtm! 

Side note: This significantly reduces the task graph size for df's with many partitions. 
Previously the number of layers in the task graph grew linearly as num partitions in the input increased.
The task graph with this change always has constant number of layers irrespective of num_partitions.

dask_sql/physical/rel/logical/limit.py
charlesbluca: > iiuc, the partition_index attribute associates a partition with a number/value

Correct, `partition_index` is expected to be the index of the partition within the overall Dask dataframe - the first partition should have `partition_index=0`, the second `partition_index=1`, and so on.

> Is there a reason why `BlockIndex` is preferred in this situation?

My understanding here is the "blocks" of a Dask dataframe are its partitions, making `BlockIndex` roughly synonymous with partition index here.

Chatting with @rjzamora, I now know that another approach to getting partition index would be to add kwarg `partition_info` to the function, which would then be populated with the partition index, among some other info. Perhaps that approach might be a little clearer to future developers?

ayushdg: I don't have a strong preference here and `BlockIndex` getting the same info from dask should be good for the sake of this pr.
For me personally, explicitly mentioning the argument name when calling `map_partitions` should be enough to give an idea of what this is, something like `df.map_partitions(..., partition_index=BlockIndex(numblocks=(df.npartitions,))`.

More generally, the reason for my confusion was primarily because I've seen `Block` like terminology used in the context of IO in dask and `partition` like terminology used in the context of dataframes.

charlesbluca: > something like `df.map_partitions(..., partition_index=BlockIndex(numblocks=(df.npartitions,))`.

Unfortunately, `map_partitions` has different handling for args versus kwargs that makes it so we must supply the `BlockIndex` as an argument in order for it to be computed for the partition functions.

I'll try out `partition_info`, since it makes it somewhat clearer what is being done and seems to be the more documented way of achieving what we want (getting partition info within the function).

rjzamora: Using `partition_info` is fine.  Under the hood, it will just add on a `BlockwiseDep` argument (one that is a bit heavier weight than `BlockIndex`).  I *think* you should feel comfortable going that route here.

If it turns out that you need to use `BlockIndex` directly in the future, you can probably make the code a bit more intuitive by naming the argument a bit more clearly:

```python
partition_indices = BlockIndex(numblocks=(df.npartitions,)
df.map_partitions(..., partition_indices)
```

ayushdg (approved):
Thanks Charles, changes lgtm!

PR #518, opened 2022-05-11
Use `dev` images for independent cluster testing

charlesbluca:
Following up on https://github.com/dask/dask-docker/pull/228, this PR switches our independent cluster testing to use Dask's `dev` images, which represent the latest changes to the dask-docker repo. This should reduce the lag between new features/fixes in the images becoming available to us, while also functioning as downstream testing for dask-docker.

I've also gone ahead and enabled `USE_MAMBA`, which should greatly improve the speed that the containers start up - though I'm still considering if there's a better way of ensuring that the containers are running rather than just calling `sleep`.

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -174,7 +174,7 @@ jobs:
           docker-compose -f .github/docker-compose.yaml up -d
 
           # Wait for installation
-          sleep 40
+          sleep 10
 
           docker logs dask-scheduler
           docker logs dask-worker
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -173,9 +173,6 @@ jobs:
         run: |
           docker-compose -f .github/docker-compose.yaml up -d
 
-          # Wait for installation
-          sleep 10
-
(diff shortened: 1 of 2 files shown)


.github/workflows/test.yml
```diff
@@ -173,9 +173,6 @@ jobs:
         run: |
           docker-compose -f .github/docker-compose.yaml up -d
 
-          # Wait for installation
-          sleep 40
```
ayushdg, line 177: Based on the logs, it looks like the worker image was still installing packages when the logs got printed, but the pytest still succeeded. Maybe we were lucky that the pytests started after that started up?

ayushdg (approved):
Small comment on the sleep, but generally the changes lgtm!

.github/workflows/test.yml
charlesbluca: Yeah that's what I assume to be the case here? I'd like a more reliable way of checking that the cluster has been successfully initialized before jumping to the next step - going to explore pinging the worker/scheduler logs to accomplish this

PR #521, opened 2022-05-12
Datafusion crossjoin

jdye64:


dask_sql/physical/rel/logical/cross_join.py
```diff
@@ -0,0 +1,41 @@
...
+
+    def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContainer:
+        # We now have two inputs (from left and right), so we fetch them both
+        dc_lhs, dc_rhs = self.assert_inputs(rel, 2, context)
+
+        df_lhs = dc_lhs.df
+        df_rhs = dc_rhs.df
+
+        # Create a 'key' column in both DataFrames to join on
+        cross_join_key = "cross_join_key_" + str(time.time_ns())
```
ayushdg, line 35: can probably use `uuid` or the `new_temporary_column` utility from utilities instead of `time.time_ns`

dask_sql/physical/rel/logical/cross_join.py
```diff
@@ -0,0 +1,41 @@
...
+        df_lhs[cross_join_key] = 1
+        df_rhs[cross_join_key] = 1
+
+        result = dd.merge(df_lhs, df_rhs, on=cross_join_key).drop(cross_join_key, 1)
```
ayushdg, line 39: I came across changes in #150 which seemed to change the implementation of cross joins (in join.py), thought I don't completely understand why the updated version has better perf. It might be worth opening an issue to test cross join perf and memory usage and see if the approach in #150 improves it.

ayushdg (changes requested):
Minor suggestions but overall things look good.

tests/integration/test_join.py
```diff
@@ -110,6 +110,24 @@ def test_join_right(c):
     assert_eq(return_df, expected_df, check_index=False)
 
 
+def test_join_cross(c, request):
```
charlesbluca, line 113: Any reason why we're getting the tables here using `request` instead of directly specifying them as fixtures to use? Typically we've only used `request` when we want to parametrize by whether or not we're using GPU, in which case the signature would look something like:

```python
@pytest.mark.parametrize(
    "user_table,department_table",
    [
        ("user_table_1", "department_table"),
        pytest.param("gpu_user_table_1", "gpu_department_table", marks=pytest.mark.gpu),
    ],
)
def test_join_cross(c, user_table, department_table, request):
```

dask_sql/physical/rel/logical/cross_join.py
charlesbluca: Yeah I've wondered why we opted to construct the cross join graph manually rather than depend on `dd.merge` for a while now - would be nice to get some insight on the impact it has performance-wise

tests/integration/test_join.py
```diff
@@ -110,6 +110,22 @@ def test_join_right(c):
     assert_eq(return_df, expected_df, check_index=False)
 
 
+def test_join_cross(c, request, user_table_1, department_table):
```
charlesbluca, line 113: Looks like we can remove `request` as it isn't being used anymore:

```suggestion
def test_join_cross(c, user_table_1, department_table):
```

charlesbluca (commented):
LGTM; just one small thing before this should be good to merge.

We can follow up on the conversation around manual graph construction vs. `dd.merge` once we have some more context around what the performance implications are

charlesbluca (approved):

ayushdg (approved):

PR #522, opened 2022-05-13
Add helper method to convert LogicalPlan to Python type

andygrove:
This PR introduces a helper function to avoid some code repetition.

### DIFF
--- a/dask_planner/src/sql/logical.rs
+++ b/dask_planner/src/sql/logical.rs
@@ -48,56 +48,40 @@ impl PyLogicalPlan {
     }
 }
 
+/// Convert a LogicalPlan to a Python equivalent type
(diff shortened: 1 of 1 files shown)


dask_planner/src/sql/logical.rs
```diff
@@ -48,56 +48,40 @@ impl PyLogicalPlan {
     }
 }
 
+/// Convert a LogicalPlan to a Python equivalent type
+fn to_py_plan<T: From<LogicalPlan>>(current_node: Option<&LogicalPlan>) -> PyResult<T> {
```
andygrove, line 52: This function accepts any generic type `T` that also implements `From<LogicalPlan>`

ayushdg (approved):

Issue #484, opened 2022-04-20
[DF] Port Limit logic to DataFusion from Calcite


jdye64:
**Is your feature request related to a problem? Please describe.**
Update all existing "LIMIT" logic to replace Apache Calcite data structures with Arrow DataFusion data structures.

**Describe the solution you'd like**
Replace instances of Calcite [RelNode](https://calcite.apache.org/javadocAggregate/org/apache/calcite/rel/RelNode.html) with correlating instances of DataFusion [Limit](https://docs.rs/datafusion/latest/datafusion/logical_plan/plan/struct.Limit.html). This includes any base classes, functions, members that might also rely on that logic.

**Describe alternatives you've considered**
None

**Additional context**
Success criteria is for all pytests in test_select.py, or others pytests using limit to be passing

jdye64:
Closed with PR 529

PR #529, opened 2022-05-16
Support for LIMIT clause with DataFusion

jdye64:
Closes: #484


dask_sql/physical/rel/logical/offset.py
```diff
@@ -0,0 +1,113 @@
...
+
+        df = self._apply_offset(df, offset, end)
+
+        print(f"Size of DF: {df.shape[0].compute()} after applying Offset: {offset}")
+
+        cc = self.fix_column_to_row_type(cc, rel.getRowType())
+        # No column type has changed, so no need to cast again
+        return DataContainer(df, cc)
+
+    def _apply_offset(self, df: dd.DataFrame, offset: int, end: int) -> dd.DataFrame:
```
charlesbluca, line 52: Clarifying question - if we do something like `SELECT ... LIMIT 100 OFFSET 50`, would that end up calling the offset plugin and then the limit plugin immediately after? Or is the offset plugin for any `LIMIT` statements with `OFFSETs`, and the limit plugin is just for plain `LIMITs`?

If the former, wondering if we can get away with using something like `tail` to do this relatively simply, since an `OFFSET` alone is essentially just a backwards `LIMIT`. If the latter, I would reccommend incorporating the changes made in https://github.com/dask-contrib/dask-sql/pull/517 here, which should improve things by switching over to `map_partitions` (note that this allows us to drop `map_on_partition_index`).

jdye64: It is the first case. 

First the `offset.py` plugin would be invoked and then the `limit.py` plugin. DataFusion treats offset and limit as different logical operators where Calcite treated them as the "same". Honestly, both could still be in the same file but it seemed more clear to have the logic split apart in separate files to more closely mirror the actual `LogicalPlan` data structure from DataFusion. Could be changed realistically though at the expense of making the logic less clear.

I'm not very familiar with tail but that sounds like something which could be used. Mind posting a code snippet or commit with those changes you are talking about?

dask_planner/src/sql/logical/offset.rs
```diff
@@ -0,0 +1,42 @@
...
+#[derive(Clone)]
+pub struct PyOffset {
+    offset: Offset,
+}
+
+#[pymethods]
+impl PyOffset {
+    #[pyo3(name = "getOffset")]
+    pub fn offset(&self) -> PyResult<PyExpr> {
+        // TODO: Waiting on DataFusion issue: https://github.com/apache/arrow-datafusion/issues/2377
```
ayushdg, line 18: Should be good to remove since this is resolved

dask_sql/physical/rel/logical/limit.py
```diff
@@ -25,82 +22,12 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
-            end = RexConverter.convert(end, df, context=context)
-
-            if offset:
-                end += offset
+        limit = RexConverter.convert(rel, rel.limit().getLimitN(), df, context=context)
 
-        df = self._apply_offset(df, offset, end)
+        # If an offset was present it would have already been processed at this point.
+        # Therefore it is always safe to start at 0 when applying the limit
+        df = df.head(limit, npartitions=-1, compute=False)
```
ayushdg, line 29: My personal opinion here is to still check if the first partition has enough elements and if not call head with `npartitions=-1`.

@charlesbluca Do you think this is worth a broader issue/discussion to see how this can be optimized? 
For example in cases like this
```python
SELECT * from really_large_dataset LIMIT 100
```
is going to read every single partition if `npartitions=-1` before returning 100 rows which isn't ideal.

dask_sql/physical/rel/logical/offset.py
```diff
@@ -0,0 +1,113 @@
...
+        cc = dc.column_container
+
+        offset_node = rel.offset()
+
+        offset = offset_node.getOffset()
+        if offset:
+            offset = RexConverter.convert(rel, offset, df, context=context)
+
+        # TODO: `Fetch` is not currently supported
+        # end = offset_node.getFetch()
```
ayushdg, line 35: Is this the logic implemented in limit, or something else?

dask_sql/physical/rel/logical/offset.py
```diff
@@ -0,0 +1,113 @@
...
+        # if end:
+        #     end = RexConverter.convert(rel, end, df, context=context)
+
+        #     if offset:
+        #         end += offset
+        end = df.shape[0].compute()
```
ayushdg, line 41: This is going to compute the full task graph for df and can be a very expensive operation. Can this be avoided?

dask_sql/physical/rel/logical/offset.py
```diff
@@ -0,0 +1,113 @@
...
+        calculate (!!!) the sizes of each partition.
+
+        After that, we can create a new dataframe from the old
+        dataframe by calculating for each partition if and how much
+        it should be used.
+        We do this via generating our own dask computation graph as
+        we need to pass the partition number to the selection
+        function, which is not possible with normal "map_partitions".
+        """
+        if not offset:
```
ayushdg, line 66: Is it possible to hit this case?

ayushdg (changes requested):

dask_planner/src/sql/logical/limit.rs
```diff
@@ -0,0 +1,32 @@
...
+    #[pyo3(name = "getLimitN")]
+    pub fn limit_n(&self) -> PyResult<PyExpr> {
+        Ok(PyExpr::from(
+            Expr::Literal(ScalarValue::UInt64(Some(self.limit.n.try_into().unwrap()))),
+            Some(self.limit.input.clone()),
+        ))
+    }
+}
+
+impl From<LogicalPlan> for PyLimit {
```
andygrove, line 25: Should we consider using `TryFrom` instead of `From` since this is fallible?

dask_sql/physical/rel/logical/offset.py
jdye64: Dang. I was worried about that. Is there a more efficient way to find the number of rows in the Dask DataFrame? I wasn't aware of any without doing a compute.

dask_planner/src/sql/logical/limit.rs
andygrove: I put up a PR to implement this - https://github.com/dask-contrib/dask-sql/pull/543

dask_sql/physical/rel/logical/offset.py
jdye64: Yes, implemented as part of Limit now. I will remove the comment.

dask_sql/physical/rel/logical/offset.py
jdye64: No, removing it.

andygrove:
Now that https://github.com/apache/arrow-datafusion/pull/2566 is merged, we should probably pick up that change in this PR

jdye64:
Bumped the version and seeing a few test failures now. I will debug those and then make another commit here shortly

andygrove:
> Bumped the version and seeing a few test failures now. I will debug those and then make another commit here shortly

Ok maybe better to follow up in a separate PR in that case?

jdye64:
> > Bumped the version and seeing a few test failures now. I will debug those and then make another commit here shortly
> 
> Ok maybe better to follow up in a separate PR in that case?

Yeah, I think that might be best.

ayushdg (approved):
Changes lgtm! Discussed offline and the logic to improve and optimize offset will be handled in a followup PR.

ayushdg:
@charlesbluca Are these tests only blocked by the independent cluster testing issues w.r.t pyarrow?

charlesbluca:
> @charlesbluca Are these tests only blocked by the independent cluster testing issues w.r.t pyarrow?

Yeah, I think this should be safe to merge as is - I'll also note that there was some issues with mismatched `cloudpickle` versions in our cluster testing that I resolved with https://github.com/dask-contrib/dask-sql/pull/529/commits/c9933777cf94cd88fe578d56aef36082cbf8899f but would like to look into more on the `main` branch

PR #536, opened 2022-05-16
Update upstream testing workflows

charlesbluca:
This PR updates the upstream testing CI by:

- making it so we now use nightly Dask conda packages instead of source install
- adding independent cluster testing to the nightly upstream testing
- making sure we install nightly Dask packages in the independent cluster for upstream testing


.github/cluster-upstream.yml
```diff
@@ -0,0 +1,21 @@
+# Docker-compose setup used during tests
```
jakirkham, line 1: How does this differ from the other docker-compose setup file still here?

.github/workflows/test-upstream.yml
```diff
@@ -69,7 +69,6 @@ jobs:
           miniforge-variant: Mambaforge
           use-mamba: true
           python-version: ${{ matrix.python }}
-          channel-priority: strict
```
jakirkham, line 72: Why is strict priority dropped?

.github/workflows/test.yml
```diff
@@ -148,7 +146,6 @@ jobs:
           miniforge-variant: Mambaforge
           use-mamba: true
           python-version: "3.8"
-          channel-priority: strict
```
jakirkham, line 151: Same question about strict priority

jakirkham (commented):
Thanks Charles! 😄

Had a couple questions below

jakirkham:
Also restarted failing CI jobs

.github/cluster-upstream.yml
charlesbluca: Only difference is that we have the Dask conda nightlies specified in `EXTRA_CONDA_PACKAGES` here - I would've liked to handle this more gracefully with some conditionals within the docker-compose file, but it seems like conditional variable setting isn't really an easy thing to do within compose files.

Open to other options here - I imagine just replacing the files with the equivalent `docker run ...` commands would probably be easier, though that comes at the cost of making local debugging somewhat more difficult.

.github/workflows/test-upstream.yml
charlesbluca: Setting strict channel priority raises an error when we attempt to update the `conda-forge` Dask packages to `dask/label/dev` - [here](https://github.com/dask-contrib/dask-sql/runs/6457470189?check_suite_focus=true) is an example of those failures.

jakirkham: The channels need to be [specified here]( https://github.com/conda-incubator/setup-miniconda#example-4-conda-options ) too

```suggestion
          channel-priority: strict
          channels: dask/label/dev,conda-forge
```

Would then drop the channel from the `mamba install` command. It may also be that command can be dropped if `dask` is in the environment file

charlesbluca: Yeah good point - I think I can probably get away with setting the channels in a conditional for the other workflow as well

.github/workflows/test.yml
```diff
@@ -149,29 +149,35 @@ jobs:
...
           pip list
           mamba list
-      - name: Optionally install upstream dev Dask / dask-ml
+      - name: Optionally install upstream dev dask-ml
         if: needs.detect-ci-trigger.outputs.triggered == 'true'
         run: |
-          python -m pip install --no-deps git+https://github.com/dask/dask
-          python -m pip install --no-deps git+https://github.com/dask/distributed
+          mamba update dask
           python -m pip install --no-deps git+https://github.com/dask/dask-ml
```
jakirkham, line 171: Maybe we should add a `dask-ml` nightly next 🙂

charlesbluca: Yeah was thinking about that 🙂 happy to do the work if Dask maintainers are okay with it, can open up an issue around this

.github/workflows/test.yml
```diff
@@ -149,29 +149,35 @@ jobs:
...
           which python
           pip list
           mamba list
-      - name: Optionally install upstream dev Dask / dask-ml
+      - name: Optionally install upstream dev dask-ml
         if: needs.detect-ci-trigger.outputs.triggered == 'true'
         run: |
-          python -m pip install --no-deps git+https://github.com/dask/dask
-          python -m pip install --no-deps git+https://github.com/dask/distributed
+          mamba update dask
```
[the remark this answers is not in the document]
charlesbluca: Despite including `dask/label/dev` in the `channels` of the Miniconda setup, it seems like the action prioritizes the channels of the environment file when creating an environment, meaning we only consider `conda-forge` packages:

https://github.com/dask-contrib/dask-sql/blob/cb3d9037b9c715e66238f27c68944837813ef7bb/continuous_integration/environment-3.10-jdk11-dev.yaml#L2-L4

The channel still has the highest priority after creating the environment though, so a `mamba update` ends up picking up the nightlies - overall this is pretty similar to what we were doing before by explicitly installing the packages with `mamba install -c dask/label/dev dask`, but allows us to keep the strict channel priority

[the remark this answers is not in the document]
charlesbluca: There's probably some hacky way to insert the `dask/label/dev` channel directly into the Conda environment file before creating the environment, which should allow us to remove this step altogether (if/once dask-ml nightlies are published); I can see if there's a trivial way to do that, otherwise I think this solution seems suitable

.github/workflows/test.yml
jakirkham: Maybe raise an issue on the Dask-ML repo asking this question. Can also link back to issue ( https://github.com/dask/community/issues/76 ) where we had this discussion about Dask & Distributed

.github/workflows/test.yml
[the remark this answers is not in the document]
jakirkham: Agreed this seems fine for now. More was trying to get a handle on what was going on. Could file an issue to follow up later

ayushdg (approved):

.github/cluster-upstream.yml
jacobtomlinson: Docker compose does support env var substitution so perhaps the extra packages could be configured that way?

https://docs.docker.com/compose/environment-variables/

Issue #470, opened 2022-04-12
[DF] Port Union logic to Datafusion


jdye64:
**Is your feature request related to a problem? Please describe.**
Update all existing "Union" logic to replace Apache Calcite data structures with Apache Datafusion data structures.

**Describe the solution you'd like**
Replace instances of Calcite [LogicalUnion](https://calcite.apache.org/javadocAggregate/org/apache/calcite/rel/logical/LogicalUnion.html) with correlating instances of Datafusion [Union](https://docs.rs/datafusion/latest/datafusion/logical_plan/plan/struct.Union.html). This includes any base classes, functions, members that might also rely on that logic.

**Describe alternatives you've considered**
None

**Additional context**
Success criteria is for all pytests in test_union.py to be passing

galipremsagar:
Resolved by #542 & #559

PR #542, opened 2022-05-18
[REVIEW] Add support for `UNION`

galipremsagar:
Fixes: #470

This PR adds `UNION` support in `dask-sql` which uses `datafusion`'s `union` logic.

### DIFF
--- a/dask_planner/src/sql/logical/union.rs
+++ b/dask_planner/src/sql/logical/union.rs
@@ -1,6 +1,5 @@
-
 use datafusion::logical_expr::logical_plan::Union;
-pub use datafusion::logical_expr::{LogicalPlan};
+pub use datafusion::logical_expr::LogicalPlan;
 
 use pyo3::prelude::*;
 
@@ -19,7 +18,6 @@ impl PyUnion {
         println!("{:?}", self.union.alias);
         Ok(false)
     }
-    
 }
 
 impl TryFrom<LogicalPlan> for PyUnion {
--- a/dask_planner/src/sql/logical/union.rs
+++ b/dask_planner/src/sql/logical/union.rs
@@ -1,0 +1,29 @@
+
+use datafusion::logical_expr::logical_plan::Union;
+pub use datafusion::logical_expr::{LogicalPlan};
+
+use pyo3::prelude::*;
+
+#[pyclass(name = "Union", module = "dask_planner", subclass)]
+#[derive(Clone)]
+pub struct PyUnion {
+    union: Union,
+}
+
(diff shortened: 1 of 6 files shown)


andygrove:
Related DataFusion issue: https://github.com/apache/arrow-datafusion/issues/2573

dask_planner/src/sql/logical/union.rs
```diff
@@ -0,0 +1,32 @@
...
+    union: Union,
+}
+
+#[pymethods]
+impl PyUnion {
+    #[pyo3(name = "all")]
+    pub fn all(&mut self) -> PyResult<bool> {
+        println!("{:?}", self.union.inputs[0]);
+        println!("{:?}", self.union.schema.metadata());
+        println!("{:?}", self.union.alias);
```
jdye64, line 18: `println!` statements should be removed. If you want you could implement `__str__` impl like 

```
fn __repr__(&self) -> &'static str {
        format!("{:?}\n{:?}\n{:?}", 
                        self.union.inputs[0],
                        self.union.schema.metadata(),
                        self.union.alias)
}
```

dask_planner/src/sql/logical/union.rs
```diff
@@ -0,0 +1,32 @@
...
+    }
+}
+
+impl TryFrom<LogicalPlan> for PyUnion {
+    type Error = PyErr;
+
+    fn try_from(logical_plan: LogicalPlan) -> Result<Self, Self::Error> {
+        match logical_plan {
+            LogicalPlan::Union(union) => Ok(PyUnion { union }),
+            _ => panic!("something went wrong here"),
```
jdye64, line 29: Can replace this with what the other implementations are using. It isn't that great either really. Need to come up with something more clear error wise later.

```
Err(py_type_err("unexpected plan"))
```

jdye64 (changes requested):
Few small changes. LGTM after that. Thanks Prem!

dask_planner/src/sql/logical/union.rs
galipremsagar: Done 👍

dask_planner/src/sql/logical/union.rs
galipremsagar: Thanks Jeremy, disabled this method for future use along with a TODO.

jdye64 (approved):

ayushdg (commented):
Changes lgtm! We should update #470 to track that union all is still pending on upstream datafusion changes.

PR #543, opened 2022-05-18
Implement TryFrom for plans

andygrove:
This allows us to remove some panics and handle errors more gracefully.

### DIFF
--- a/dask_planner/src/sql/exceptions.rs
+++ b/dask_planner/src/sql/exceptions.rs
@@ -1,9 +1,10 @@
 use datafusion::error::DataFusionError;
 use pyo3::{create_exception, PyErr};
+use std::fmt::Debug;
 
 create_exception!(rust, ParsingException, pyo3::exceptions::PyException);
 
-pub fn py_type_err(e: DataFusionError) -> PyErr {
+pub fn py_type_err(e: impl Debug) -> PyErr {
     PyErr::new::<pyo3::exceptions::PyTypeError, _>(format!("{:?}", e))
 }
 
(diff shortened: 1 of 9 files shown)


dask_planner/src/sql/logical.rs
```diff
@@ -42,18 +42,19 @@ impl PyLogicalPlan {
 
     /// Gets the index of the column from the input schema
     pub(crate) fn get_index(&mut self, col: &Column) -> usize {
-        let proj: projection::PyProjection = self.current_node.clone().unwrap().into();
+        let proj: projection::PyProjection = self.current_node.clone().unwrap().try_into().unwrap();
```
andygrove, line 45: I had to unwrap here because this function does not currently return `PyResult`

dask_planner/src/sql/exceptions.rs
```diff
@@ -1,9 +1,10 @@
 use datafusion::error::DataFusionError;
 use pyo3::{create_exception, PyErr};
+use std::fmt::Debug;
 
 create_exception!(rust, ParsingException, pyo3::exceptions::PyException);
 
-pub fn py_type_err(e: DataFusionError) -> PyErr {
+pub fn py_type_err(e: impl Debug) -> PyErr {
```
andygrove, line 7: I'm not sure if this is the best pattern to use. Sometimes we want to pass a string and sometimes we want to pass a `DataFusionError` and they both implement the `Debug` trait. Perhaps it would be clearer to have separate functions for each?

jdye64 (approved):
LGTM. If you want to change the `py_type_err` function based on the comment you left I'm fine with that but either way I think this is a massive improvement with or without that and worth merging in now so as we open more PRs in the next few days it is already present.

jdye64:
@andygrove has a small merge conflict after another PR was merged but that is all

ayushdg (approved):

Issue #476, opened 2022-04-15
[DF] Enable DataFusion CBO


jdye64:
**Is your feature request related to a problem? Please describe.**
DataFusion offers support for optimizing a LogicalPlan (CBO - Cost Based Optimization). We currently are not utilizing that functionality however. The idea being we get everything working without it to stabilize and then write the plumbing code to enable it after that. This issue serves as a reminder that the CBO still needs to be enabled.

**Describe the solution you'd like**
Make calls to DataFusion API to optimize `LogicalPlan`

**Describe alternatives you've considered**
None

**Additional context**
Just a couple of lines of code for this

jdye64:
Closed with #558

PR #558, opened 2022-05-31
Enable DataFusion CBO and introduce DaskSqlOptimizer

jdye64:
Closes: #476


dask_sql/context.py
```diff
@@ -481,7 +487,7 @@ def sql(
                 for df_name, df in dataframes.items():
                     self.create_table(df_name, df, gpu=gpu)
 
-            rel, select_fields, _ = self._get_ral(sql)
+            rel, select_fields, _ = self._get_ral(sql, config_options)
```
ayushdg, line 490: We can avoid passing this explicitly and use the `dask_config.get` directly within `_get_ral`. 

It has the added advantage of being able to use config options set in a dask yaml or explicitly set in the context that's not explicitly passed in here

tests/integration/test_join.py
```diff
@@ -121,6 +121,9 @@ def test_join_cross(c, user_table_1, department_table):
     assert_eq(return_df, expected_df, check_index=False)
 
 
+@pytest.mark.skip(
```
ayushdg, line 124: While some of the optimization work in ongoing, we can maybe create an issue for these cases where the optimized plan isn't handled where the issue can track the optimized and non optimized version of the plans.

ayushdg (changes requested):
Rust changes lgtm! Minor suggestions on the python side

dask_sql/utils.py
```diff
@@ -94,6 +94,19 @@ def __init__(self, sql, validation_exception_string):
         super().__init__(validation_exception_string.strip())
 
 
+class OptimizationException(Exception):
+    """
+    Helper class for formatting exceptions that occur while trying to
+    optimize a logical plan
+    """
+
+    def __init__(self, sql, exception_string):
```
charlesbluca, line 103: Is there a reason we need to pass on `sql` to this class initialization?

dask_sql/context.py
```diff
@@ -829,17 +835,23 @@ def _get_ral(self, sql):
...
-        # Optimization might remove some alias projects. Make sure to keep them here.
-        select_names = [field for field in rel.getRowType().getFieldList()]
+        # Optimize the `LogicalPlan` or skip if configured
+        if dask_config.get("sql.skip_optimize"):
+            rel = nonOptimizedRel
+        else:
+            try:
+                rel = self.context.optimize_relational_algebra(nonOptimizedRel)
+            except DFOptimizationException as oe:
+                rel = nonOptimizedRel
```
charlesbluca, line 845: Nitpicky, but would it be more direct to make this config option `optimize` and have it do the opposite of what it's doing now?

```suggestion
        if dask_config.get("sql.optimize"):
            try:
                rel = self.context.optimize_relational_algebra(nonOptimizedRel)
            except DFOptimizationException as oe:
                rel = nonOptimizedRel
        else:
            rel = nonOptimizedRel
```

dask_sql/sql-schema.yaml
```diff
@@ -31,3 +31,8 @@ properties:
         type: boolean
         description: |
           Whether to try pushing down filter predicates into IO (when possible).
+
+      skip_optimize:
```
charlesbluca, line 35:
```suggestion
      optimize:
```

dask_sql/sql.yaml
```diff
@@ -7,3 +7,5 @@ sql:
     case_sensitive: True
 
   predicate_pushdown: True
+
+  skip_optimize: True
```
charlesbluca, line 11:
```suggestion
  optimize: False
```

charlesbluca (commented):
Small nitpick around the config option itself:

jdye64:
Confused why all the tests in this PR pass locally but not in CI?? Trying to figure that out.

Locally:
```65 passed, 329 skipped, 1 xfailed, 200 warnings in 27.66s```

charlesbluca (approved):

ayushdg (approved):

PR #559, opened 2022-06-01
[REVIEW] Fix issue with duplicates in column renaming

galipremsagar:
When we have a dataframe `df` with columns: `['a', 'b']` and run a project statement:
```
SELECT a AS "I", a as "II" FROM df
```

there is an AssertionError that occurs here: https://github.com/dask-contrib/dask-sql/blob/11dcf222c317a5327b31a8a4e71eca7c3f61d54e/dask_sql/datacontainer.py#L62

This is the root cause for `test_union_mixed` failures. This PR attempts to fix this issue by introducing a new `rename` method so that other parts of the code-base can be switched to call `rename_handle_duplicates` on case-by-case basis if we start seeing similar issues with other `rename` calls.

### DIFF
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -91,14 +91,16 @@ class ColumnContainer:
         duplicates in `from_columns`
         """
         cc = self._copy()
-        for column_from, column_to in zip(from_columns, to_columns):
-            backend_column = self._frontend_backend_mapping[str(column_from)]
-            cc._frontend_backend_mapping[str(column_to)] = backend_column
+        cc._frontend_backend_mapping.update(
+            {
+                str(column_to): self._frontend_backend_mapping[str(column_from)]
+                for column_from, column_to in zip(from_columns, to_columns)
+            }
+        )
 
         columns = dict(zip(from_columns, to_columns))
         cc._frontend_columns = [
-            str(columns[col]) if col in columns else col
-            for col in self._frontend_columns
+            str(columns.get(col, col)) for col in self._frontend_columns
         ]
 
         return cc
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -83,6 +83,26 @@ class ColumnContainer:
 
         return cc
 
+    def rename_handle_duplicates(
+        self, from_columns: List[str], to_columns: List[str]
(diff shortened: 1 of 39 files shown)


dask_sql/datacontainer.py
```diff
@@ -83,6 +83,26 @@ def rename(self, columns: Dict[str, str]) -> ColumnContainer:
...
+        """
+        Same as `rename` but additionally handles presence of
+        duplicates in `from_columns`
+        """
+        cc = self._copy()
+        for column_from, column_to in zip(from_columns, to_columns):
+            backend_column = self._frontend_backend_mapping[str(column_from)]
+            cc._frontend_backend_mapping[str(column_to)] = backend_column
+
+        columns = dict(zip(from_columns, to_columns))
```
charlesbluca, line 98: Would it make sense to create `columns` at the beginning since we need to iterate through `zip(from_columns, to_columns)` multiple times? Also feel like we could probably simplify the for loop to list comprehension since we can directly set `_frontend_backend_mapping`:

```suggestion
        columns = dict(zip(from_columns, to_columns))

        cc._frontend_backend_mapping.update({
            str(column_to): self._frontend_backend_mapping[str(column_from)]
            for column_from, column_to in columns.items()
        })
```

dask_sql/datacontainer.py
```diff
@@ -83,6 +83,26 @@ def rename(self, columns: Dict[str, str]) -> ColumnContainer:
...
+        cc._frontend_columns = [
+            str(columns[col]) if col in columns else col
```
charlesbluca, line 100:
```suggestion
            str(columns.get(col, col))
```

dask_sql/datacontainer.py
galipremsagar: I have updated the loop to be a list comprehension but moving the `columns` creation before updating `_frontend_backend_mapping` is the actual root cause of the issue. For ex:
```python
>>> from_cols = ['a', 'a']
>>> to_cols = ['b', 'd']
>>> dict(zip(from_cols, to_cols))
{'a': 'd'}
```

We don't want to eliminate `'b'` until we update `_frontend_backend_mapping`, that when `test_union_mixed` case starts failing.

charlesbluca (approved):
LGTM

PR #562, opened 2022-06-02
Add CODEOWNERS file

charlesbluca:
This should hopefully automate the process of requesting review for DF-related work; starting out pretty general with the hopes that we can get more granular with codeowners as time goes on.

Keep in mind that the people/orgs listed here must have write access to the repo


ayushdg (approved):
Thanks for this!

jdye64 (approved):

PR #566, opened 2022-06-06
Upgrade DataFusion version & support non-equijoin join conditions

andygrove:
This PR upgrades to a recent version of DataFusion. There are two changes in DataFusion that had to be addressed here:

- `get_table_provider` now returns a logical `TableSource` instead of a physical `TableProvider`
- Joins now support filter conditions other than equi-join conditions

Changes in this PR:

- Remove `DaskTableProvider`
- Update `join_condition()` to handle all filter conditions

### DIFF
--- a/dask_planner/src/sql/logical/join.rs
+++ b/dask_planner/src/sql/logical/join.rs
@@ -25,13 +25,7 @@ impl PyJoin {
             .join
             .on
             .iter()
-            .map(|(l, r)| {
-                binary_expr(
-                    Expr::Column(l.clone()),
-                    Operator::Eq,
-                    Expr::Column(r.clone()),
-                )
-            })
+            .map(|(l, r)| Expr::Column(l.clone()).eq(Expr::Column(r.clone())))
             .collect();
 
         // other filter conditions
--- a/dask_planner/src/sql/logical/join.rs
+++ b/dask_planner/src/sql/logical/join.rs
@@ -3,7 +3,7 @@ use crate::sql::column;
 
 use datafusion::common::Column;
 use datafusion::logical_expr::{
-    col,
+    and, binary_expr, col,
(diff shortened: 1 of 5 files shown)


dask_planner/src/sql/logical/join.rs
```diff
@@ -20,36 +20,35 @@ pub struct PyJoin {
...
+                    Expr::Column(l.clone()),
+                    Operator::Eq,
+                    Expr::Column(r.clone()),
+                )
+            })
+            .collect();
 
-                root_expr = Expr::BinaryExpr {
-                    left: Box::new(root_expr),
-                    op: Operator::Eq,
```
andygrove, line 41: This looks like it was incorrect before and should have been `Operator::and`

dask_planner/src/sql/logical/join.rs
```diff
@@ -20,36 +20,29 @@ pub struct PyJoin {
...
-            )
-        } else {
-            panic!("Join Length: {}, Encountered a Join with more than a single column for the join condition. This is not currently supported
-            until DataFusion makes some changes to allow for Joining logic other than just Equijoin.", self.join.on.len())
+        // other filter conditions
+        if let Some(filter) = &self.join.filter {
+            filters.push(filter.clone());
         }
+
+        assert!(filters.len() > 0);
```
andygrove, line 36: Is this a safe assumption? Does this method get called if there is a join with no join condition? If so, should it return `Option<PyExpr>`?

andygrove:
@jdye64 Please take a look when you get a chance

dask_planner/src/sql/logical/join.rs
jdye64: yep, good catch

jdye64 (approved):
LGTM

Issue #569, opened 2022-06-06
[BUG] Column name selection gives incorrect result


ayushdg:
**What happened**:
Selecting a specific column from df returns the first column

**What you expected to happen**:
Correct column being returned

**Minimal Complete Verifiable Example**:

```python
import pandas as pd
from dask_sql import Context

df = pd.DataFrame({"a":[1,2,3],"b":[4,5,6]})

c = Context()
c.create_table("df",df)

c.sql("SELECT * from df").compute()

   a  b
0  1  4
1  2  5
2  3  6

c.sql("SELECT b from df").compute()

   b
0  1
1  2
2  3
```

**Anything else we need to know?**:
I think it might have been introduced in #559 but I'm not sure. 
I also wonder why none of the tests caught this. 

**Environment**:

- dask-sql version: datafusion-sql-planner branch with #558 
- Python version: 3.8
- Operating System: ubuntu18.04
- Install method (conda, pip, source): pip

galipremsagar:
On the `datafusion-sql-planner` branch with latest changes, I don't seem to run into this issue:
```python
>>> import pandas as pd
df = pd.DataFrame({"a":[1,2,3],"b":[4,5,6]})>>> from dask_sql import Context
>>> 
>>> df = pd.DataFrame({"a":[1,2,3],"b":[4,5,6]})
>>> c = Context()
>>> c.create_table("df",df)
>>> c.sql("SELECT * from df").compute()
   a  b
0  1  4
1  2  5
2  3  6
>>> c.sql("SELECT b from df").compute()
   b
0  4
1  5
2  6
```

@ayushdg could you check if you have an outdated branch?

randerzander:
@galipremsagar this is w/ #558

ayushdg:
Can confirm that this is with #558 and doesn't show up when setting `sql.optimize:False`

galipremsagar:
Ah, Thanks @randerzander. It would then be easier for @jdye64 to know if some behavior is changing because of https://github.com/dask-contrib/dask-sql/pull/558

jdye64:
Hmm interesting. Will try to take a look at this today. Thanks for pointing it out. Nothing obvious jumps out to me what could be causing it.

jdye64:
#571 resolved this.

PR #571, opened 2022-06-07
Datafusion invalid projection

jdye64:
Closes: #569

### DIFF
--- a/dask_planner/src/sql/logical/table_scan.rs
+++ b/dask_planner/src/sql/logical/table_scan.rs
@@ -1,0 +1,54 @@
+use datafusion::logical_expr::LogicalPlan;
+use datafusion::logical_plan::TableScan;
+
+use crate::sql::exceptions::py_type_err;
+use pyo3::prelude::*;
+
+#[pyclass(name = "TableScan", module = "dask_planner", subclass)]
+#[derive(Clone)]
+pub struct PyTableScan {
+    pub(crate) table_scan: TableScan,
+}
+
+#[pymethods]
+impl PyTableScan {
+    #[pyo3(name = "getTableScanProjects")]
+    fn scan_projects(&mut self) -> PyResult<Vec<String>> {
+        let mut projs: Vec<String> = Vec::new();
+        match &self.table_scan.projection {
+            Some(e) => {
+                for index in e {
+                    projs.push(
+                        (&*self.table_scan.source.schema())
+                            .field(*index)
+                            .name()
+                            .clone(),
+                    );
+                }
+            }
+            None => (),
+        }
+        Ok(projs)
+    }
+
+    /// If the 'TableScan' contains columns that should be projected during the
+    /// read return True, otherwise return False
+    #[pyo3(name = "containsProjections")]
+    fn contains_projections(&self) -> bool {
+        match &self.table_scan.projection {
+            Some(_) => true,
+            None => false,
+        }
+    }
+}
+
+impl TryFrom<LogicalPlan> for PyTableScan {
+    type Error = PyErr;
+
+    fn try_from(logical_plan: LogicalPlan) -> Result<Self, Self::Error> {
+        match logical_plan {
+            LogicalPlan::TableScan(table_scan) => Ok(PyTableScan { table_scan }),
+            _ => Err(py_type_err("unexpected plan")),
+        }
+    }
+}
--- a/dask_planner/src/sql/logical/table_scan.rs
+++ b/dask_planner/src/sql/logical/table_scan.rs
@@ -14,31 +14,23 @@ pub struct PyTableScan {
 impl PyTableScan {
     #[pyo3(name = "getTableScanProjects")]
     fn scan_projects(&mut self) -> PyResult<Vec<String>> {
-        let mut projs: Vec<String> = Vec::new();
         match &self.table_scan.projection {
-            Some(e) => {
-                for index in e {
-                    projs.push(
-                        (&*self.table_scan.source.schema())
-                            .field(*index)
-                            .name()
-                            .clone(),
-                    );
-                }
+            Some(indices) => {
+                let schema = self.table_scan.source.schema();
+                Ok(indices
+                    .iter()
+                    .map(|i| schema.field(*i).name().to_string())
+                    .collect())
             }
-            None => (),
+            None => Ok(vec![]),
         }
-        Ok(projs)
     }
 
     /// If the 'TableScan' contains columns that should be projected during the
     /// read return True, otherwise return False
     #[pyo3(name = "containsProjections")]
     fn contains_projections(&self) -> bool {
-        match &self.table_scan.projection {
-            Some(_) => true,
-            None => false,
-        }
+        self.table_scan.projection.is_some()
     }
 }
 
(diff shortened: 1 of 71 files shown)


jdye64:
#558 should be merged in first. This PR was based off 558

jdye64:
#558 is merged now @charlesbluca so feel free to review this at your leisure

tests/integration/test_select.py
```diff
@@ -222,3 +222,13 @@ def test_case_when_no_else(c):
...
 
     assert_eq(actual_df, expected_df)
+
+
+def test_singular_column_projection_simple(c):
+    df = pd.DataFrame({"a": [1, 2, 3], "b": [4, 5, 6]})
+    c.create_table("df", df)
+
+    wildcard_result = c.sql("SELECT * from df").compute()
+    single_col_result = c.sql("SELECT b from df").compute()
```
charlesbluca, line 232: We shouldn't need to `compute` here, and generally it's better to avoid whenever possible so we get some extra testing on the Dask collection itself:

```suggestion
    wildcard_result = c.sql("SELECT * from df")
    single_col_result = c.sql("SELECT b from df")
```

dask_sql/physical/rel/logical/table_scan.py
```diff
@@ -48,11 +51,16 @@ def convert(
...
-        field_specifications = [str(f) for f in row_type.getFieldNames()]
-        cc = cc.limit_to(field_specifications)
+        # If the 'TableScan' instance contains projected columns only retrieve those columns
+        # otherwise get all projected columns from the 'Projection' instance, which is contained
+        # in the 'RelDataType' instance, aka 'row_type'
+        if table_scan.containsProjections():
+            field_specifications = table_scan.getTableScanProjects()
+        else:
+            row_type = table.getRowType()
+            field_specifications = [str(f) for f in row_type.getFieldNames()]
```
charlesbluca, line 61: Nitpicky, but since we don't end up using `row_type` anywhere else:

```suggestion
            field_specifications = [str(f) for f in table.getRowType().getFieldNames()]
```

dask_sql/physical/rel/logical/table_scan.py
```diff
@@ -48,11 +51,16 @@ def convert(
...
         cc = dc.column_container
 
-        # Make sure we only return the requested columns
-        row_type = table.getRowType()
-        field_specifications = [str(f) for f in row_type.getFieldNames()]
-        cc = cc.limit_to(field_specifications)
+        # If the 'TableScan' instance contains projected columns only retrieve those columns
+        # otherwise get all projected columns from the 'Projection' instance, which is contained
+        # in the 'RelDataType' instance, aka 'row_type'
+        if table_scan.containsProjections():
```
ayushdg, line 57: Not directly related to invalid projections, but we based on this, it needs to be seen if predicate pushdown is working as expected. 
Probably a followup issue to investigate this.

jdye64: I tested locally and this does enable that it is working. I will add a pytest proving that however.

dask_planner/src/sql/logical/table_scan.rs
```diff
@@ -0,0 +1,54 @@
+use datafusion::logical_expr::LogicalPlan;
+use datafusion::logical_plan::TableScan;
+
+use crate::sql::exceptions::py_type_err;
+use pyo3::prelude::*;
+
+#[pyclass(name = "TableScan", module = "dask_planner", subclass)]
+#[derive(Clone)]
+pub struct PyTableScan {
+    pub(crate) table_scan: TableScan,
```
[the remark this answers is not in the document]
jdye64: `pub(crate)` implies "protected" in most other languages. So yes, it needs to be "protected" `pub (crate)` so that the `logical.rs` file can use it

tests/integration/test_select.py
```diff
@@ -222,3 +222,13 @@ def test_case_when_no_else(c):
     expected_df = pd.DataFrame({"C": [None, 1, 1, 1, None]})
 
     assert_eq(actual_df, expected_df)
+
+
+def test_singular_column_projection_simple(c):
```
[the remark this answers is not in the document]
jdye64: Yes, now that #558 is merged `sql.optimize` is always true however. This is why the error was previously only seen on the #558 PR branch.

So with the PR this test is occurring using the optimizer by default as are all other tests as well.

dask_planner/src/sql/logical/table_scan.rs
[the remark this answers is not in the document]
ayushdg: Thanks for that context. In that case other similar logical plan variants can also implement this in their attributed in a followup pr.

dask_sql/physical/rel/logical/table_scan.py
jdye64: @ayushdg just added the test, let me know what you think. I read a locally generated Parquet file with differing combinations of columns and validate just the specified columns are read.

dask_planner/src/sql/logical/table_scan.rs
```diff
@@ -0,0 +1,54 @@
...
+                        (&*self.table_scan.source.schema())
+                            .field(*index)
+                            .name()
+                            .clone(),
+                    );
+                }
+            }
+            None => (),
+        }
+        Ok(projs)
```
andygrove, line 31: We can avoid the mutable state and use functional code to map indices to field names. Also, this version just calls the table `schema()` method once, avoiding some unnecessary clones of Arc pointers, and removes the use of `&*`.

```suggestion
        match &self.table_scan.projection {
            Some(indices) => {
                let schema = self.table_scan.source.schema();
                Ok(indices
                    .iter()
                    .map(|i| schema.field(*i).name().to_string())
                    .collect())
            }
            None => Ok(vec![]),
        }
```

jdye64: love it. Will make the change

tests/integration/test_select.py
```diff
@@ -222,3 +222,34 @@ def test_case_when_no_else(c):
...
+    "input_cols",
+    [
+        ["a"],
+        ["a", "b"],
+        ["a", "d"],
+        ["d", "a"],
+        ["a", "b", "d"],
+    ],
+)
+def test_predicate_pushdown(c, parquet_ddf, input_cols):
```
ayushdg, line 247: Given that the projection step will filter out the columns anyway, it's tough to figure out if column selection is being pushed down to the IO with this test. We might need to do something similar to https://github.com/dask-contrib/dask-sql/pull/433/files to ascertain if the IO step is picking up these filters. 

Though it's non blocking and can be done in a followup pr.

ayushdg (approved):

dask_sql/physical/rel/logical/table_scan.py
```diff
@@ -48,11 +51,15 @@ def convert(
...
+            field_specifications = table_scan.getTableScanProjects()
+        else:
+            field_specifications = [str(f) for f in table.getRowType().getFieldNames()]
```
galipremsagar, line 60:
```suggestion
            field_specifications = table.getRowType().getFieldNames()
```

Do we have to loop over and cast field names with `str`?

galipremsagar (approved):
LGTM, added a non-blocking question

charlesbluca (approved):

PR #576, opened 2022-06-13
Datafusion upstream merge

jdye64:
Merge upstream/main into the datafusion-sql-planner branch after the recent release.


dask_sql/mappings.py
```diff
@@ -283,15 +283,17 @@ def cast_column_to_type(col: dd.Series, expected_type: str):
...
-    current_float = pd.api.types.is_float_dtype(current_type)
-    expected_integer = pd.api.types.is_integer_dtype(expected_type)
-    if current_float and expected_integer:
-        logger.debug("...truncating...")
-        # Currently "trunc" can not be applied to NA (the pandas missing value type),
-        # because NA is a different type. It works with np.NaN though.
-        # For our use case, that does not matter, as the conversion to integer later
-        # will convert both NA and np.NaN to NA.
-        col = da.trunc(col.fillna(value=np.NaN))
+    if pd.api.types.is_integer_dtype(expected_type):
```
ayushdg, line 286: Some of these changes were added to incorporate the `timestampdiff` related operations

dask_sql/physical/rex/core/call.py
```diff
@@ -169,11 +169,10 @@ def div(self, lhs, rhs):
         # We do not need to truncate in this case
         # So far, I did not spot any other occurrence
         # of this function.
-        if isinstance(result, datetime.timedelta):
-            return result
-        else:  # pragma: no cover
-            result = da.trunc(result)
+        if isinstance(result, (datetime.timedelta, np.timedelta64)):
```
ayushdg, line 172: Same

dask_sql/physical/rex/core/call.py
```diff
@@ -818,6 +814,7 @@ class RexCallPlugin(BaseRexPlugin):
         "/int": IntDivisionOperator(),
         # special operations
         "cast": CastOperation(),
+        "reinterpret": CastOperation(),
```
ayushdg, line 817: Same

ayushdg (changes requested):
Given that some of the changes here are ones that were added for timestampdiff. It might make sense to exclude all the changes for now that were made in #495 apart from tests or include everything and remove the parts that are not relevant later.

jdye64:
@ayushdg Does the last commit address what you were talking about?

ayushdg:
> @ayushdg Does the last commit address what you were talking about?

It does but I now realize it's still missing some of the old logic that was updated for timestampdiff, eg places like [this]( https://github.com/dask-contrib/dask-sql/pull/495/files#diff-d3cd58a1b69c1ecccbefe8b23cd8e3c78cfc005841dd6711a094e75f388a289eL290).


To ensure that some of these changes don't slip through, I'm actually considering just including all the changes made to #495 and then removing bit that are not needed in a followup. Apologies if that increases the amount work needed on this pr.

ayushdg (approved):
Some changes from #581 also made it in. But changes lgtm!

PR #579, opened 2022-06-15
Don't move jar to local mvn repo

ksonj:
Since dask-sql does not rely on the mvn repository during runtime but instead adds the jar to the project installation dir and modifies the classpath by reading from pkg_resources it shouldn't be necessary to run `mvn install` here. `mvn package` should suffice.

### DIFF
--- a/setup.py
+++ b/setup.py
@@ -19,7 +19,7 @@ def install_java_libraries(dir):
         raise OSError(
             "Can not find the mvn (maven) binary. Make sure to install maven before building the jar."
         )
-    command = [maven_command, "clean", "install", "-f", "pom.xml"]
+    command = [maven_command, "clean", "package", "-f", "pom.xml"]
     subprocess.check_call(command, cwd=os.path.join(dir, "planner"))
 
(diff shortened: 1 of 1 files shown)


GPUtester:

charlesbluca:
ok to test

ayushdg:
Thanks for this PR @ksonj. 
I'm not too familiar with the maven packaging process but it looks like `package` creates a jar while `install` also installed it in the local repository for other maven packages to pick it up. 

Dask-sql (python) doesn't use the maven repository and just relies on the jar, and I don't think there are other libraries just relying on our dask-sql jar so I think we should be good here. 
Also tagging @jdye64 since you're more familiar with this and might have thoughts.

jdye64 (approved):
This makes sense to me. While some people might want the resulting jar in their local maven repository I agree that the vast majority of Dask-SQL users do not. The actual moving of the jar to the needed location is taken care of at https://github.com/dask-contrib/dask-sql/blob/main/setup.py#L26 anyway. Thanks @ksonj

Issue #463, opened 2022-04-12
[DF] Port Filter logic to Datafusion


jdye64:
**Is your feature request related to a problem? Please describe.**
Update all existing "Filter" logic to replace Apache Calcite data structures with Apache Datafusion data structures.

**Describe the solution you'd like**
Replace instances of Calcite [LogicalFilter](https://calcite.apache.org/javadocAggregate/org/apache/calcite/rel/logical/LogicalFilter.html) with correlating instances of Datafusion [Filter](https://docs.rs/datafusion/latest/datafusion/logical_plan/plan/struct.Filter.html). This includes any base classes, functions, members that might also rely on that logic.

**Describe alternatives you've considered**
None

**Additional context**
Success criteria is for all pytests in test_filter.py to be passing

ChrisJar:
I can take care of this one

jdye64:
Closed with PR #581

PR #581, opened 2022-06-16
Datafusion filter

jdye64:
Closes: #463

### DIFF
--- a/dask_planner/src/sql/logical.rs
+++ b/dask_planner/src/sql/logical.rs
@@ -9,8 +9,9 @@ mod filter;
 mod join;
 mod limit;
 mod offset;
-pub mod projection;
+mod projection;
 mod sort;
+mod table_scan;
 mod union;
 
 use datafusion_common::{Column, DFSchemaRef, DataFusionError, Result};
@@ -54,7 +55,7 @@ fn to_py_plan<T: TryFrom<LogicalPlan, Error = PyErr>>(
     current_node: Option<&LogicalPlan>,
 ) -> PyResult<T> {
     match current_node {
-        Some(plan) => plan.clone().try_into().into(),
+        Some(plan) => plan.clone().try_into(),
         _ => Err(py_type_err("current_node was None")),
     }
 }
@@ -111,6 +112,11 @@ impl PyLogicalPlan {
         to_py_plan(self.current_node.as_ref())
     }
 
+    /// LogicalPlan::TableScan as PyTableScan
+    pub fn table_scan(&self) -> PyResult<table_scan::PyTableScan> {
+        to_py_plan(self.current_node.as_ref())
+    }
+
     /// Gets the "input" for the current LogicalPlan
     pub fn get_inputs(&mut self) -> PyResult<Vec<PyLogicalPlan>> {
         let mut py_inputs: Vec<PyLogicalPlan> = Vec::new();
@@ -136,7 +142,7 @@ impl PyLogicalPlan {
     pub fn get_current_node_schema_name(&self) -> PyResult<&str> {
         match &self.current_node {
             Some(e) => {
-                let sch: &DFSchemaRef = e.schema();
+                let _sch: &DFSchemaRef = e.schema();
                 //TODO: Where can I actually get this in the context of the running query?
                 Ok("root")
             }
@@ -150,7 +156,7 @@ impl PyLogicalPlan {
     #[pyo3(name = "getCurrentNodeTableName")]
     pub fn get_current_node_table_name(&mut self) -> PyResult<String> {
         match self.table() {
-            Ok(dask_table) => Ok(dask_table.name.clone()),
+            Ok(dask_table) => Ok(dask_table.name),
             Err(_e) => Err(PyErr::new::<pyo3::exceptions::PyTypeError, _>(
                 "Unable to determine current node table name",
             )),
@@ -206,7 +212,7 @@ impl PyLogicalPlan {
             .iter()
             .map(|f| RelDataTypeField::from(f, schema.as_ref()))
             .collect::<Result<Vec<_>>>()
-            .map_err(|e| py_type_err(e))?;
+            .map_err(py_type_err)?;
         Ok(RelDataType::new(false, rel_fields))
     }
 }
--- a/dask_planner/src/sql/logical.rs
+++ b/dask_planner/src/sql/logical.rs
@@ -13,11 +13,8 @@ pub mod projection;
 mod sort;
 mod union;
 
-use datafusion::logical_expr::LogicalPlan;
-
-use datafusion::common::{DataFusionError, Result};
-use datafusion::logical_plan::DFSchemaRef;
-use datafusion::prelude::Column;
+use datafusion_common::{Column, DFSchemaRef, DataFusionError, Result};
+use datafusion_expr::LogicalPlan;
 
 use crate::sql::exceptions::py_type_err;
 use pyo3::prelude::*;
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -115,6 +115,15 @@ class ReduceOperation(Operation):
 
     def reduce(self, *operands, **kwargs):
         if len(operands) > 1:
+            if pd.api.types.is_datetime64_dtype(operands[0].dtype):
+                # Doing math against dates requires special types. Introduce those here.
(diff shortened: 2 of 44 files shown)


dask_sql/physical/rex/core/call.py
```diff
@@ -115,6 +115,25 @@ def __init__(self, operation: Callable, unary_operation: Callable = None):
...
+            # in the list are Timedelta for the operation to work.
+            if date_operands:
+                # Operands is a Set, since we are altering it must convert to a List
+                operands = list(operands)
+
+                # Knowing there are datetime types in the operands check for incompatable other types
+                # If found, convert them to Timedelta
+                for idx, operand in enumerate(operands):
+                    # Default to `Day`/`D` since that is what PostgreSQL does
+                    if isinstance(operand, (np.int64, str)):
```
ayushdg, line 134: In which case would we need to cast a string type to a `timedelta64[D]`?

ayushdg (commented):
Generally changes lgtm! 

We need to add more testing around different date time arithmetic and operations. We can probably do it in a followup pr but should open an issue for tracking. 

Especially for cases where sql `interval` types are being added/subtracted from dates/timestamps

dask_sql/physical/rex/core/call.py
jdye64: Case I was addressing was some SQL clauses like `TIMESTAMP '2022-06-17'` for example. Where `TIMESTAMP` is accepting a string literal value and not a column.

ayushdg: In this case doesn't the `TIMESTAMP '2022-06-17` literal get casted to a np.datetime64 object?

jdye64: Good point. It probably should be but I am for sure seeing `str` instance like that show up here. It is only happening with GPU. I bet somewhere the cast operations are not getting full coverage on GPU specific types.

ayushdg: Perfect. Can we open up an issue for tracking that one?

jdye64: Done, https://github.com/dask-contrib/dask-sql/issues/589

dask_sql/physical/rex/core/call.py
```diff
@@ -115,6 +115,25 @@ def __init__(self, operation: Callable, unary_operation: Callable = None):
...
+                        operands[idx] = np.timedelta64(operand, "D")
```
charlesbluca, line 135: Following up on our internal sync:

```suggestion
                    if isinstance(operand, np.int64):
                        operands[idx] = np.timedelta64(operand, "D")
                    elif isinstance(operand, str):
                        operands[idx] = np.datetime64(operand)
```

dask_sql/physical/rex/core/call.py
```diff
@@ -115,6 +115,25 @@ def __init__(self, operation: Callable, unary_operation: Callable = None):
...
+            # If there are datetime64 operands we need to make sure that the other operands
+            # in the list are Timedelta for the operation to work.
+            if date_operands:
+                # Operands is a Set, since we are altering it must convert to a List
+                operands = list(operands)
+
+                # Knowing there are datetime types in the operands check for incompatable other types
+                # If found, convert them to Timedelta
+                for idx, operand in enumerate(operands):
+                    # Default to `Day`/`D` since that is what PostgreSQL does
```
charlesbluca, line 133: Would probably add a `TODO` here bringing up the fact that it would be nice to push this scalar conversion up further so that we can assess if we are dealing with a `DATE` or `TIMESTAMP`.

dask_sql/physical/rex/core/call.py
```diff
@@ -169,11 +188,10 @@ def div(self, lhs, rhs):
         # We do not need to truncate in this case
         # So far, I did not spot any other occurrence
         # of this function.
-        if isinstance(result, datetime.timedelta):
+        if isinstance(result, (datetime.timedelta, np.timedelta64)):
             return result
```
charlesbluca, line 192: Could we add a test to cover this case? I ran through the test suite with a `breakpoint` and nothing hit this case

galipremsagar (dismissed):

dask_planner/src/sql/logical.rs
```diff
@@ -206,14 +208,14 @@ impl PyLogicalPlan {
...
-            .fields()
-            .iter()
-            .map(|f| RelDataTypeField::from(f, schema.as_ref()))
-            .collect::<Result<Vec<_>>>()
-            .map_err(py_type_err)?;
-        Ok(RelDataType::new(false, rel_fields))
+        let mut all_rel_fields: HashSet<RelDataTypeField> = HashSet::new();
+        for schema in self.original_plan.all_schemas() {
+            for field in schema.fields() {
+                all_rel_fields.insert(RelDataTypeField::from(field, schema.as_ref()).unwrap());
```
wence-, line 214: Previously an `Err` rather than an `Ok` would be propagated to the caller via `PyResult`, this instead `panic!`s which is more unfriendly.

wence- (commented):
Some comments/suggestions

dask_planner/src/sql/logical.rs
```diff
@@ -206,14 +208,14 @@ impl PyLogicalPlan {
...
     pub fn row_type(&self) -> PyResult<RelDataType> {
-        let schema = self.original_plan.schema();
-        let rel_fields: Vec<RelDataTypeField> = schema
-            .fields()
-            .iter()
-            .map(|f| RelDataTypeField::from(f, schema.as_ref()))
-            .collect::<Result<Vec<_>>>()
-            .map_err(py_type_err)?;
-        Ok(RelDataType::new(false, rel_fields))
+        let mut all_rel_fields: HashSet<RelDataTypeField> = HashSet::new();
```
jdye64, line 211: This is causing non-deterministic behavior. I assumed, incorrectly, that HashSet honored ordering but it does not. Will change to another datastructure.

wence-: `itertools` has a `unique()` adaptor so maybe
```
use itertools::Itertools; # needs itertools as a cargo dependency
```
and
```rust
    pub fn row_type(&self) -> PyResult<RelDataType> {
        let all_rel_fields = self
            .original_plan
            .all_schemas()
            .into_iter()
            .flat_map(|schema| {
                schema
                    .fields()
                    .iter()
                    .map(|f| RelDataTypeField::from(f, schema.as_ref()))
            })
            .collect::<Result<Vec<_>>>()
            .map_err(py_type_err)?
            .into_iter()
            .unique()
            .collect::<_>();
        Ok(RelDataType::new(false, all_rel_fields))
    }
```

wence-: Although it feels like there must be a way of doing that unwrapping of the `Iterator<Result<T, Err>>` without consing two new `Vec`s

dask_sql/physical/rex/core/call.py
```diff
@@ -115,10 +126,39 @@ def __init__(self, operation: Callable, unary_operation: Callable = None):
...
+        #             # a date type
+        #             if isinstance(operand, np.int64):
+        #                 operands[idx] = np.timedelta64(operand, "D")
+        #             elif isinstance(operand, str):
+        #                 operands[idx] = np.datetime64(operand)
+
+        #     enriched_with_kwargs = lambda kwargs: (
+        #         lambda x, y: self.operation(x, y, **kwargs)
+        #     )
+        #     return reduce(enriched_with_kwargs(kwargs), operands)
```
ayushdg, line 161: Commented code

ayushdg (approved):
Based on the discussion in https://github.com/dask-contrib/dask-sql/pull/581#discussion_r903041162 is it right to assume that this pr only handles datetime objects for now and a followup pr will be opened to handle #584?

jdye64:
> Based on the discussion in https://github.com/dask-contrib/dask-sql/pull/581#discussion_r903041162 is it right to assume that this pr only handles datetime objects for now and a followup pr will be opened to handle #584?

Yes, that is correct

dask_sql/physical/rex/core/call.py
charlesbluca: Do we want to keep this in for now, or can we remove?

```suggestion
```

charlesbluca (approved):

randerzander:
rerun tests

Issue #594, opened 2022-06-23
[DF] Implement `DaskTableSource::supports_filter_pushdown`


andygrove:
**Is your feature request related to a problem? Please describe.**
DataFusion can push predicates down to `TableScan` nodes in the logical plan but only if the underlying `TableSource` supports predicate pushdown. the `DaskTableSource` does not yet implement `supports_filter_pushdown` so no predicates get pushed down currently.

**Describe the solution you'd like**
Implement `DaskTableSource::supports_filter_pushdown`

**Describe alternatives you've considered**
None

**Additional context**
None

PR #595, opened 2022-06-23
Push down null filters into TableScan

andygrove:
Closes https://github.com/dask-contrib/dask-sql/issues/594

Changes in this PR:

- Implement `DaskTableSource::supports_filter_pushdown` so that it accepts simple `IS NOT NULL` filters
- Change default `nullable` to true

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -73,7 +73,7 @@ impl ContextProvider for DaskSQLContext {
                         let mut fields: Vec<Field> = Vec::new();
                         // Iterate through the DaskTable instance and create a Schema instance
                         for (column_name, column_type) in &table.columns {
-                            fields.push(Field::new(column_name, column_type.data_type(), false));
+                            fields.push(Field::new(column_name, column_type.data_type(), true));
                         }
 
                         resp = Some(Schema::new(fields));
--- a/dask_planner/src/sql/table.rs
+++ b/dask_planner/src/sql/table.rs
@@ -38,6 +38,8 @@ impl TableSource for DaskTableSource {
         self.schema.clone()
     }
 
+    // temporarily disable clippy until TODO comment below is addressed
+    #[allow(clippy::if_same_then_else)]
     fn supports_filter_pushdown(
         &self,
         filter: &Expr,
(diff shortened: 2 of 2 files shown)


jdye64:
Ye olde pre-commit got you =)

dask_planner/src/sql/table.rs
```diff
@@ -36,6 +37,36 @@ impl TableSource for DaskTableSource {
...
+    fn supports_filter_pushdown(
+        &self,
+        filter: &Expr,
+    ) -> datafusion_common::Result<TableProviderFilterPushDown> {
+        let mut filters = vec![];
+        split_conjunction(filter, &mut filters);
+        if filters.iter().all(|f| is_supported_push_down_expr(*f)) {
+            // TODO this should return Exact but we cannot make that change until we
+            // are actually pushing the TableScan filters down to the reader because
+            // returning Exact here would remove the Filter from the plan
```
ayushdg, line 52: Is this TODO an action item on the dask-sql side?

dask_planner/src/sql.rs
```diff
@@ -73,7 +73,7 @@ impl ContextProvider for DaskSQLContext {
                         let mut fields: Vec<Field> = Vec::new();
                         // Iterate through the DaskTable instance and create a Schema instance
                         for (column_name, column_type) in &table.columns {
-                            fields.push(Field::new(column_name, column_type.data_type(), false));
+                            fields.push(Field::new(column_name, column_type.data_type(), true));
```
jdye64, line 76: Makes a lot of sense given what Spark SQL does

ayushdg (approved):
Small question for my understanding but changes look good

dask_planner/src/sql/table.rs
andygrove: Yes. With this change, the logical plan will now have filters pushed down to the `TableScan` and those need mapping over to the actual read operation.

dask_planner/src/sql/table.rs
```diff
@@ -36,6 +37,36 @@ impl TableSource for DaskTableSource {
...
+            Ok(TableProviderFilterPushDown::Inexact)
+        } else {
+            Ok(TableProviderFilterPushDown::Unsupported)
+        }
+    }
+}
+
+fn is_supported_push_down_expr(expr: &Expr) -> bool {
+    match expr {
+        // for now, we just attempt to push down simple IS NOT NULL filters on columns
```
jdye64, line 66: Yes, while we can handle more than this I think it makes a lot of sense to handle this low hanging fruit situation. I added issue https://github.com/dask-contrib/dask-sql/issues/596 to cover the remaining situations. That can be handled in a follow up PR

jdye64 (approved):
LGTM

dask_planner/src/sql/table.rs
jdye64: https://github.com/dask-contrib/dask-sql/issues/596 made an issue for this FYI

PR #598, opened 2022-06-24
Consolidate limit/offset logic in partition func

### LINKED ISSUE #597: [DF] [ENH] Update limit/offset logic to reflect upstream changes
With #588, we have once again consolidated the logic for offsets and limits. Right now, we are computing the two separately:

- offsets are computed using a custom `map_partitions` function
- limits are computed from the OFFSET result using `head`

This can probably be consolidated, such that the custom `map_partitions` function also computes limits when necessary, with `head` only being used when no offset is specified.

charlesbluca:
This restores the original consolidated computation of limit/offsets in `limit_partition_func`, while also opting to keep the simplified computation of limits when an offset is not specified, which is what we opted for in #529.

Closes #597

### DIFF
--- a/dask_sql/physical/rel/logical/limit.py
+++ b/dask_sql/physical/rel/logical/limit.py
@@ -30,28 +30,33 @@ class DaskLimitPlugin(BaseRelPlugin):
         limit = RexConverter.convert(rel, rel.limit().getFetch(), df, context=context)
         offset = RexConverter.convert(rel, rel.limit().getSkip(), df, context=context)
 
-        # Apply Offset to DataFrame
-        df = self._apply_offset(df, offset) if offset else df
+        # apply offset to limit if specified
+        if limit and offset:
+            limit += offset
 
-        # Limit DataFrame
-        df = df.head(limit, npartitions=-1, compute=False) if limit else df
-
+        # apply limit and/or offset to DataFrame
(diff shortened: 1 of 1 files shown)


dask_sql/physical/rel/logical/limit.py
```diff
@@ -30,28 +30,33 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
         """
-        Limit the dataframe to the window [offset, end].
+        Limit the dataframe to the window [offset, limit].
 
         Unfortunately, Dask does not currently support row selection through `iloc`, so this must be done using a custom partition function.
         However, it is sometimes possible to compute this window using `head` when an `offset` is not specified.
         """
+        # if no offset is specified we can use `head` to compute the window
+        if not offset:
+            return df.head(limit, npartitions=-1, compute=False)
```
charlesbluca, line 53: @ayushdg not sure if you'd prefer we check the first partition first to see if we can avoid the `npartitions=-1` here?

ayushdg: I think it's fine to keep it as is and revisit if we encounter perf issues.

ayushdg (approved):
Thanks a lot. Changes lgtm!

Issue #556, opened 2022-05-27
[DF] EmpyRelation queries cause errors


jdye64:
**Is your feature request related to a problem? Please describe.**
Any SQL query not performed against a backing table, Ex: `"SELECT 1 AS a, 1.5 AS b, 'x' AS c`, results in the error `NotImplementedError: No relational conversion for node type EmptyRelation available (yet).`

**Describe the solution you'd like**
Empty backing table queries should be supported.

**Describe alternatives you've considered**
None

**Additional context**
Likely the cleanest way to achieve this is introduce this relation as a `no-op` clause in the `rel/convert.py` logic leading to the `project.py` logic to handle it

ChrisJar:
Happy to work on this!

jdye64:
Closed with PR #611

PR #611, opened 2022-07-11
Datafusion empty relation

jdye64:
Closes: #556

### DIFF
--- a/dask_sql/physical/rel/logical/project.py
+++ b/dask_sql/physical/rel/logical/project.py
@@ -1,9 +1,6 @@
 import logging
 from typing import TYPE_CHECKING
 
-import dask.dataframe as dd
-import pandas as pd
-
 from dask_planner.rust import RexType
 from dask_sql.datacontainer import DataContainer
 from dask_sql.physical.rel.base import BaseRelPlugin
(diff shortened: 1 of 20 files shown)


dask_sql/physical/rel/logical/project.py
```diff
@@ -70,6 +73,14 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
         if new_columns:
+            # If the Dask DataFrame is currently empty we must do a little more work and convert the python
+            # literal into a Series so that Dask will assign the columns AND values
+            if len(df.columns) == 0:
+                new_columns = {
+                    k: dd.from_pandas(pd.Series([v]), npartitions=1).clear_divisions()
+                    for (k, v) in new_columns.items()
+                }
+
             df = df.assign(**new_columns)
```
ayushdg, line 84: This `len` check can be expensive in the case of complex queries and should possibly be avoided if possible. It might be worth looking into what calcite was returning for these type of expressions and evaluate if this is needed.

ayushdg (changes requested):

dask_sql/physical/rel/logical/project.py
jdye64: Is the `len` check an almost instantaneous operation on the `columns` since that is local data that doesn't require a compute? At first I was using `df.empty()` which I understand could be expensive. There was a warning from that saying to use this `len(df.columns)` instead.

ayushdg: Ah, I actually missed it. You're right that `df.columns` is essentially a free check. I think this approach might work in that case. 
I just checked the calcite branch and it looks like we always default to returning a dataframe with 1 row if there's no input table as such so we don't run into the same issue as here.

jdye64: What does that resulting "empty" dataframe look like? Dimension wise? Just a dummy column and single value? Something like ...

```
| dummy_col |
0
```

ayushdg (approved):
Thanks for the changes!

dask_sql/physical/rel/logical/project.py
ayushdg: Yeah,
|0|
0

PR #614, opened 2022-07-11
Add argument for rust version in gpuci scripts

### LINKED ISSUE #526: [DF] Update rust version in gpuCI
Based on the discussion in https://github.com/dask-contrib/dask-sql/pull/508#pullrequestreview-964023289 we should add the argument for the rust version in our gpuci scrips.

ChrisJar:
Closes #526

### DIFF
--- a/continuous_integration/gpuci/axis.yaml
+++ b/continuous_integration/gpuci/axis.yaml
@@ -10,4 +10,7 @@ LINUX_VER:
 RAPIDS_VER:
 - "22.08"
 
+RUST_VER:
+- "1.60"
+
 excludes:


charlesbluca:
Yeah the Rust version used in gpuCI is installed in the Dockerfile through this conda env creation command:

https://github.com/rapidsai/dask-build-environment/blob/eec03d28714e523e4066d0649a739adbb5a51121/dask_sql.Dockerfile#L24-L35

So to bump the min version of Rust here we would want to open up a PR bumping `RUST_VER` here:

https://github.com/rapidsai/dask-build-environment/blob/eec03d28714e523e4066d0649a739adbb5a51121/dask_sql.Dockerfile#L11

Alternatively, if we expect to be bumping Rust's min version regularly, we could opt to move the installation of those build dependencies to the gpuCI script itself, so that it can be kept in sync more easily across CPU/GPU testing

jdye64:
I'm fine with just doing the manual sync for now. Once we stabilize our Rust development that version shouldn't be changing but maybe once a year or so.

ChrisJar:
Updated the rust version in the dask_sql dockerfile: https://github.com/rapidsai/dask-build-environment/pull/38

PR #628, opened 2022-07-14
Datafusion version bump

jdye64:
It has been awhile and several improvements have been made on the DataFusion side. Bumping versions.


dask_planner/src/sql/optimizer.rs
```diff
@@ -40,7 +40,7 @@ impl DaskSqlOptimizer {
     ) -> Result<LogicalPlan, DataFusionError> {
         let mut resulting_plan: LogicalPlan = plan;
         for optimization in &self.optimizations {
-            match optimization.optimize(&resulting_plan, &OptimizerConfig::new()) {
+            match optimization.optimize(&resulting_plan, &mut OptimizerConfig::new()) {
                 Ok(optimized_plan) => resulting_plan = optimized_plan,
                 Err(e) => {
                     return Err(e);
```
andygrove, line 46: We'll need to change the code here to be able to skip rules that fail.

```suggestion
                        println!(
                            "Skipping optimizer rule {} due to unexpected error: {}",
                            optimization.name(),
                            e
                        );
```

dask_sql/physical/rel/logical/distinct.py
```diff
@@ -0,0 +1,28 @@
...
+    """
+
+    class_name = "Distinct"
+
+    def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContainer:
+        (dc,) = self.assert_inputs(rel, 1, context)
+        df = dc.df
+        cc = dc.column_container
+
+        df = df.drop_duplicates()
```
andygrove, line 26: The `Distinct` plan was previously represented as an `Aggregate` with grouping expressions and no aggregate expressions. I don't know how this was being mapped to Dask but we would probably want to continue mapping in the same way. I am assuming that the failing tests are because of the change here but I could be wrong.

jdye64: ahhh ok that is good to know. I think I could just map it back to the old way then. Thanks for the info.

dask_planner/src/sql/optimizer.rs
```diff
@@ -40,10 +40,14 @@ impl DaskSqlOptimizer {
     ) -> Result<LogicalPlan, DataFusionError> {
         let mut resulting_plan: LogicalPlan = plan;
         for optimization in &self.optimizations {
-            match optimization.optimize(&resulting_plan, &OptimizerConfig::new()) {
+            match optimization.optimize(&resulting_plan, &mut OptimizerConfig::new()) {
```
ayushdg, line 43: Curious why this needs to be `mut` now

ayushdg (approved):
Minor suggestions. Changes generally lgtm

dask_planner/src/sql/optimizer.rs
andygrove: There is an Id generator in the config strict now, for generating aliases

jdye64:
The CI failure is simply a code cov upload failure. If everyone is good with this PR I think it is good to merge. @ayushdg could you merge if you approve?

charlesbluca (approved):

PR #630, opened 2022-07-15
Conda gpu env

### LINKED ISSUE #627: [ENH] Add dask-cudf to development environment yml
Dask-cudf isn't included in the development environment yml files, causing pytests to fail with `--rungpu` out of the box.

Datafusion branch: https://github.com/dask-contrib/dask-sql/blob/datafusion-sql-planner/continuous_integration/environment-3.10-dev.yaml

jdye64:
Closes #627

### DIFF
--- a/continuous_integration/environment-3.10-gpu-dev.yaml
+++ b/continuous_integration/environment-3.10-gpu-dev.yaml
@@ -1,0 +1,50 @@
+name: dask-sql
+channels:
+  - rapidsai
+  - nvidia
+  - conda-forge
+  - nodefaults
+dependencies:
+  - dask-ml=2022.1.22
+  - dask=2022.3.0
+  - fastapi=0.69.0
+  - intake=0.6.0
+  - jsonschema
+  - lightgbm
+  - maturin=0.12.8
+  - mlflow
+  - mock
+  - nest-asyncio
+  - pandas=1.1.2
+  - pre-commit
+  - prompt_toolkit
+  - psycopg2
+  - pyarrow=3.0.0
+  - pygments
+  - pyhive
+  - pytest-cov
+  - pytest-xdist
+  - pytest
+  - python=3.10
+  - rust=1.60.0
+  - scikit-learn=1.0.0
+  - setuptools-rust=1.1.2
+  - sphinx
+  - tpot
+  - tzlocal=2.1
+  - uvicorn=0.11.3
+  # fugue dependencies; remove when we conda install fugue
+  - adagio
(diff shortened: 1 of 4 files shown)


ayushdg:
cc: @charlesbluca Curious to know if you have opinions on this and if you have a preference for where these files should live and how many we should maintain

continuous_integration/environment-3.10-gpu-dev.yaml
```diff
@@ -0,0 +1,50 @@
...
+  - ciso8601
+  - fs
+  - pip
+  - qpd
+  - triad
+  # GPU dependencies
+  - cudatoolkit=11.5
+  - cudf=22.06
+  - dask-cudf
+  - dask-cuda
```
charlesbluca, line 48: We'll probably want to add the other GPU packages installed in the image to this list:

https://github.com/rapidsai/dask-build-environment/blob/8da1d032d855d487f760749f8bffeb1e532caea9/dask_sql.Dockerfile#L25-L35

charlesbluca (commented):
Thanks for the ping @ayushdg!

My preference here would be to have:

- a single environment file matching up with the gpuCI testing `PYTHON_VER`
- for this file to reside within `continuous_integration/gpuci/`
- potentially for this file to contain only the extra packages installed in [this step](https://github.com/rapidsai/dask-build-environment/blob/8da1d032d855d487f760749f8bffeb1e532caea9/dask_sql.Dockerfile#L24-L35) of the Dockerfile

I'd ideally like this file to make it so we can easily recreate the gpuCI testing environment, which is done by creating the environment created by `environment-{PYTHON_VER}-jdk11-dev.yaml` and then installing the gpuCI requirements on top of this environment. In this case, we'd achieve this with something like:

```
conda env create -f continuous_integration/environment-$PYTHON_VER-jdk11-dev.yaml
conda env update -n dask-sql -f continuous_integration/gpuci/environment.yaml
```

I also think that keeping this file in the gpuCI folder would remove ambiguity on its intended purpose.

As an aside, I think that having the gpuCI requirements specified within the repo is a great way to make the contents of the gpuCI environment more transparent! It might be worthwhile to see if we can find a way to pull these requirements in from the repo during the image builds, which would make it a lot clearer how to bump packages in the gpuCI images (thinking of https://github.com/dask-contrib/dask-sql/pull/614). But would probably like for something like that to be a Dask-wide change, so can hold off on that for now.

jdye64:
This has been handled with another PR

PR #631, opened 2022-07-15
Add helper function for converting expression lists to Python

andygrove:


jdye64 (approved):
Nice addition @andygrove thanks for the add!

PR #632, opened 2022-07-15
Add optimizer rule to simplify expressions

andygrove:
This PR adds two additional optimizer rules. Here is an example of a small improvement that it makes to one of the queries I tested with:

## Before

```
Filter: #date_dim.d_month_seq BETWEEN Int64(1189) AND Int64(1189) + Int64(11) AND #date_dim.d_date_sk IS NOT NULL
```

## After

```
Filter: #date_dim.d_month_seq BETWEEN Int64(1189) AND Int64(1200) AND #date_dim.d_date_sk IS NOT NULL
```


andygrove:
It looks like some parts of this rule are problematic for us, which is a shame. Perhaps we need to make this rule configurable so that we can pick the parts we want. I will look into this.

PR #633, opened 2022-07-16
Add environment file & documentation for GPU tests 

charlesbluca:
A different approach to #630; adds a new environment file to `continuous_integration/gpuci/` specifying the additional packages that get installed on top of the CI environment for gpuCI, along with some documentation on how to use this file to run GPU tests.

Planning to follow up on this with https://github.com/rapidsai/dask-build-environment/pull/39, which would use this file to install requirements into the gpuCI images, which should improve transparency on how the GPU tests work.

### DIFF
--- a/continuous_integration/gpuci/environment.yaml
+++ b/continuous_integration/gpuci/environment.yaml
@@ -1,0 +1,17 @@
+name: gpuci
+channels:
+  - rapidsai
+  - rapidsai-nightly
+  - nvidia
+dependencies:
+  - rust>=1.60.0
+  - setuptools-rust>=1.2.0
+  - cudatoolkit=11.5
+  - cudf=22.08
+  - cuml=22.08
+  - dask-cudf=22.08
+  - dask-cuda=22.08
+  - numpy>=1.20.0
+  - ucx-proc=*=gpu
+  - ucx-py=0.27
+  - xgboost=*=cuda_*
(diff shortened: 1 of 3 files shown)


continuous_integration/gpuci/environment.yaml
```diff
@@ -0,0 +1,17 @@
+name: gpuci
+channels:
+  - rapidsai
+  - rapidsai-nightly
+  - nvidia
+dependencies:
+  - rust>=1.60.0
```
ayushdg, line 7: Should we target these changes for the datafusion branch?

charlesbluca: That would make sense, though on gpuCI's end we would need to grab this file from one of the branches to pull in during image builds ([example of this being done with the environment file](https://github.com/rapidsai/dask-build-environment/blob/8da1d032d855d487f760749f8bffeb1e532caea9/dask_sql.Dockerfile#L14)), and it just seemed to make more sense to target the `main` for this.

If we want to have different versions of this file in `main` and the datafusion branch, that might motivate using separate images for gpuCI running on datafusion PRs vs. PRs into `main`, though not sure if that's worth the extra builds since the images only differ by a handful of added packages

ayushdg: That makes sense w.r.t CI. It might also make sense to include this file (in a separate PR) in the datafusion branch even if we don't use it for gpuCI since one of the motivations for the file was for users working on that branch to be able to setup gpu environments to run tests locally

ayushdg (approved):

PR #636, opened 2022-07-18
Plugins support multiply types

### LINKED ISSUE #635: [DF] Support multiple LogicalPlan types/names per plugin
The current plugin model allows for a single name/type to be used for each plugin. While straightforward this prevents the same plugin from being used for multiple `LogicalPlan` types. For example, `aggregate.py` could be used for both the `LogicalPlan` type of `Aggregate` but also for `Distinct`. It is currently not possible to support that however.

Modify the `utils.py` and existing plugins to provide a `List[str]` of names/types they support. For example change `aggregate.py` line ```class_name = "Aggregate"``` to ```class_name = ["Aggregate", "Distinct"]```. Therefore allowing the plugin to be used for both types.

None

**Additional context**
None

jdye64:
Closes #635

### DIFF
--- a/dask_sql/utils.py
+++ b/dask_sql/utils.py
@@ -67,8 +67,7 @@ class Pluggable:
         if not replace and all(name in Pluggable.__plugins[cls] for name in names):
             return
 
-        for name in names:
-            Pluggable.__plugins[cls][name] = plugin
+        Pluggable.__plugins[cls].update({name: plugin for name in names})
 
     @classmethod
     def get_plugin(cls, name):
--- a/dask_sql/utils.py
+++ b/dask_sql/utils.py
@@ -2,7 +2,7 @@ import importlib
 import logging
 from collections import defaultdict
 from datetime import datetime
-from typing import TYPE_CHECKING, Any, Dict
+from typing import TYPE_CHECKING, Any, Dict, List
 from unittest.mock import patch
 from uuid import uuid4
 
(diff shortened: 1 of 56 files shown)


dask_sql/utils.py
```diff
@@ -62,12 +62,12 @@ class Pluggable:
     __plugins = defaultdict(dict)
 
     @classmethod
-    def add_plugin(cls, name, plugin, replace=True):
+    def add_plugin(cls, names: List[str], plugin, replace=True):
```
charlesbluca, line 65: Wondering if we want to allow both `str` and `List[str]` and just have handling for the string case like

```python
if isinstance(names, str):
    names = [names]
```

That would greatly simplify the diff here

jdye64: Yes, that is a much better approach. Will make the change.

charlesbluca (approved):

dask_sql/utils.py
```diff
@@ -62,12 +62,15 @@ class Pluggable:
...
         """Add a plugin with the given name"""
-        if not replace and name in Pluggable.__plugins[cls]:
+        if isinstance(names, str):
+            names = [names]
+
+        if not replace and all(name in Pluggable.__plugins[cls] for name in names):
             return
 
-        Pluggable.__plugins[cls][name] = plugin
+        Pluggable.__plugins[cls].update({name: plugin for name in names})
```
ayushdg, line 73: I think this logic can also be updated to honor `replace` better for situations where some names are already in the dictionary but not all.

PR #638, opened 2022-07-20
Datafusion expand scalarvalue catchall

jdye64:
We were previously "catching" all encountered `ScalarValue` variants in a single arm and panic-ing. This is the first step in not doing that.

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -403,6 +403,7 @@ impl PyExpr {
                 ScalarValue::Float32(_value) => Ok(String::from("Float32")),
                 ScalarValue::Float64(_value) => Ok(String::from("Float64")),
                 ScalarValue::Decimal128(_value, ..) => Ok(String::from("Decimal128")),
+                ScalarValue::Dictionary(..) => Ok(String::from("Dictionary")),
                 ScalarValue::Int8(_value) => Ok(String::from("Int8")),
                 ScalarValue::Int16(_value) => Ok(String::from("Int16")),
                 ScalarValue::Int32(_value) => Ok(String::from("Int32")),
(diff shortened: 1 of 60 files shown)


dask_sql/physical/rex/core/literal.py
```diff
@@ -145,8 +145,13 @@ def convert(
         elif literal_type == "Null":
             literal_type = SqlTypeName.NULL
             literal_value = None
+        elif literal_type == "IntervalDayTime":
+            literal_type = SqlTypeName.INTERVAL
+            literal_value = rex.getIntervalDayTimeValue()
```
ayushdg, line 150: What kind of literal_value is returned in this case? Is it a string or an integer?

jdye64: i64

ayushdg: In that case we might have to update the logic in sql to python value, since the conversion there seems to assume it's a string and do some computation instead.

jdye64:
Would it be helpful if I made it return a String instead? That is possible with a little wrapper work

dask_sql/physical/rex/core/literal.py
ayushdg: I think int should be fine as long as we update the logic here as well: https://github.com/dask-contrib/dask-sql/blob/datafusion-sql-planner/dask_sql/mappings.py#L131.

I'm surprised all tests passed. Maybe none actually test the case where an interval type is passed?

jdye64:
@ayushdg can we merge this one now?

ayushdg:
> @ayushdg can we merge this one now?

Could we add some tests around interval types or move the interval handling to a followup pr if the other functionality is needed. There's still some issues to iron out there, and statements like `SELECT INTERVAL '3 days' as t` fail.

jdye64:
@ayushdg the failures are unrelated and are part of codecov issues that @charlesbluca is working on I. #666. Does this PR look good to you know with the changes?

dask_planner/src/expression.rs
```diff
@@ -633,6 +637,24 @@ impl PyExpr {
         }
     }
 
+    #[pyo3(name = "getIntervalDayTimeValue")]
```
ayushdg, line 640: We should also add methods to extract this information from interval yearMonth and MonthDayNano types

dask_sql/mappings.py
```diff
@@ -134,6 +140,8 @@ def sql_to_python_value(sql_type: "SqlTypeName", literal_value: Any) -> Any:
 
         return literal_value
 
+    elif sql_type == SqlTypeName.INTERVAL_DAY:
+        return timedelta(days=literal_value[0], milliseconds=literal_value[1])
     elif sql_type == SqlTypeName.INTERVAL:
```
ayushdg, line 145: Same as above

ayushdg (approved):
Changes lgtm! The other interval types can be added in a followup pr.

PR #639, opened 2022-07-21
Add max version constraint for `fugue`

charlesbluca:
It looks like the recent release of Fugue 0.7.0 has bumped its `qpd` dependency to a version that only has python support up to 3.8. I'm not sure if this is the cause for the recent Fugue-related failures, but it does mean that at least for now, we should constrain to `fugue<0.7.0`, where 3.9+ support is guaranteed.

In the long run, we should probably see what the blockers are to allowing 3.9+ support on `qpd` again, cc @goodwanghan in case you have any additional context to provide here.

### DIFF
--- a/continuous_integration/environment-3.10-jdk11-dev.yaml
+++ b/continuous_integration/environment-3.10-jdk11-dev.yaml
@@ -33,7 +33,7 @@ dependencies:
 - uvicorn>=0.11.3
 # fugue dependencies; remove when we conda install fugue
 - adagio
-- antlr4-python3-runtime
+- antlr4-python3-runtime<4.10
 - ciso8601
 - fs
 - pip
--- a/continuous_integration/environment-3.10-jdk11-dev.yaml
+++ b/continuous_integration/environment-3.10-jdk11-dev.yaml
@@ -33,7 +33,7 @@ dependencies:
 - uvicorn>=0.11.3
 # fugue dependencies; remove when we conda install fugue
 - adagio
-- antlr4-python3-runtime<4.10
+- antlr4-python3-runtime
 - ciso8601
 - fs
 - pip
--- a/continuous_integration/environment-3.10-jdk11-dev.yaml
+++ b/continuous_integration/environment-3.10-jdk11-dev.yaml
@@ -40,4 +40,5 @@ dependencies:
 - qpd
 - triad
 - pip:
-  - fugue[sql]>=0.5.3
+  # TODO: remove max version constraint when fugue adds back python 3.9+ support
+  - fugue[sql]>=0.5.3,<0.7.0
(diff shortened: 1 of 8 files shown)


goodwanghan:
Hi, I apologize for the inconvenience, the qpd constraint should not exist. How about I make the change on the QPD side so that you don't need this PR?

charlesbluca:
@goodwanghan that would be great! I'm not 100% sure if that's the cause of the failures we've been seeing in CI, but resolving that would certainly narrow things down a lot 🙂

charlesbluca:
It looks like the failures here persist even with the pip package of `qpd=0.3.1` (which doesn't have the `python<3.9` constraint) 😕 so think for now we might want to go ahead and merge this in to unblock tests, and follow up later with a resolution for whatever the underlying issue is.

ayushdg:
> It looks like the failures here persist even with the pip package of `qpd=0.3.1` (which doesn't have the `python<3.9` constraint) 😕 so think for now we might want to go ahead and merge this in to unblock tests, and follow up later with a resolution for whatever the underlying issue is.

I see so the issue isn't necessarily qpd and python 3.9+ related but still shows up only on python 3.9 and above right?

goodwanghan:
I think the problem is on qpd, 

```python
 classifiers=[
        # "3 - Alpha", "4 - Beta" or "5 - Production/Stable"
        "Development Status :: 3 - Alpha",
        "Intended Audience :: Developers",
        "Topic :: Software Development :: Libraries :: Python Modules",
        "License :: OSI Approved :: Apache Software License",
        "Programming Language :: Python :: 3.6",
        "Programming Language :: Python :: 3.7",
        "Programming Language :: Python :: 3.8",
        "Programming Language :: Python :: 3 :: Only",
    ],
```

I will go ahead adding the versions and publish 0.3.2 to unblock you guys

goodwanghan:
I just published qpd 0.3.2 can you bump up the version to see if that works for you?

continuous_integration/environment-3.10-jdk11-dev.yaml
```diff
@@ -40,4 +40,5 @@ dependencies:
 - qpd
 - triad
 - pip:
-  - fugue[sql]>=0.5.3
+  # TODO: remove max version constraint when fugue adds back python 3.9+ support
```
ayushdg, line 43: Thanks for confirming. In that case can we reword this to indicate that it's not 3.9+ related

ayushdg (approved):
Minor suggestion but lgtm!

continuous_integration/environment-3.10-jdk11-dev.yaml
charlesbluca: Yup! Sorry, forgot to bump this message once I realized the failures were unrelated

Issue #617, opened 2022-07-11
[ENH] Automatically register dask-sql as Fugue's default Dask SQL engine when it's installed


goodwanghan:
**Is your feature request related to a problem? Please describe.**
Currently there are manual steps in the code to start using dask-sql with Fugue.

**Describe the solution you'd like**
I want to automate this, as long as dask-sql is installed, it should replace qpd to be the default dask engine for Fugue.

This now can be done easily, I will go ahead make the code change in your repo to make the auto registration.

charlesbluca:
Is there anything else to be done on the Fugue end of things, or can this be closed with #640

PR #640, opened 2022-07-22
Set Dask-sql as the default Fugue Dask engine when installed

goodwanghan:
This is to address https://github.com/dask-contrib/dask-sql/issues/617.

Unit tests have all passed and the coverage on the change is 100%

A notebook has been added to show how the integration works automatically.


GPUtester:

goodwanghan:
I have also removed the hacks around Fugue and reduced the dependencies on Fugue in the testing environments.

goodwanghan:
@charlesbluca @ayushdg please take a look

continuous_integration/environment-3.10-jdk11-dev.yaml
```diff
@@ -31,14 +31,5 @@ dependencies:
...
-- antlr4-python3-runtime<4.10
-- ciso8601
-- fs
-- pip
-- qpd
-- triad
 - pip:
-  # FIXME: tests are failing with fugue 0.7.0
-  - fugue[sql]>=0.5.3,<0.7.0
+  - fugue>=0.7.0
```
charlesbluca, line 35: A couple questions/comments here:

- I notice we're just pip installing `fugue` without the `sql` extra - is this now included in the core package?
- Is there any reason we need to bump the minimum version of Fugue here? Not sure if there's some changes between 0.5.3 and 0.7.0 that can't be rectified with compatibility code
- If we _do_ opt to bump our minimum Fugue version to 0.7.0, we can start using the conda package of Fugue instead of the pip package, and remove the `pip` section of the dev environments

continuous_integration/environment-3.8-jdk11-dev.yaml
```diff
@@ -31,13 +31,5 @@ dependencies:
...
-- adagio
-- antlr4-python3-runtime<4.10
-- ciso8601
-- fs
-- pip
-- qpd
-- triad
 - pip:
-  - fugue[sql]==0.5.3
+  - fugue>=0.7.0
```
charlesbluca, line 35: The python 3.8 testing environments stay pinned to our minimum supported version to make sure it still works as expected - so depending on what we want our minimum Fugue dependency to be, this would need to be changed to:

```suggestion
  - fugue==0.7.0
```

dask_sql/integrations/fugue.py
```diff
@@ -15,9 +17,24 @@
 
 from dask_sql.context import Context
 
-register_execution_engine(
-    "dask", lambda conf: DaskSQLExecutionEngine(conf), on_dup="overwrite"
-)
+
+@run_at_def
+def register_engines() -> None:
```
charlesbluca, line 22: Should we preface this function with an underscore to emphasize that it's not intended for external use?

```suggestion
def _register_engines() -> None:
```

setup.py
```diff
@@ -112,14 +112,16 @@ def build(self):
...
+        "fugue": ["fugue>=0.7.0"],
     },
     entry_points={
         "console_scripts": [
             "dask-sql-server = dask_sql.server.app:main",
             "dask-sql = dask_sql.cmd:main",
-        ]
+        ],
+        "fugue.plugins": [
+            "dasksql = dask_sql.integrations.fugue:register_engines[fugue]"
```
charlesbluca, line 123: Following up on my last suggestion:

```suggestion
            "dasksql = dask_sql.integrations.fugue:_register_engines[fugue]"
```

charlesbluca (commented):
Thanks for resolving the test failures @goodwanghan 🎉 a couple of comments, mostly around the changes to our Fugue dependency:

notebooks/FugueSQL.ipynb
```diff
@@ -0,0 +1,510 @@
...
+      "text/plain": [
+       "<IPython.core.display.HTML object>"
+      ]
+     },
+     "metadata": {},
+     "output_type": "display_data"
+    }
+   ],
+   "source": [
+    "%%fsql dask\n",
```
charlesbluca, line 86: This cell magic needs to be registered manually to work correctly, right? If so, should we add a block like

```python
from fugue_notebook import setup

setup()
```

Preceding this and explain what's happening there? At least on my end, this notebook doesn't work without running that block.

continuous_integration/environment-3.10-jdk11-dev.yaml
goodwanghan: 1. Yes in >=0.7.0 sql is built in, we don't need any extra
2. There are a lot of changes, but the most important thing is dependency. Because of the breaking change of antlr and duckdb, before 0.7.0 there has to be version caps, now in 0.7.0 we removed all version caps.
3. I feel conda is heavy, adding a package can make it more confused, but pip is light, also it doesn't expose other dependencies to the sensitive conda system. But if you prefer, we can change it to conda dependency.

continuous_integration/environment-3.8-jdk11-dev.yaml
goodwanghan: Sorry, I need to make the same change to 3.8

dask_sql/integrations/fugue.py
goodwanghan: Sure, will change

setup.py
goodwanghan: Will do

notebooks/FugueSQL.ipynb
goodwanghan: This is another big change in 0.7.0, where we start deprecating `fugue_notebook`, instead, it's separated into [fugue-jupyter](https://pypi.org/project/fugue-jupyter/), I will update the notebook. Great catch, thanks!

charlesbluca (approved):

tests/integration/test_fugue.py
```diff
@@ -32,11 +33,20 @@ def test_simple_statement():
...
     assert_eq(return_df, pd.DataFrame({"a": [1], "b": ["world"]}))
 
+    result = fugue_sql.fsql(
+        "SELECT * FROM df WHERE a > 0 YIELD DATAFRAME AS result",
+        df=pdf,
+    ).run(client)
+
+    return_df = result["result"].as_pandas()
+    assert_eq(return_df, pd.DataFrame({"a": [1], "b": ["world"]}))
+    client.close()
```
charlesbluca, line 43: Just noticed, we shouldn't need to close the client here manually as the fixture should do clean up on its own:

```suggestion
    client.close()
```

Issue #606, opened 2022-06-29
[BUG] Registering an invalid UDF causes all subsequent `sql` calls to fail


charlesbluca:
**What happened**:
When registering an invalid UDF (i.e. one with an unsupported return type), all subsequent calls to `Context.sql` fail when attempting to coerce this unsupported Python type into a SQL type.

**What you expected to happen**:
I would expect that an invalid UDF would raise an error at registration time, or at least get discarded if it's been found to be invalid after registration.

**Minimal Complete Verifiable Example**:

```python
import numpy as np
from dask_sql import Context

c = Context()

def f(x):
    return x

c.register_function(f, "f", [("x", np.datetime64)], np.timedelta64)

c.sql("SELECT 1 + 2")
```

**Environment**:

- dask-sql version: 2022.6.0
- Python version: 3.9
- Operating System: ubuntu 20.04
- Install method (conda, pip, source): conda

PR #641, opened 2022-07-22
Validate UDF metadata

brandon-b-miller:
Closes https://github.com/dask-contrib/dask-sql/issues/606

### DIFF
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -4,6 +4,8 @@ from typing import Any, Dict, List, Tuple, Union
 import dask.dataframe as dd
 import pandas as pd
 
+from dask_sql.mappings import python_to_sql_type
+
 ColumnType = Union[str, int]
 
 FunctionDescription = namedtuple(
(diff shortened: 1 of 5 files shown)


dask_sql/datacontainer.py
```diff
@@ -198,11 +200,10 @@ def __init__(self, func, row_udf: bool, params, return_type=None):
...
         self.names = [param[0] for param in params]
 
-        if return_type is None:
-            # These UDFs go through apply and without providing
-            # a return type, dask will attempt to guess it, and
-            # dask might be wrong.
-            raise ValueError("Return type must be provided")
+        # validate UDF metadata
+        for dt in (*(param[1] for param in params), return_type):
+            _ = python_to_sql_type(dt)
```
charlesbluca, line 205: Happy to push this off to a follow up PR if it would require a lot of time to implement, but since we are computing the SQL types of the params / return Python types here, it could be worthwhile to cache this information now and grab it when we later call `python_to_sql_type` for UDF preparation:

https://github.com/dask-contrib/dask-sql/blob/8a7330948a7a8603c4e25297c2038f1f7541aad2/dask_sql/context.py#L751
https://github.com/dask-contrib/dask-sql/blob/8a7330948a7a8603c4e25297c2038f1f7541aad2/dask_sql/context.py#L775

charlesbluca (approved):
Thanks for the work here @brandon-b-miller! Overall LGTM, just a question around another additional improvement that could be made, don't intend for it to be blocking

dask_sql/datacontainer.py
charlesbluca: I played around with this a bit and ended up getting something working in https://github.com/brandon-b-miller/dask-sql/pull/1 🙂

brandon-b-miller: Thanks @charlesbluca ! will take a look at this later this morning

PR #643, opened 2022-07-25
Introduce Inverse Rex Operation

jdye64:
Adds an Rex operation `InverseOperation` to handle Expr(s) from DataFusion's `Negative` Expr variant. This is slightly different from the already existing `NotOperation` and can be applied to values other than boolean.

### DIFF
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -298,8 +298,8 @@ class IsTrueOperation(Operation):
         return not pd.isna(df) and df is not None and not np.isnan(df) and bool(df)
 
 
-class InverseOperation(Operation):
-    """The inverse operator"""
+class NegativeOperation(Operation):
+    """The negative operator"""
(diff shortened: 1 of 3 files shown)


dask_sql/physical/rex/core/call.py
```diff
@@ -867,6 +880,7 @@ class RexCallPlugin(BaseRexPlugin):
         "not like": NotOperation().of(LikeOperation()),
         "like": LikeOperation(),
         "similar to": SimilarOperation(),
+        "negative": InverseOperation(),
```
charlesbluca, line 883: For clarity, should we rename this to `NegativeOperation`? Inverse is a somewhat vague term which could indicate a different type of inverse from the _additive_ inverse we're computing here.

jdye64: yep, that is a good point

charlesbluca (approved):

PR #646, opened 2022-07-26
Bump setuptools-rust from 1.1.1 -> 1.4.1

jdye64:
Bumping setuptools-rust version to take advantage of some new Rust PyO3 enhancements we can use during our Rust work

### DIFF
--- a/continuous_integration/environment-3.10-dev.yaml
+++ b/continuous_integration/environment-3.10-dev.yaml
@@ -26,7 +26,7 @@ dependencies:
 - python=3.10
 - rust>=1.60.0
 - scikit-learn>=1.0.0
-- setuptools-rust>=1.1.2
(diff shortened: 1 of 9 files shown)


continuous_integration/environment-3.10-dev.yaml
```diff
@@ -13,7 +13,7 @@ dependencies:
 - mlflow
 - mock
 - nest-asyncio
-- pandas>=1.1.2
+- pandas>=1.4.0
```
charlesbluca, line 16: Just leaving a note that we'll need to bump our Pandas version to get around a `setuptools` version ceiling added in https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/204 that would otherwise cause conflicts

charlesbluca (approved):

PR #650, opened 2022-07-27
Add test queries to gpuCI checks

charlesbluca:


continuous_integration/gpuci/build.sh
```diff
@@ -37,6 +37,17 @@ gpuci_logger "Activate conda env"
...
+gpuci_logger "Install awscli"
+gpuci_mamba_retry install -y -c conda-forge awscli
+
+gpuci_logger "Download TPC-DS dataset"
+gpuci_retry aws s3 cp --only-show-errors s3://rapidsai-data/tpcx-bb-data/tpc-ds/sf1/parquet_2gb/ tests/unit/data/ --recursive
+
+gpuci_logger "Pull in query files"
+git clone https://github.com/databricks/spark-sql-perf.git
+mkdir -p tests/unit/queries
+cp spark-sql-perf/src/main/resources/tpcds_2_4/* tests/unit/queries
```
charlesbluca, line 49: Just pointing out that this would probably make more sense to do during the image builds versus on a per-run basis; once this is in can refactor to make that the case

continuous_integration/gpuci/build.sh
```diff
@@ -55,4 +66,4 @@ conda config --show-sources
 conda list --show-channel-urls
 
 gpuci_logger "Python py.test for dask-sql"
-py.test $WORKSPACE -n 4 -v -m gpu --rungpu --cov-config="$WORKSPACE/.coveragerc" --cov=dask_sql --cov-report=xml:"$WORKSPACE/dask-sql-coverage.xml" --cov-report term
+py.test $WORKSPACE -n 4 -v -m gpu --runqueries --rungpu --junitxml="$WORKSPACE/junit-dask-sql.xml" --cov-config="$WORKSPACE/.coveragerc" --cov=dask_sql --cov-report=xml:"$WORKSPACE/dask-sql-coverage.xml" --cov-report term
```
ayushdg, line 69: Is there a way to make the query tests another test in our suite that's optional rather than bundled with the regular tests?
If not that's okay and we can probably skip/ignore failing queries for now.

tests/unit/test_queries.py
```diff
@@ -0,0 +1,56 @@
...
+
+
+@pytest.fixture(scope="module")
+def gpu_c():
+    pytest.importorskip("dask_cudf")
+
+    # Lazy import, otherwise the pytest framework has problems
+    from dask_sql.context import Context
+
+    c = Context()
```
ayushdg, line 34: Even though it's single gpu, I'm inclined to try using a local cluster/cudacluster to test this with the distributed scheduler. Though if that's tougher to incorporate, this is a good starting point.

ayushdg (approved):
Minor comments, changes generally lgtm! Thanks

continuous_integration/gpuci/build.sh
charlesbluca: Yeah, two options I could think of are:

1. moving the test file to a different directory from `tests/` like `queries/`, so that calling `pytest tests` would skip over it in the search entirely
2. not include the file in the repo, instead pulling in from somewhere else during the gpuCI run 

My personal preference would be 1 (though it's not perfect, running just `pytest` would still catch the queries) since we don't have a clear place to publish the testing files outside of this repo

continuous_integration/gpuci/build.sh
```diff
@@ -37,6 +37,15 @@ gpuci_logger "Activate conda env"
 . /opt/conda/etc/profile.d/conda.sh
 conda activate dask_sql
 
+gpuci_logger "Install awscli"
+gpuci_mamba_retry install -y -c conda-forge awscli
+
+gpuci_logger "Download parquet dataset"
+gpuci_retry aws s3 cp --only-show-errors s3://rapidsai-data/tpcx-bb-data/tpc-ds/sf1/parquet_2gb/ tests/unit/data/ --recursive
```
charlesbluca, line 44:
```suggestion
gpuci_retry aws s3 cp --only-show-errors "${dask-sql-bucket-name}parquet_2gb/" tests/unit/data/ --recursive
```

continuous_integration/gpuci/build.sh
```diff
@@ -37,6 +37,15 @@ gpuci_logger "Activate conda env"
...
+
+gpuci_logger "Download query files"
+gpuci_retry aws s3 cp --only-show-errors s3://rapidsai-data/tpcx-bb-data/tpc-ds/sf1/queries/ tests/unit/queries/ --recursive
```
charlesbluca, line 47:
```suggestion
gpuci_retry aws s3 cp --only-show-errors "${dask-sql-bucket-name}queries/" tests/unit/queries/ --recursive
```

PR #654, opened 2022-08-01
Merge `main` into `datafusion-sql-planner`

charlesbluca:


dask_sql/physical/rex/core/call.py
```diff
@@ -876,6 +876,7 @@ class RexCallPlugin(BaseRexPlugin):
         "/int": IntDivisionOperator(),
         # special operations
         "cast": CastOperation(),
+        "reinterpret": CastOperation(),
```
ayushdg, line 879: we might be good to remove this since it's calcite specific

ayushdg (approved):
Minor comment but changes lgtm

Issue #269, opened 2021-10-20
[ENH] Consider not persisting by default when using create table as select syntax. 


ayushdg:
**Is your feature request related to a problem? Please describe.**
Based on discussions in #218 dask-sql no longer persists by default when creating tables from dataframes using the `context.create_table` syntax. However when using the create_table syntax from sql the behavior can be slightly different. 
```sql
CREATE TABLE my_table WITH (...)
```
will not persist the table by default unless the param `persist` is passed in the query. 

On the other hand
```sql
CREATE TABLE my_table AS ( SELECT * from table_2)
```
does persist `my_table`.

**Describe the solution you'd like**
Given that the persist behavior is slightly different for similar sounding operations `CREATE_TABLE` in different situations can be confusing and it might make sense to standardize on this. 

**Describe alternatives you've considered**
A workaround to the issue above is using the `CREATE VIEW as SELECT` syntax which does not persist the table. Mentioned in the dask-sql docs [here](https://dask-sql.readthedocs.io/en/latest/pages/sql/creation.html?highlight=create%20table%20as#create-view-as).

**Additional context**
Since the `CTAS` syntax does not support custom arguments with the `WITH` keyword similar to `CREATE TABLE WITH` I'm not sure how easy it would be to implement this feature in the sql grammar. 

here's the snippet that actually sets this to true for persisting: https://github.com/dask-contrib/dask-sql/blob/2010dbd6471457dd50da31385b762f0ebbc3c91f/planner/src/main/codegen/includes/create.ftl#L27-L28

jdye64:
happy to do this one

jdye64:
@ayushdg could this possible be a candidate for your latest dask configuration work?

ayushdg:
In some sense yes. 
There's two aspects to this work:

1. Updating this logic on the Java side to accept a config option and potentially also allowing users to set this config within sql itself using the set syntax as discussed here: `SET dask.sql.{schema_name}.{table_name}.persist = true;`

2. Allowing users to pass this config from the python side. 

The latest config updates will make the second aspect work better and allow users to provide the behavior on a per query basis. I'll need to think a bit more about how it can be done on a per table basis.

PR #656, opened 2022-08-03
[DF] Add support for `CREATE TABLE | VIEW AS` statements

charlesbluca:
Adds support for `CREATE TABLE AS` and `CREATE VIEW AS` statements. Some follow up stuff to look into:

- parsing the `schema_name` through DataFusion instead of hard-coding
- parsing whether or not to persist created tables instead of hard-coding
- consolidating the shared logic between `CreateTableAsPlugin.convert` and `Context.sql` in one utility function

### DIFF
--- a/dask_planner/src/sql/logical/create_memory_table.rs
+++ b/dask_planner/src/sql/logical/create_memory_table.rs
@@ -43,7 +43,7 @@ impl PyCreateMemoryTable {
     pub fn get_if_not_exists(&self) -> PyResult<bool> {
         match &self.create_memory_table {
             Some(create_memory_table) => Ok(create_memory_table.if_not_exists),
-            None => Ok(false),
+            None => Ok(false), // TODO: in the future we may want to set this based on dialect
         }
     }
 
--- a/dask_planner/src/sql/logical/create_memory_table.rs
+++ b/dask_planner/src/sql/logical/create_memory_table.rs
@@ -64,6 +64,14 @@ impl PyCreateMemoryTable {
             },
         }
     }
+
+    #[pyo3(name = "isTable")]
+    pub fn is_table(&self) -> PyResult<bool> {
+        match &self.create_memory_table {
+            Some(_) => Ok(true),
+            None => Ok(false),
+        }
+    }
 }
 
 impl TryFrom<LogicalPlan> for PyCreateMemoryTable {
--- a/dask_planner/src/sql/logical/create_memory_table.rs
+++ b/dask_planner/src/sql/logical/create_memory_table.rs
@@ -14,63 +14,69 @@ pub struct PyCreateMemoryTable {
 impl PyCreateMemoryTable {
     #[pyo3(name = "getName")]
     pub fn get_name(&self) -> PyResult<String> {
-        match &self.create_memory_table {
-            Some(create_memory_table) => Ok(create_memory_table.name.clone()),
+        Ok(match &self.create_memory_table {
+            Some(create_memory_table) => create_memory_table.name.clone(),
             None => match &self.create_view {
-                Some(create_view) => Ok(create_view.name.clone()),
-                None => Err(py_type_err(
-                    "Encountered a non CreateMemoryTable/CreateView type in get_input",
-                )),
+                Some(create_view) => create_view.name.clone(),
+                None => {
+                    return Err(py_type_err(
+                        "Encountered a non CreateMemoryTable/CreateView type in get_input",
+                    ))
+                }
             },
-        }
+        })
     }
 
     #[pyo3(name = "getInput")]
     pub fn get_input(&self) -> PyResult<PyLogicalPlan> {
-        match &self.create_memory_table {
-            Some(create_memory_table) => Ok(PyLogicalPlan {
+        Ok(match &self.create_memory_table {
+            Some(create_memory_table) => PyLogicalPlan {
                 original_plan: (*create_memory_table.input).clone(),
                 current_node: None,
-            }),
+            },
             None => match &self.create_view {
-                Some(create_view) => Ok(PyLogicalPlan {
+                Some(create_view) => PyLogicalPlan {
                     original_plan: (*create_view.input).clone(),
                     current_node: None,
-                }),
-                None => Err(py_type_err(
-                    "Encountered a non CreateMemoryTable/CreateView type in get_input",
-                )),
+                },
+                None => {
+                    return Err(py_type_err(
+                        "Encountered a non CreateMemoryTable/CreateView type in get_input",
+                    ))
+                }
             },
-        }
+        })
     }
 
     #[pyo3(name = "getIfNotExists")]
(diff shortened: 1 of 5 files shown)


dask_sql/physical/rel/custom/create_table_as.py
```diff
@@ -32,29 +34,52 @@ class CreateTableAsPlugin(BaseRelPlugin):
...
+            cc = dc.column_container
+            cc = cc.rename(
+                {
+                    df_col: select_name
+                    for df_col, select_name in zip(cc.columns, select_names)
+                }
+            )
+            dc = DataContainer(dc.df, cc)
+
+        df = dc.assign()
```
charlesbluca, line 83: This is lifted from `Context.sql` to compute the table from the `LogicalPlan`, since (to my knowledge) there isn't an obvious way to get the generating SQL for both `CreateMemoryTable` and `CreateView` (though the latter has a [`definition`](https://docs.rs/datafusion/latest/datafusion/logical_plan/struct.CreateView.html#structfield.definition) attribute that could be useful here).

Going to explore moving this shared code to a utility function to cut down on duplication

dask_sql/physical/rel/custom/create_table_as.py
```diff
@@ -32,29 +34,52 @@ class CreateTableAsPlugin(BaseRelPlugin):
...
+            if create_memory_table.getIfNotExists():
                 return
-            elif not sql.getReplace():
+            elif not create_memory_table.getOrReplace():
                 raise RuntimeError(
                     f"A table with the name {table_name} is already present."
                 )
 
-        sql_select = sql.getSelect()
-        persist = bool(sql.isPersist())
```
charlesbluca, line 51: I'm interested in how this would've been called in a standard `c.sql(...)` query? The only place throughout the tests where I've seen `PERSIST` used is with Fugue, but I assumed this was being handled on Fugue's end.

dask_planner/src/sql/logical/create_memory_table.rs
```diff
@@ -0,0 +1,78 @@
...
+            },
+            current_node: None,
+        })
+    }
+
+    #[pyo3(name = "getIfNotExists")]
+    pub fn get_if_not_exists(&self) -> PyResult<bool> {
+        match &self.create_memory_table {
+            Some(create_memory_table) => Ok(create_memory_table.if_not_exists),
+            None => Ok(false),
```
charlesbluca, line 46: Noting that right now I opted to hard-code `CREATE VIEW AS` statements to raise an error if the table already exists; chatting with @jdye64 it seems like this could be dialect-specific, and so I will add a TODO here to make that clear

dask_sql/physical/rel/custom/create_table_as.py
ayushdg: This information actually comes from the java/calcite side of things. 
https://github.com/dask-contrib/dask-sql/commit/8abc48ce324af7fea0c5fb363a26fb7a70f1a252#diff-258722e0cdeaf803a15b6933e9575641e2d83264e0d8c4cd5cb2ecdd000e74b5R113 
If you see the snippet here, things were set up so that `persist` is true if the user uses the `CREATE TABLE as` syntax but set to false if using the `create_view` syntax.

dask_sql/context.py
```diff
@@ -845,10 +813,44 @@ def _get_ral(self, sql):
         logger.debug(f"_get_ral -> LogicalPlan: {rel}")
         logger.debug(f"Extracted relational algebra:\n {rel_string}")
 
+        return rel, rel_string
+
+    def _compute_table_from_rel(self, rel: "LogicalPlan", return_futures: bool = False):
```
ayushdg, line 818: Should the default for `return_futures` here be `True` in line with c.sql?

dask_sql/physical/rel/custom/create_table_as.py
ayushdg: We can follow the same pattern here if preferred or leave it as is and make it configurable in a followup pr based on some discussion here: https://github.com/dask-contrib/dask-sql/issues/269

ayushdg (changes requested):
Minor suggestions but generally things lgtm!

dask_planner/src/sql/logical/create_memory_table.rs
```diff
@@ -0,0 +1,78 @@
...
+}
+
+#[pymethods]
+impl PyCreateMemoryTable {
+    #[pyo3(name = "getName")]
+    pub fn get_name(&self) -> PyResult<String> {
+        match &self.create_memory_table {
+            Some(create_memory_table) => Ok(format!("{}", create_memory_table.name)),
+            None => match &self.create_view {
+                Some(create_view) => Ok(format!("{}", create_view.name)),
```
jdye64, line 20: `format!("{}",` is not needed, can just use `Ok(create_vire.name)`

dask_planner/src/sql/logical/create_memory_table.rs
```diff
@@ -0,0 +1,78 @@
...
+    create_memory_table: Option<CreateMemoryTable>,
+    create_view: Option<CreateView>,
+}
+
+#[pymethods]
+impl PyCreateMemoryTable {
+    #[pyo3(name = "getName")]
+    pub fn get_name(&self) -> PyResult<String> {
+        match &self.create_memory_table {
+            Some(create_memory_table) => Ok(format!("{}", create_memory_table.name)),
```
jdye64, line 18: `format!("{}",` is not needed, can just use `Ok(create_vire.name)`

dask_sql/context.py
jdye64: Yeah, should probably set this to `True`

jdye64 (changes requested):
Few nits but overall looks good. Thanks!

dask_planner/src/sql/logical/create_memory_table.rs
```diff
@@ -0,0 +1,78 @@
...
+            None => match &self.create_view {
+                Some(create_view) => Ok(format!("{}", create_view.name)),
+                None => panic!("Encountered a non CreateMemoryTable/CreateView type in get_name"),
```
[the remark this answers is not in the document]
charlesbluca: Good point - might also be worth it to do a follow up PR applying this change to some of the other plan types, which also panic in similar situations

dask_sql/physical/rel/custom/create_table_as.py
charlesbluca: I think for now I will follow the original pattern, but add a TODO to indicate that there is ongoing discussion around the behavior.

dask_planner/src/sql/logical/create_memory_table.rs
charlesbluca: Cool, will open up a follow up to make this consistent across the Rust code as I took this pattern from another file

ayushdg (approved):
Thanks for this! Changes lgtm!

jdye64 (approved):
LGTM

Issue #621, opened 2022-07-12
[DF] Add `COT` math function


jdye64:
**Is your feature request related to a problem? Please describe.**
Cotangent function is not supported out of the box with DataFusion. Register a `COT` math function and provide all bindings.

**Describe the solution you'd like**
Register a function in `DaskSQLContext` for `COT` that calculates cotangent

**Describe alternatives you've considered**
None

**Additional context**
None

sarahyurick:
Can be closed, thanks!

ayushdg:
closed via #657

PR #657, opened 2022-08-03
`COT` function

sarahyurick:
Looking at Issue #621 ... the changes so far allow for basic COT functionality with code like:

```
from dask_sql import Context

client = Client()
c = Context()

c.sql("""
    SELECT COT(42) AS "the answer"
""", return_futures=False)
```

Mostly just wanting to check that there aren't any obvious issues with how I'm implementing this? Maybe want `da.tan()` instead of numpy?

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -108,6 +108,11 @@ impl ContextProvider for DaskSQLContext {
             let rtf: ReturnTypeFunction = Arc::new(|_| Ok(Arc::new(DataType::Int64)));
             return Some(Arc::new(ScalarUDF::new("year", &sig, &rtf, &fun)));
         }
+        if "cot".eq(name) {
+            let sig = Signature::variadic(vec![DataType::Float64], Volatility::Immutable);
+            let rtf: ReturnTypeFunction = Arc::new(|_| Ok(Arc::new(DataType::Float64)));
+            return Some(Arc::new(ScalarUDF::new("cot", &sig, &rtf, &fun)));
+        }
 
         // Loop through all of the user defined functions
         for (_schema_name, schema) in &self.schemas {
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -603,14 +603,6 @@ class YearOperation(Operation):
         df = convert_to_datetime(df)
         return df.year
 
-class CotOperation(Operation):
-    def __init__(self):
-        super().__init__(self.do_cot)
-
-    def do_cot(self, df: SeriesOrScalar):
-        df = 1 / np.tan(df)
-        return df
-
 
 class CeilFloorOperation(PredicateBasedOperation):
     """
@@ -951,7 +943,6 @@ class RexCallPlugin(BaseRexPlugin):
         # Temporary UDF functions that need to be moved after this POC
         "datepart": DatePartOperation(),
         "year": YearOperation(),
-        "cot": CotOperation(),
     }
 
     def convert(
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -603,6 +603,14 @@ class YearOperation(Operation):
         df = convert_to_datetime(df)
         return df.year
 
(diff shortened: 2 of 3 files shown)


GPUtester:

Admins can comment `ok to test` to allow this one PR to run or `add to allowlist` to allow all future PRs from the same author to run.

charlesbluca:
add to allowlist

sarahyurick:
Also looks like there are already tests for the trig functions in `tests/integration/test_rex.py` and `tests/integration/test_postgres.py` that are currently being skipped over. What's the protocol for adding unit tests in this case?

jdye64 (approved):
LGTM. Thanks for the PR!

One thing to note, I started this nasty pattern of placing these types of functions here in `sql.rs` and they will be moved to out DaskDialect implementation once we complete that. For now they are fine here.

sarahyurick:
Ok thanks @jdye64 , makes sense! I'll update to use `da.tan()` instead of `np.tan()` and fix the style issue. Should I make a specific test for `COT` somewhere or just leave as is, since some tests already exist but are currently being skipped over?

dask_planner/src/sql.rs
```diff
@@ -108,6 +108,11 @@ impl ContextProvider for DaskSQLContext {
             let rtf: ReturnTypeFunction = Arc::new(|_| Ok(Arc::new(DataType::Int64)));
             return Some(Arc::new(ScalarUDF::new("year", &sig, &rtf, &fun)));
         }
+        if "cot".eq(name) {
```
ayushdg, line 111: Question mainly for @jdye64. Do you have a preference on adding some functions to the dialect in rust vs registering it as a udf on the python side by creating a DaskFunction?

dask_sql/physical/rex/core/call.py
```diff
@@ -943,6 +951,7 @@ class RexCallPlugin(BaseRexPlugin):
         # Temporary UDF functions that need to be moved after this POC
         "datepart": DatePartOperation(),
         "year": YearOperation(),
+        "cot": CotOperation(),
```
ayushdg, line 954: The existing implementation already has cot: https://github.com/dask-contrib/dask-sql/blob/f55c27664e5a06e9405cb1372a19ca60e0c53eae/dask_sql/physical/rex/core/call.py#L914

We might be able to get by without this and just having the parser recognize this function

ayushdg (changes requested):

dask_sql/physical/rex/core/call.py
sarahyurick: You're right - it still works without the additions to `call.py`, so I'll remove them now

dask_planner/src/sql.rs
jdye64: I would prefer to have certain functions, like `cot`, which are quite common and in most DB engines be in the Rust dialect. However, once the Dialect is completed you will be able to interact with it through Python and add new `udfs` of course.

charlesbluca (commented):
Thanks for the work here @sarahyurick 😄 to answer your question around testing, would it make sense to unskip the `COT` testing functions and comment out the queries / column assignments testing the unimplemented functions?

tests/integration/test_rex.py
```diff
@@ -334,55 +333,56 @@ def test_math_operations(c, df):
             , ACOS(b) AS "acos"
             , ASIN(b) AS "asin"
             , ATAN(b) AS "atan"
-            , ATAN2(a, b) AS "atan2"
-            , CBRT(b) AS "cbrt"
```
charlesbluca, line 338: We should be able to comment these out using [SQL comments](https://www.w3schools.com/Sql/sql_comments.asp), so with something like:

```sql
            -- , ATAN2(a, b) AS "atan2"
            -- , CBRT(b) AS "cbrt"
```

sarahyurick: Good to know, thanks! I'll update it right now.

charlesbluca (approved):
LGTM! Looks like the failures here are related to CodeCov upload failures

ayushdg (approved):
lgtm!

PR #658, opened 2022-08-03
[DF] Add support for `DROP TABLE` statements

charlesbluca:
Depends on #656


jdye64:
Gonna wait on #656 before reviewing, if I forget poke me again =)

jdye64 (approved):
LGTM. thanks for the change!

Issue #510, opened 2022-05-09
[DF] Replace `panic!` instances with PyErr(s)


jdye64:
**Is your feature request related to a problem? Please describe.**
There are several places in the existing Rust codebase that `panic!` are hardcoded. We should replace all of those instances with `PyErr` instances that are properly propagated to Python for exception handling and reporting to the end user.

**Describe the solution you'd like**
Have all `panic!` instances properly generate `PyErr` instances 

**Describe alternatives you've considered**
None

**Additional context**
None

PR #662, opened 2022-08-04
Use `PyErrs` for all Python-facing methods in `dask_planner`

charlesbluca:
As a follow up on reviews I received in #656, this PR replaces the `panic` calls within Python-facing methods in `dask_planner` and replaces them with `Err(PyErr))`, making modifications to signatures and logic where necessary to get this working.

This PR also adds/modifies the custom exceptions so that they can be used throughout the codebase instead of `PyErr::new`.

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -126,10 +126,10 @@ impl PyExpr {
     /// Column in the SQL parse tree or not
     #[pyo3(name = "isInputReference")]
     pub fn is_input_reference(&self) -> PyResult<bool> {
-        match &self.expr {
-            Expr::Column(_col) => Ok(true),
-            _ => Ok(false),
-        }
+        Ok(match &self.expr {
+            Expr::Column(_col) => true,
+            _ => false,
+        })
     }
 
     #[pyo3(name = "toString")]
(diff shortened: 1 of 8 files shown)


jdye64:
This would also resolve #510

dask_planner/src/expression.rs
```diff
@@ -213,36 +215,36 @@ impl PyExpr {
...
-            Expr::Wildcard => panic!("Wildcard!!!"),
-            Expr::InSubquery { .. } => "Subquery",
-            Expr::ScalarUDF { .. } => "ScalarUDF",
-            Expr::Exists { .. } => "Exists",
-            Expr::ScalarSubquery(..) => "ScalarSubquery",
-            Expr::QualifiedWildcard { .. } => "Wildcard",
-            Expr::GroupingSet(..) => "GroupingSet",
-        })
+    pub fn get_expr_type(&self) -> PyResult<String> {
+        match &self.expr {
```
jdye64, line 219: Style wise would it make more since to move back to the `String::from(...)` at the match pattern level so individual `Ok(...)` are not needed and then have a single arm for all the "Err" Expr types that were previously `panic(..)`. Like this ....

```
Expr::ScalarVariable(..) 
| Expr::Not(..)
| Expr::IsNotNull(..)
| Expr::Negative(..)
| Expr::GetIndexedField {..}
| Expr::IsNull(..)
| Expr::Case{..}
| Expr::TryCast{..}
| Expr::WindowFunction{..}
| Expr::AggregateUDF{..}
| Expr::Wildcard => Err(py_type_err("Encountered unsupported expression type: {:?}", &self.expr))

jdye64 (changes requested):
This has been long due. Thanks for taking it on

dask_planner/src/expression.rs
charlesbluca: If we opt to do this, is there any reason you prefer specifying all the variants explicitly rather than using a catch all?

jdye64: Technically of course either way works. With lower level languages like Rust though the style usually opts for making things more verbose so the behavior is more clear and anyone can look at the code and clearly see the Expr variants that trigger the condition. With the default "catch all" its not immediately clear without consulting the documentation. Also if say a new Expr variant is added I think you would want the build to fail and not just default on its behavior and manually listing each variant achieves that. Honestly I think its just opinion based and either is fine.

charlesbluca: That logic is sound to me! Happy to stick to enumerating all variants, would you be opposed to doing this for some of the other matches in the expressions implementation where we currently use catch alls?

jdye64 (approved):
Long overdue and much appreciated!

jdye64:
I'm going to merge this since the failures are simply codecov failing to upload. Thanks again @charlesbluca

charlesbluca:
Thanks @jdye64! Yeah for context I'm currently working on #666 to make it so Codecov uploads won't block CI

PR #666, opened 2022-08-08
Update CodeCov upload step in CI

charlesbluca:
A couple changes to our CodeCov upload step in CI:

- Bumps the action to the latest stable version
- Removes the `token` param, which shouldn't be necessary for public repos (and doesn't save us from unexpected 404 errors)
- Removes the non-default setting for `fail_ci_if_error` so that if are unable to upload the report for some reason it won't register tests as failing

### DIFF
--- a/.github/workflows/test.yml
+++ b/.github/workflows/test.yml
@@ -126,10 +126,7 @@ jobs:
           path: junit/test-results.xml
       - name: Upload coverage to Codecov
         if: github.repository == 'dask-contrib/dask-sql'
-        uses: codecov/codecov-action@v2
-        with:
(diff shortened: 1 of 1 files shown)


jdye64:
Looking at the CI failures they all seem one off. @charlesbluca , the changes look fine, CI is failing all be it erroneously. You fine with me merging this?

charlesbluca:
I resolved the failures with #668, so maybe once that is in we can merge upstream here and verify that things are working as expected?

jdye64:
@charlesbluca you wanna merge this with upstream now that #668 or me?

PR #667, opened 2022-08-08
[DF] use datafusion 9956f80f197550051db7debae15d5c706afc22a3

andygrove:
Upgrade to the latest version of DataFusion and Arrow.

Changes:

- `DataTypes::Decimal` was renamed to `Decimal128` and `Decimal256` was added, and these are now both supported here

### DIFF
--- a/dask_planner/src/dialect.rs
+++ b/dask_planner/src/dialect.rs
@@ -1,6 +1,6 @@
 use core::iter::Peekable;
 use core::str::Chars;
-use sqlparser::dialect::Dialect;
+use datafusion_sql::sqlparser::dialect::Dialect;
 
 #[derive(Debug)]
 pub struct DaskSqlDialect {}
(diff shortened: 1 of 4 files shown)


dask_planner/src/dialect.rs
```diff
@@ -1,6 +1,6 @@
 use core::iter::Peekable;
 use core::str::Chars;
-use sqlparser::dialect::Dialect;
+use datafusion_sql::sqlparser::dialect::Dialect;
```
andygrove, line 3: We can use the version of sqlparser re-exported by DataFusion rather than manage this dependency ourselves

jdye64: Great!

jdye64 (approved):
Thanks for the version bump Andy!

jdye64:
@andygrove we had some upstream CI issues. Could you merge that latest upstream/datafusion-sql-planner and push again? I think that should sort out the CI build

andygrove:
@jdye64 upmerged and bumped the versions again - note there is a change to decimal 128/256 support now

PR #668, opened 2022-08-08
Generalize analyze/sample tests to resolve CI failures

charlesbluca:
Looks like CI failures cropped up with the release of Dask 2022.8.0, which I imagine made some subtle changes to the behavior of `Series.quantile` and `DataFrame.sample`.

The broken tests both were doing some pretty specific checks that didn't seem too resistant to breakage (e.g. asserting the exact value / length for fixed random results), so I went ahead and reimplemented these tests so that instead of checking against given values, we are now asserting that dask-sql's implementation matches that of Dask, which should hopefully make it so these tests don't fail as often moving forward.

### DIFF
--- a/tests/integration/test_analyze.py
+++ b/tests/integration/test_analyze.py
@@ -1,52 +1,31 @@
 import pandas as pd
 
+from dask_sql.mappings import python_to_sql_type
 from tests.utils import assert_eq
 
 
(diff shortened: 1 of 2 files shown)


ayushdg (approved):
lgtm! thanks Charles

PR #672, opened 2022-08-09
[DF] Resolve UDF test failures

charlesbluca:
This PR fixes a faulty closure in `get_function_meta` for the handling of UDF return types, which was responsible for the majority of the failures in `test_function.py`.


jdye64 (approved):
Great find and fix @charlesbluca looks good to me

ayushdg (approved):
Changes lgtm! Feel free to skip the multiple_definitions test and handle it in a followup pr if needed.

PR #673, opened 2022-08-10
Always raise error for duplicate UDF names if `replace=False`

charlesbluca:
Currently, `register_function` has unexpected behavior when attempting to register a `UDF` with an identical name, `func`, and `row_udf` but different input/output types with `replace=False`, with the schema's list of `FunctionDescriptions` getting updated with `FunctionDescriptions` of the new function but keeping those of the old one. This doesn't have an impact on main, but breaks things in the datafusion branch where additional function registration has to happen on the Rust end.

This PR makes it so we immediately raise an error if we see that the user is attempting to register a function that already exists with `replace=False`, rather than doing an equality check on the function, which IMO can cause issues if we add more attributes to the `UDF` object that need to be inspected; this should also open up a path in the datafusion branch to update the UDF replace logic to handle function registration in Rust.

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -909,19 +909,17 @@ class Context:
             f = UDF(f, row_udf, parameters, return_type)
         lower_name = name.lower()
         if lower_name in schema.functions:
-            if replace:
-                schema.function_lists = list(
-                    filter(
-                        lambda f: f.name.lower() != lower_name,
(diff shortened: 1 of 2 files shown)


charlesbluca:
Closing as I didn't think of the case where we will want to overload identical functions for different input types - we might need to think of a more nuanced situation for this for datafusion

PR #674, opened 2022-08-10
Bump arrow version to 6.0.0

charlesbluca:
conda-forge is bumping the supported range of Arrow versions to be 6-9 and dropping support for versions below 6 (xref https://github.com/conda-forge/conda-forge-pinning-feedstock/pull/3203), so we need to bump our minimum `pyarrow` testing dependency to 6.

cc @galipremsagar


galipremsagar (approved):

charlesbluca:
Looks like failures are Windows specific - looking into them now

charlesbluca:
The failures stemmed from an issue with MLFlow and older versions of the protobuf conda package that I opened in https://github.com/mlflow/mlflow/issues/6072; not really sure what can be done there but bumping from Arrow 6.0.0 -> 6.0.1 bumps our protobuf version to be new enough where it isn't a problem anymore.

PR #675, opened 2022-08-11
Stop overwriting aggregations on same column

### LINKED ISSUE #655: [BUG][Datafusion] Queries with multiple distinct aggregations return incorrect results
**What happened**:
Queries that include multiple distinct aggregations on the same column (ex: `SUM(a)` and `AVG(a)`) return incorrect results.

```python
import pandas as pd
from dask_sql import Context

c = Context()

df = pd.DataFrame({"a":[1,2,4]})
c.create_table("df", df)

c.sql("SELECT SUM(a), AVG(a) FROM df").compute()
```
returns:
```
   SUM(df.a)  AVG(df.a)
0        2.0        2.0
```
and:
```python
import pandas as pd
from dask_sql import Context

c = Context()

df = pd.DataFrame({"a":[1,2,4]})
c.create_table("df", df)

c.sql("SELECT AVG(a), SUM(a) FROM df").compute()
```
returns:
```
   AVG(df.a)  SUM(df.a)
0          7          7
```

**Environment**:
- dask-sql version: 8/1 datafusion build
- Python version: 3.9
- Operating System: Ubuntu 18.04.4
- Install method (conda, pip, source): source

ChrisJar:
Closes #655

### DIFF
--- a/dask_sql/physical/rel/logical/aggregate.py
+++ b/dask_sql/physical/rel/logical/aggregate.py
@@ -185,9 +185,17 @@ class DaskAggregatePlugin(BaseRelPlugin):
         if len(output_column_order) == 1 and output_column_order[0] == "UInt8(1)":
             backend_output_column_order = [df_agg.columns[0]]
         else:
+
+            def try_get_backend_by_frontend_name(oc):
+                try:
+                    return cc.get_backend_by_frontend_name(oc)
+                except KeyError:
+                    return oc
+
             backend_output_column_order = [
-                cc.get_backend_by_frontend_name(oc) for oc in output_column_order
+                try_get_backend_by_frontend_name(oc) for oc in output_column_order
             ]
+
         cc = ColumnContainer(df_agg.columns).limit_to(backend_output_column_order)
 
         cc = self.fix_column_to_row_type(cc, rel.getRowType())
@@ -370,7 +378,7 @@ class DaskAggregatePlugin(BaseRelPlugin):
                 )
 
             # Finally, extract the output column name
-            output_col = str(inputs[0].column_name(rel))
+            output_col = expr.toString()
 
             # Store the aggregation
             key = filter_column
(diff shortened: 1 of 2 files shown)


dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -185,9 +185,17 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
         if len(output_column_order) == 1 and output_column_order[0] == "UInt8(1)":
             backend_output_column_order = [df_agg.columns[0]]
         else:
+
+            def try_get_backend_by_frontend_name(oc):
+                try:
+                    return cc.get_backend_by_frontend_name(oc)
+                except KeyError:
+                    return oc
+
```
ayushdg, line 194: do you think this is worth moving to the column_container class? or do you think this is a one off case needed only here?

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -370,7 +378,7 @@ def _collect_aggregations(
                 )
 
             # Finally, extract the output column name
-            output_col = str(inputs[0].column_name(rel))
+            output_col = expr.toString()
```
ayushdg, line 381: Out of curiosity, to any of the PyExpr methods like `get_column_name` return the correct results, or do we have to rely on the stringified name for the Expr?

ayushdg (changes requested):
Could you add a test with the original issue highlighting this (or uncomment any existing one if being skipped) that tests for these cases

dask_sql/physical/rel/logical/aggregate.py
ChrisJar: Yeah the `column_name` function just returns the input column

dask_sql/physical/rel/logical/aggregate.py
ChrisJar: I think this is a pretty one off case

ayushdg (approved):
changes lgtm! @charlesbluca could you also take a look, since you're somewhat familiar with this part of the codebase

PR #682, opened 2022-08-11
[DF] Add handling for overloaded UDFs

charlesbluca:
This PR changes the implementation of `DaskFunctions` to support overloaded UDF definitions:

- the `return_type` attribute has been replaced with `return_types`, a `HashMap<Vec<PyDataType>, PyDataType>`, mapping the potential input types of a UDF to its return type
- this mapping is updated whenever an overloaded UDF definition is introduced
- the mapping is used as part of the `ScalarUDF`'s `ReturnTypeFunction` to assert which return type should be used for the given UDF call

Additional things that need to be done:

- [ ] clean up the implementation to cut down on unnecessary clones
- [ ] add handling to drop `DaskFunctions` when `replace=True` for `register_function`


dask_planner/src/sql.rs
```diff
@@ -140,12 +140,21 @@ impl ContextProvider for DaskSQLContext {
         }
 
         // Loop through all of the user defined functions
-        for schema in self.schemas.values() {
+        let schemas = self.schemas.clone();
+        for schema in schemas.values() {
```
andygrove, line 144: We no longer need this clone now that the function hash map contains `Arc` references that we clone before passing to the closure.

```suggestion
        for schema in self.schemas.values() {
```

dask_planner/src/sql.rs
```diff
@@ -140,12 +140,21 @@ impl ContextProvider for DaskSQLContext {
...
             for (fun_name, function) in &schema.functions {
                 if fun_name.eq(name) {
                     let sig = Signature::variadic(vec![DataType::Int64], Volatility::Immutable);
```
andygrove, line 147: Is it intentional that we are declaring all of the functions to accept a single parameter of type `Int64` ?

dask_planner/src/sql/schema.rs
```diff
@@ -29,7 +32,24 @@ impl DaskSchema {
...
+                (*e).lock()
+                    .unwrap()
+                    .return_types
+                    .insert(input_types.clone(), return_type.clone());
+            })
+            .or_insert(Arc::new(Mutex::new(DaskFunction::new(
+                name,
+                input_types,
+                return_type,
+            ))));
```
andygrove, line 53: We can use `or_insert_with` with a closure here so that we only create a new `DaskFunction` if we don't find one in the map. Kudos to Clippy for this suggestion.

```suggestion
            .or_insert_with(|| {
                Arc::new(Mutex::new(DaskFunction::new(
                    name,
                    input_types,
                    return_type,
                )))
            });
```

dask_planner/src/sql.rs
charlesbluca: Think that's just an artifact of copying the code from the above function meta definitions - I think what we would want here is to take the input types stored as the keys of ``function.return_types` and make multiple `TypeSignature::Exacts` that are then compiled in a `Signature::one_of`.

This still leaves something to be desire though, as IIUC this means that if we passed through inputs that didn't match any signatures, but could be coerced into matching input (e.g. we pass `Int64` columns into a function defined for `Float64`), we will error out, where the Calcite branch would do this coercion.

andygrove: I would expect DataFusion to also coerce the types.

jdye64 (approved):
I think there are some larger improvements we can make to custom functions as a whole. However, given the limitations we currently have these changes look good to me and think we should move forward with them until the larger changes are made.

Issue #320, opened 2021-11-16
[ENH] Support correlated subqueries


randerzander:
I'd like to be able to execute correlated subqueries.

It looks like the following example is successfully parsed by Calcite, but there's a logic error in mapping to Dask DataFrame calls:
```
import pandas as pd

df = pd.DataFrame({'id': [0, 1, 2], 'name': ['a', 'b', 'c'], 'val': [0, 1, 2]})

c.create_table('test', ddf)
c.sql("""
select name, val, id from test a
where val >
  (select avg(val) from test where id = a.id)
  """)
```

Result:
```
---------------------------------------------------------------------------
ValueError                                Traceback (most recent call last)
/tmp/ipykernel_1522695/1713961665.py in <module>
      6 
      7 c.create_table('test', ddf)
----> 8 c.sql("""
      9 select name, val, id from test a
     10 where val >

~/conda/envs/dsql-dask-main/lib/python3.8/site-packages/dask_sql/context.py in sql(self, sql, return_futures, dataframes)
    436         rel, select_names, _ = self._get_ral(sql)
    437 
--> 438         dc = RelConverter.convert(rel, context=self)
    439 
    440         if dc is None:

~/conda/envs/dsql-dask-main/lib/python3.8/site-packages/dask_sql/physical/rel/convert.py in convert(cls, rel, context)
     58             f"Processing REL {rel} using {plugin_instance.__class__.__name__}..."
     59         )
---> 60         df = plugin_instance.convert(rel, context=context)
     61         logger.debug(f"Processed REL {rel} into {LoggableDataFrame(df)}")
     62         return df

~/conda/envs/dsql-dask-main/lib/python3.8/site-packages/dask_sql/physical/rel/logical/project.py in convert(self, rel, context)
     27     ) -> DataContainer:
     28         # Get the input of the previous step
---> 29         (dc,) = self.assert_inputs(rel, 1, context)
     30 
     31         df = dc.df

~/conda/envs/dsql-dask-main/lib/python3.8/site-packages/dask_sql/physical/rel/base.py in assert_inputs(rel, n, context)
     83         from dask_sql.physical.rel.convert import RelConverter
     84 
---> 85         return [RelConverter.convert(input_rel, context) for input_rel in input_rels]
     86 
     87     @staticmethod

~/conda/envs/dsql-dask-main/lib/python3.8/site-packages/dask_sql/physical/rel/base.py in <listcomp>(.0)
[...]

PR #683, opened 2022-08-11
Correlated subqueries

sarahyurick:
Previously, we would get a `ValueError: Not all divisions are known, can't align partitions. Please use set_index to set the index.` for something like:

```
from dask_sql import Context
import dask.dataframe as dd
import pandas as pd

c = Context()

names = ["Miracle", "Sunshine", "Pretty woman", "Handsome man", "Barbie", "Cool painting", "Black square #1000", "Mountains"]
prices = [300, 700, 2800, 2300, 250, 5000, 50, 1300]
ids = [11, 12, 13, 14, 15, 16, 17, 18]
artist_id = [1, 1, 2, 2, 3, 3, 3, 4]
paintings = dd.from_pandas(pd.DataFrame({"id": ids, "name": names, "artist_id": artist_id, "listed_price": prices}), npartitions=1)
c.create_table("paintings", paintings)

sql1 = """
SELECT name, listed_price
FROM paintings
WHERE listed_price > (
    SELECT AVG(listed_price)
    FROM paintings
)
"""
c.sql(sql1).compute()
```

Not sure if this is the way we should go about this (not generalizable enough?), but here is an initial quick fix for that example. The general idea is that since we are comparing `listed_price` to a 1x1 table containing `AVG(listed_price)`, the latter has to be converted to a single value by calling `compute()` and with casting.

### DIFF
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -133,7 +133,13 @@ class ReduceOperation(Operation):
                 )
             ):
                 operands = tuple(map(as_timelike, operands))
-            return reduce(partial(self.operation, **kwargs), operands)
+            try:
+                return reduce(partial(self.operation, **kwargs), operands)
+            except ValueError:
+                return reduce(
+                    partial(self.operation, **kwargs),
+                    (operands[0], float(operands[1][operands[1].columns[0]].loc[0].compute()))
+                )
         else:
             return self.unary_operation(*operands, **kwargs)
 


sarahyurick:
Fixes example in #320

dask_sql/physical/rex/core/call.py
```diff
@@ -133,7 +133,13 @@ def reduce(self, *operands, **kwargs):
                 )
             ):
                 operands = tuple(map(as_timelike, operands))
-            return reduce(partial(self.operation, **kwargs), operands)
+            try:
+                return reduce(partial(self.operation, **kwargs), operands)
+            except ValueError:
+                return reduce(
+                    partial(self.operation, **kwargs),
+                    (operands[0], float(operands[1][operands[1].columns[0]].loc[0].compute()))
```
charlesbluca, line 141: I'd imagine we'd want to generalize the typecast here to handle other potential aggregating functions, but ATM can't think of an immediate way to do this.

charlesbluca (commented):
Thanks for opening @sarahyurick! While running through the issue repro you shared, I notice we also get some DataFusion warnings / errors:

```
Skipping optimizer rule decorrelate_scalar_subquery due to unexpected error: scalar subqueries must have a filter to be correlated at /home/nfs/charlesb/.cargo/git/checko
uts/arrow-datafusion-71ae82d9dec9a01c/6c32098/datafusion/optimizer/src/decorrelate_scalar_subquery.rs:177                                                                 
caused by                                                                                                                                                                 
Error during planning: Could not coerce into Filter! at /home/nfs/charlesb/.cargo/git/checkouts/arrow-datafusion-71ae82d9dec9a01c/6c32098/datafusion/expr/src/logical_plan
/plan.rs:1127
```

cc @andygrove in case you have some thoughts on this

Maybe @j

andygrove:
> cc @andygrove in case you have some thoughts on this

I filed an issue against DataFusion to add support for this type of query: https://github.com/apache/arrow-datafusion/issues/3266

sarahyurick:
Looks like this was resolved on the DataFusion side with https://github.com/apache/arrow-datafusion/pull/3287 !

PR #687, opened 2022-08-15
Use `compiler` function in nightly recipe, pin to Rust 1.62.1

charlesbluca:
- switches back to using the `compiler` function to set our Rust build requirement, which should be unblocked with https://github.com/conda-forge/rust-activation-feedstock/pull/22 in
- specifies the Rust build requirement with a conda build config file, as suggested by @jakirkham 
- pins our CI to this Rust version (1.62.1) to match with the nightly builds

### DIFF
--- a/continuous_integration/environment-3.10-dev.yaml
+++ b/continuous_integration/environment-3.10-dev.yaml
@@ -25,7 +25,7 @@ dependencies:
 - pytest-xdist
 - pytest
 - python=3.10
-- rust>=1.62.1
+- rust=1.62.1
 - scikit-learn>=1.0.0
 - setuptools-rust>=1.4.1
 - sphinx
--- a/continuous_integration/recipe/meta.yaml
+++ b/continuous_integration/recipe/meta.yaml
@@ -22,7 +22,7 @@ build:
 
 requirements:
   build:
-    - rust>=1.62.1
+    - {{ compiler('rust') }}
     - setuptools-rust>=1.4.1
   host:
     - pip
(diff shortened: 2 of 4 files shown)


continuous_integration/environment-3.10-dev.yaml
```diff
@@ -25,7 +25,7 @@ dependencies:
 - pytest-xdist
 - pytest
 - python=3.10
-- rust>=1.62.1
+- rust=1.62.1
```
jakirkham, line 28: Curious why we want to pin as opposed to just having a lower bound?

charlesbluca: I figured we would want our CI running against the same version of Rust we use to compile the conda packages, rather than potentially picking up new versions

continuous_integration/recipe/meta.yaml
```diff
@@ -22,7 +22,7 @@ build:
 
 requirements:
   build:
-    - rust>=1.62.1
+    - {{ compiler('rust') }}
     - setuptools-rust>=1.4.1
```
jakirkham, line 26: Would space this out and the one below. Know it isn't affected by this change. Just noticed it and though this would be a good opportunity to fix it

```suggestion
    - setuptools-rust >=1.4.1
```

continuous_integration/environment-3.10-dev.yaml
jakirkham: Ok. The main trick is keeping this up-to-date when it is pinned tightly like this.

jakirkham (commented):
Looks good Charles!

Had a very minor suggestion

jakirkham (approved):

ayushdg (approved):

PR #692, opened 2022-08-16
Configure `clippy` to error on warnings

charlesbluca:
Adds `-D warnings` to the `clippy` entrypoint in our pre-commit config so that warnings will now cause our linting to fail. Will also attempt to fix the existing warnings in this PR.


.pre-commit-config.yaml
```diff
@@ -24,7 +24,7 @@ repos:
       - id: cargo-check
         args: ['--manifest-path', './dask_planner/Cargo.toml', '--verbose', '--']
       - id: clippy
-        args: ['--manifest-path', './dask_planner/Cargo.toml', '--verbose', '--']
+        args: ['--manifest-path', './dask_planner/Cargo.toml', '--verbose', '--', '-D', 'warnings']
```
charlesbluca, line 27: Just want to note that we can also configure `clippy` to automatically fix some of the more easily corrected warnings with

```suggestion
        args: ['--manifest-path', './dask_planner/Cargo.toml', '--fix', '--allow-dirty', '--verbose', '--', '-D', 'warnings']
```

Was just hesitant to do this by default as the error message when leaving out `--allow-dirty` seemed to imply that this could result in destructive changes; @andygrove @jdye64 do you have any preferences here?

andygrove: I would prefer that we fix any warnings ourselves, even if we are often just copying & pasting Clippy's suggestions. I think this is an excellent way to learn more about writing idiomatic Rust code.

jdye64: Agree with Andy's comments here

jdye64:
For clarity is the plan to manually fix all the clippy warnings in this PR or a subsequent one?

charlesbluca:
@jdye64 plan is to manually fix all the clippy warnings in this PR, I've managed to narrow it down to only two:

```
error: deref on an immutable reference
   --> src/sql/types.rs:272:36
    |
272 |     pub fn from_string(input_type: &str) -> Self {
    |                                    ^ help: if you would like to reborrow, try removing `&*`: `&`
    |
    = note: `-D clippy::borrow-deref-ref` implied by `-D warnings`
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#borrow_deref_ref

error: deref on an immutable reference
   --> src/sql.rs:229:34
    |
229 |     pub fn parse_sql(&self, sql: &str) -> PyResult<Vec<statement::PyStatement>> {
    |                                  ^ help: if you would like to reborrow, try removing `&*`: `&`
    |
    = help: for further information visit https://rust-lang.github.io/rust-clippy/master/index.html#borrow_deref_ref

error: could not compile `dask_planner` due to 2 previous errors
```

jdye64 (approved):
LGTM

PR #693, opened 2022-08-16
Rust parsing support for CREATE MODEL statements

jdye64:


ayushdg:
cc: @andygrove for reviews as well.

andygrove:
I left some suggestions on the DROP MODEL PR, which I now realize I should have left here instead.

dask_planner/src/parser.rs
```diff
@@ -0,0 +1,183 @@
...
+    /// Parse a SQL CREATE statement
+    pub fn parse_create(&mut self) -> Result<DaskStatement, ParserError> {
+        match self.parser.parse_identifier() {
+            Ok(ident) => match ident.value.to_lowercase().as_str() {
+                "create" => self.parse_create_model(),
+                _ => Ok(DaskStatement::Statement(Box::from(
+                    self.parser.parse_create()?,
+                ))),
+            },
+            Err(e) => Err(e),
```
ayushdg, line 140: Given that there are other sql statements such as `create table/create view` which also start with create we probably need to specifically check for the create model statement here.

dask_planner/src/parser.rs
```diff
@@ -0,0 +1,183 @@
...
+                Ok(DaskStatement::Statement(Box::from(
+                    self.parser.parse_statement()?,
+                )))
+            }
+        }
+    }
+
+    /// Parse a SQL CREATE statement
+    pub fn parse_create(&mut self) -> Result<DaskStatement, ParserError> {
+        match self.parser.parse_identifier() {
```
ayushdg, line 133: we could use the `parse_keyword` method here since `create` is a recognized keyword in sqlparser. 

More generally if this was not recognized as a keyword by sqlparser, would we have to create our own keyword list in dask-sql?

ayushdg (changes requested):

dask_planner/src/parser.rs
jdye64: doh! Yes, that was dumb. It is suppose to be "model" not "create" will change that

dask_planner/src/parser.rs
jdye64: As comment ^^ suggests I had just made a typo. Should be "model" not "create". "model" is not a `Keyword` so we use Ident instead.

jdye64:
> I left some suggestions on the DROP MODEL PR, which I now realize I should have left here instead.

No worries I noticed that and tried to address those requests here manually. I'm pretty sure I got them all but might want to look again after CI passes.

dask_planner/src/parser.rs
```diff
@@ -0,0 +1,196 @@
...
+                }
+            }
+            _ => {
+                // use the native parser
+                Ok(DaskStatement::Statement(Box::from(
+                    self.parser.parse_create()?,
+                )))
+            }
+        }
+    }
```
ayushdg, line 157: I wonder if there's a common utility we could move this logic to that handles w.keywords and/or w.value since I see this being reused for a lot of custom sql syntax.

dask_planner/src/sql.rs
```diff
@@ -289,6 +290,30 @@ impl DaskSQLContext {
...
+        match dask_statement {
+            DaskStatement::Statement(statement) => {
+                let planner = SqlToRel::new(self);
+                planner.statement_to_plan(DFStatement::Statement(statement))
+            }
+            DaskStatement::CreateModel(create_model) => Ok(LogicalPlan::Extension(Extension {
+                node: Arc::new(CreateModelPlanNode {
+                    model_name: create_model.name,
+                    input: self._logical_relational_algebra(DaskStatement::Statement(Box::new(
+                        create_model.select,
```
ayushdg, line 309: Does this ensure that the optimization rules in dask-sql also get applied to the select part of the statement?

ayushdg (approved):
Minor questions for my own info but changes generally lgtm! Thanks a lot @jdye64.

cc: @andygrove if you have another round of reviews before we merge this one in.

PR #695, opened 2022-08-16
Support for DROP MODEL parsing in Rust

jdye64:


dask_planner/src/parser.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub fn parse_create(&mut self) -> Result<DaskStatement, ParserError> {
+        match self.parser.parse_identifier() {
+            Ok(ident) => match ident.value.to_lowercase().as_str() {
+                "create" => self.parse_create_model(),
+                _ => Ok(DaskStatement::Statement(Box::from(
+                    self.parser.parse_create()?,
+                ))),
+            },
+            Err(e) => Err(e),
+        }
```
andygrove, line 156: No need to match on the `Result`. We can just use `?` instead.

```suggestion
        let ident = self.parser.parse_identifier()?;
        match ident.value.to_lowercase().as_str() {
            "create" => self.parse_create_model(),
            _ => Ok(DaskStatement::Statement(Box::from(
                self.parser.parse_create()?,
            ))),
        }
```

dask_planner/src/parser.rs
```diff
@@ -0,0 +1,221 @@
...
+    pub fn parse_drop(&mut self) -> Result<DaskStatement, ParserError> {
+        match self.parser.parse_identifier() {
+            Ok(ident) => match ident.value.to_lowercase().as_str() {
+                "model" => self.parse_drop_model(),
+                _ => Ok(DaskStatement::Statement(Box::from(
+                    self.parser.parse_drop()?,
+                ))),
+            },
+            Err(e) => Err(e),
+        }
```
andygrove, line 169: Same comment here:

```suggestion
        let ident = self.parser.parse_identifier()?;
        match ident.value.to_lowercase().as_str() {
            "model" => self.parse_drop_model(),
            _ => Ok(DaskStatement::Statement(Box::from(
                self.parser.parse_drop()?,
            ))),
        }
```

dask_planner/src/parser.rs
```diff
@@ -0,0 +1,221 @@
...
+                            | Value::DoubleQuotedString(key_val) => key_val,
+                            _ => "".to_string(),
+                        },
+                        _ => "".to_string(),
+                    },
+                    right,
+                ),
+                _ => panic!("Expected BinaryOp, Key/Value pairs, found: {}", f),
+            })
+            .collect();
```
andygrove, line 198: We can avoid the panic here and use Result instead.

```suggestion
        let _kv_pairs: Vec<(String, &Box<Expr>)> = kv_binexprs
            .iter()
            .map(|f| match f {
                Expr::BinaryOp { left, op: _, right } => match *left.clone() {
                    Expr::Value(value) => match value {
                        Value::EscapedStringLiteral(key_val)
                        | Value::SingleQuotedString(key_val)
                        | Value::DoubleQuotedString(key_val) => Ok((key_val, right)),
                        _ => Ok(("".to_string(), right)),
                    },
                    _ => Ok(("".to_string(), right)),
                },

                _ => parser_err!(format!("Expected BinaryOp, Key/Value pairs, found: {}", f)),
            })
            .collect::<Result<Vec<_>, ParserError>>()?;
        let _kv_pairs: HashMap<String, &Box<Expr>> = _kv_pairs.into_iter().collect();

```

dask_planner/src/sql/logical.rs
```diff
@@ -183,12 +193,32 @@ impl PyLogicalPlan {
...
             LogicalPlan::Analyze(_analyze) => "Analyze",
-            LogicalPlan::Extension(_extension) => "Extension",
             LogicalPlan::Subquery(_sub_query) => "Subquery",
             LogicalPlan::SubqueryAlias(_sqalias) => "SubqueryAlias",
             LogicalPlan::CreateCatalogSchema(_create) => "CreateCatalogSchema",
             LogicalPlan::CreateCatalog(_create_catalog) => "CreateCatalog",
             LogicalPlan::CreateView(_create_view) => "CreateView",
+            // Further examine and return the name that is a possible Dask-SQL Extension type
+            LogicalPlan::Extension(extension) => {
+                if let true = extension
```
andygrove, line 203:
```suggestion
                if extension
```

dask_planner/src/parser.rs
andygrove: The magic here is that returning results in the `iter` gives us a `Vec<Result<_>>` which we can transform into `<Result<Vec_>>` using the turbofish notation in `collect::<Result<Vec<_>, ParserError>>()?`

andygrove (approved):
LGTM

PR #696, opened 2022-08-16
Refactor LIMIT computation to always use `head` when possible

### LINKED ISSUE #686: [DF] select * limit 5 seems does a full scan
I'm struggling to find a programmatic reproducer for this, but on the datafusion-sql-planner branch:
```
c.sql("SELECT * FROM large_table limit 5")
```
results in reading the entire dataset before filtering at the end, instead of reading from a single partition.

Less reproducible, but from the daily weather data:
```
res = c.sql("select * from weather limit 5")
io_layer = list(res.dask.layers.keys())[0]
partitions = len(list(res.dask.layers[io_layer].keys()))
partitions
```
```
445
```

If I understand layers correctly, my select w/ limit statement is reading all partitions in the dataset.

charlesbluca:
Takes @sarahyurick's suggestion to use `map_partitions` to compute partition lengths in our LIMIT check, which should allow us to always use `head` to compute a LIMIT when no OFFSET is specified.

Also added in a configuration variable to control whether or not we actually compute this `map_partitions` call, which we may or may not want depending on query complexity and table size.

Closes #686

### DIFF
--- a/dask_sql/physical/rel/logical/limit.py
+++ b/dask_sql/physical/rel/logical/limit.py
@@ -2,6 +2,8 @@ from typing import TYPE_CHECKING
 
 import dask.dataframe as dd
 from dask import config as dask_config
+from dask.blockwise import Blockwise
+from dask.layers import DataFrameIOLayer
 
 from dask_sql.datacontainer import DataContainer
 from dask_sql.physical.rel.base import BaseRelPlugin
@@ -51,24 +53,27 @@ class DaskLimitPlugin(BaseRelPlugin):
         Unfortunately, Dask does not currently support row selection through `iloc`, so this must be done using a custom partition function.
         However, it is sometimes possible to compute this window using `head` when an `offset` is not specified.
         """
+        # if no offset is specified we can use `head` to compute the window
+        if not offset:
+            # if `check-first-partition` enabled, check if we have a relatively simple Dask graph and if so,
+            # check if the first partition contains our desired window
+            if (
+                dask_config.get("sql.limit.check-first-partition")
+                and all(
+                    [
+                        isinstance(layer, (DataFrameIOLayer, Blockwise))
+                        for layer in df.dask.layers.values()
+                    ]
+                )
+                and end <= len(df.partitions[0])
+            ):
+                return df.head(end, compute=False)
+
+            return df.head(end, npartitions=-1, compute=False)
+
         # compute the size of each partition
         # TODO: compute `cumsum` here when dask#9067 is resolved
         partition_borders = df.map_partitions(lambda x: len(x))
-
-        # if no offset is specified we can use `head` to compute the window
-        if not offset:
-            if dask_config.get("sql.limit.check-partitions"):
-                partition_borders = partition_borders.compute()
-                nrows = 0
-                npartitions = 0
-                for n in partition_borders:
-                    nrows += n
-                    npartitions += 1
-                    if end <= nrows:
-                        break
-            else:
-                npartitions = -1
-            return df.head(end, npartitions=npartitions, compute=False)
 
         def limit_partition_func(df, partition_borders, partition_info=None):
             """Limit the partition to values contained within the specified window, returning an empty dataframe if there are none"""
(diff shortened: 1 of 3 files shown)


dask_sql/physical/rel/logical/limit.py
```diff
@@ -50,17 +51,24 @@ def _apply_limit(self, df: dd.DataFrame, offset: int, end: int) -> dd.DataFrame:
...
-            if first_partition_length >= end:
-                return df.head(end, compute=False)
-
         # compute the size of each partition
         # TODO: compute `cumsum` here when dask#9067 is resolved
         partition_borders = df.map_partitions(lambda x: len(x))
 
+        # if no offset is specified we can use `head` to compute the window
+        if not offset:
+            if dask_config.get("sql.limit.check-partitions"):
```
charlesbluca, line 60: Right now this would be defaulting to true, so that we would trigger a graph computation on all LIMITs; interested in what others think a sensible default here would be

dask_sql/physical/rel/logical/limit.py
```diff
@@ -50,17 +51,24 @@ def _apply_limit(self, df: dd.DataFrame, offset: int, end: int) -> dd.DataFrame:
...
+                partition_borders = partition_borders.compute()
```
ayushdg, line 61: With this config set to true we compute the whole task graph prior to get partition borders and then return head with `npartitions<df.npartitions`
Without this config we return head with all partitions computed at runtime. 
So in this scenario wouldn't we end up computing the whole task graph at least once, no matter which config was provided?

dask_sql/physical/rel/logical/limit.py
```diff
@@ -50,17 +51,24 @@ def _apply_limit(self, df: dd.DataFrame, offset: int, end: int) -> dd.DataFrame:
...
-            # is already enough, we will just use this
-            first_partition_length = len(df.partitions[0])
-            if first_partition_length >= end:
-                return df.head(end, compute=False)
-
         # compute the size of each partition
         # TODO: compute `cumsum` here when dask#9067 is resolved
         partition_borders = df.map_partitions(lambda x: len(x))
 
+        # if no offset is specified we can use `head` to compute the window
```
sarahyurick, line 58: For comparison, the implementation I used that worked for simple queries was:
```
# if no offset is specified we can use `head` to compute the window
if not offset:
    partition_divisions = partition_borders.divisions
    try:
        ind = list(map(lambda i: i >= limit, partition_borders)).index(True)
    except ValueError:
        ind = -1
    return df.head(limit, npartitions=-1, compute=False)
```

dask_sql/physical/rel/logical/limit.py
charlesbluca: Hmm, I see what you mean - we do save one graph compute setting `check-partitions` to false, but I forgot that the `map_partitions` call forces a persist of the entire table versus a single partition with something like `len(df.partitions[0])`, so we're not getting the memory saving that was originally intended with the check 😕

dask_sql/physical/rel/logical/limit.py
charlesbluca: Thanks for sharing @sarahyurick! After further exploration of using `divisions` to keep track of partition length, I realize it might be difficult to handle certain edge cases - for example, the following input table

```python
from dask.datasets import timeseries

ddf = timeseries().set_index("x")
ddf.divisions
# (-0.9999997982529687, -0.9243231218501866, ..., 0.9017005336188877, 0.9999980053737976)
```

has divisions that don't coincide with the number of rows in each partition. Unless there's a relatively cheap way to check if a table's index is essentially an unaltered `RangeIndex`, this might push us back to using the partition lengths directly.

sarahyurick: Ah ok, I see what you mean

charlesbluca:
rerun tests

charlesbluca:
rerun tests

ayushdg (approved):

PR #697, opened 2022-08-17
Parsing logic for SHOW SCHEMAS

jdye64:


dask_planner/src/sql/logical.rs
```diff
@@ -183,12 +200,39 @@ impl PyLogicalPlan {
...
             LogicalPlan::Analyze(_analyze) => "Analyze",
-            LogicalPlan::Extension(_extension) => "Extension",
             LogicalPlan::Subquery(_sub_query) => "Subquery",
             LogicalPlan::SubqueryAlias(_sqalias) => "SubqueryAlias",
             LogicalPlan::CreateCatalogSchema(_create) => "CreateCatalogSchema",
             LogicalPlan::CreateCatalog(_create_catalog) => "CreateCatalog",
             LogicalPlan::CreateView(_create_view) => "CreateView",
+            // Further examine and return the name that is a possible Dask-SQL Extension type
+            LogicalPlan::Extension(extension) => {
+                if extension
```
andygrove, line 210: it might be worth adding `let node = extension.node.as_any()` so we don't have to repeat `extension.node.as_any()` in each `if` statement

dask_planner/src/sql/logical/show_schema.rs
```diff
@@ -0,0 +1,98 @@
...
+    }
+
+    fn schema(&self) -> &DFSchemaRef {
+        &self.schema
+    }
+
+    fn expressions(&self) -> Vec<Expr> {
+        // there is no need to expose any expressions here since DataFusion would
+        // not be able to do anything with expressions that are specific to
+        // DROP MODEL
```
andygrove, line 41:
```suggestion
        // SHOW SCHEMAS
```

dask_planner/src/sql/logical/show_schema.rs
```diff
@@ -0,0 +1,98 @@
...
+}
+
+#[pymethods]
+impl PyShowSchema {
+    #[pyo3(name = "getLike")]
+    fn get_like(&self) -> PyResult<String> {
+        match &self.show_schema.like {
+            Some(e) => Ok(e.clone()),
+            None => Ok("".to_string()),
+        }
```
andygrove, line 73: This could also be written in a functional style, but I'm not sure if it is worth changing :man_shrugging: 

```suggestion
        Ok(self
            .show_schema
            .like
            .as_ref()
            .map(|e| e.clone())
            .unwrap_or_else(|| "".to_string()))
```

dask_planner/src/sql/logical/show_schema.rs
```diff
@@ -0,0 +1,98 @@
...
+                {
+                    Ok(PyShowSchema {
+                        show_schema: ext.clone(),
+                    })
+                } else {
+                    Err(py_type_err("unexpected plan"))
+                }
+            }
+            _ => Err(py_type_err("unexpected plan")),
+        }
```
andygrove, line 96: Here is another approach that removes the need to have two places to return an `Err`. It is quite verbose though so not sure if this is an improvement or not.

```suggestion
        match logical_plan {
            LogicalPlan::Extension(Extension { node })
                if node
                    .as_any()
                    .downcast_ref::<ShowSchemasPlanNode>()
                    .is_some() =>
            {
                let ext = node
                    .as_any()
                    .downcast_ref::<ShowSchemasPlanNode>()
                    .expect("ShowSchemasPlanNode");
                Ok(PyShowSchema {
                    show_schema: ext.clone(),
                })
            }
            _ => Err(py_type_err("unexpected plan")),
        }
```

andygrove (commented):
Left a couple of code style suggestions, but LGTM.

dask_planner/src/sql/logical/show_schema.rs
jdye64: Well I need to sharpen my functional skills so gonna change it, thanks for the suggestion!

continuous_integration/gpuci/environment.yaml
```diff
@@ -15,3 +15,30 @@ dependencies:
   - ucx-proc=*=gpu
   - ucx-py=0.27
   - xgboost=*=cuda_*
+  - dask-ml>=2022.1.22
```
jdye64, line 18: @charlesbluca I added these dependencies to the gpu script. I know we don't need them for CI but since we are using this as the place for the "dev" environment file it seemed like they were needed. You ok with this?

charlesbluca: My preference would be to keep these dependencies out of this file for now and use the method of updating the existing CI environment with this one, which is what the gpuCI images are currently doing:

https://github.com/rapidsai/dask-build-environment/blob/main/dask_sql.Dockerfile

If we do ever end up refactoring the image builds to do a single environment creation, then maybe it would make more sense to add these dependencies - otherwise, it just adds another environment file to keep track of that isn't being monitored by any form of CI IMO

PR #698, opened 2022-08-17
Uncomment test_aggregations in groupby pytests

ChrisJar:


jdye64:
@ChrisJar this is marked as draft, is it ready for review? If so can you mark as ready for review?

ChrisJar:
It's not quite ready for review. It's blocked by a typecasting issue that @charlesbluca was looking into.

jdye64:
@ChrisJar I updated the branch with the latest from upstream. I shouldn't have done that and hope it didn't mess you up locally. Might want to run a `git pull` locally to get those changes

charlesbluca:
Tried uncommenting `test_aggregations`, looks like the issue that we're facing here is a failure to parse queries with aggregations on binary expressions:

```
RuntimeError: SchemaError(FieldNotFound { qualifier: Some("user_table_1"), name: "b", valid_fields: Some(["user_table_1.user_id", "every(user_table_1.b = Int64(3))", "bit_and(user_table_1.b)", "bit_or(user_table_1.b)", "MIN(user_table_1.b)", "single_value(user_table_1.b)", "AVG(user_table_1.b)"]) })
```

The underlying issue here is that we don't have a branch to handle `Expr::BinaryExpr` in [`expr_to_field`](https://github.com/dask-contrib/dask-sql/blob/6c4c5d3e018da8e4400704a2421f8e13b3fad884/dask_planner/src/expression.rs#L704-L716), and so we end up passing the expression directly to `datafusion_expr::utils::exprlist_to_fields`, which gives the above schema error.

 Not sure if `exprlist_to_fields` should be able to get the field from this expression (from the error message it looks like it's included in `valid_fields`), or if we should be adding specific logic for the `BinaryExpr` case here.

jdye64:
@charlesbluca and @ChrisJar I suspect that https://github.com/dask-contrib/dask-sql/pull/732 will help with this issue. After it merges I would suggest merging with upstream and hopefully that resolves the issue.

ChrisJar:
Changes added in #738

PR #699, opened 2022-08-17
Support for parsing SHOW TABLES FROM grammar

jdye64:


jdye64:
I'm thinking and hoping that we might actually be able to leverage DataFusion for this and not need the custom parsing logic. I'm still looking around and experimenting and will report back my findings. Would still need the `PyShowTables` logic just could remove the parsing logic which would be nice

jdye64:
Ok, according the @andygrove we actually should be doing it this way so this PR can be reviewed now.

andygrove (approved):
LGTM

PR #702, opened 2022-08-18
Don't check order for filtered groupby test

### LINKED ISSUE #701: ⚠️ Upstream CI failed ⚠️
[Workflow Run URL](https://github.com/dask-contrib/dask-sql/actions/runs/2900737686)

charlesbluca:
Looks like changes in Dask have made the ordering of one of the results in `test_group_by_filtered` different, causing failures - this PR adds `check_index=False` to the relevant assertion, since maintaining a sorted ordering for groupbys wasn't a guarantee to begin with.

Perhaps it would be worth thinking about if we want to make this a default kwarg for our `assert_eq` shim, considering we aren't concerned with ordering unless something like an `ORDER BY` statement is being tested (cc @ayushdg).

Closes #701

### DIFF
--- a/tests/integration/test_groupby.py
+++ b/tests/integration/test_groupby.py
@@ -109,7 +109,7 @@ def test_group_by_filtered(c):
         },
     )
 
-    assert_eq(return_df, expected_df)
+    assert_eq(return_df, expected_df, check_index=False)
 
     return_df = c.sql(
         """


ayushdg (approved):
> Perhaps it would be worth thinking about if we want to make this a default kwarg for our assert_eq shim, considering we aren't concerned with ordering unless something like an ORDER BY statement is being tested (cc @ayushdg).

Thanks for triaging. I agree that it makes sense to set this to false for all groupby (and maybe join) related tests. But I do think  for standard rex calls and other tests we do want to ensure that input ordering is maintained (similar to sql). So it might make sense to limit this to groupby (and join) tests for now.

Unrelated but gpuCi failures will probably be fixed via https://github.com/rapidsai/cudf/pull/11561

charlesbluca:
Going to merge this for now to unblock CI, we can think about relaxing the `assert_eq` checks for specific tests in a follow up

PR #714, opened 2022-08-23
Datafusion create table with

jdye64:


dask_planner/src/parser.rs
```diff
@@ -375,4 +437,48 @@ impl<'a> DaskParser<'a> {
             schema_name,
         })))
     }
+
+    /// Retrieves the table_schema and table_name from a `TableFactor` instance
+    fn elements_from_tablefactor(tbl_factor: &TableFactor) -> (String, String) {
```
charlesbluca, line 442: Would it make sense to break this function (along with`options_from_tablefactor`)  off into a separate module? It seems like a utility for `parse_create_table` that contrasts with the other `parse_*` methods of this module.

dask_sql/utils.py
```diff
@@ -134,44 +130,23 @@ def __str__(self):
...
     """
 
-    def convert_literal(value):
-        if isinstance(value, "LogicalPlan"):
-            operator_mapping = {
-                "ARRAY": list,
-                "MAP": lambda x: dict(zip(x[::2], x[1::2])),
-                "MULTISET": set,
-                "ROW": tuple,
-            }
```
charlesbluca, line 154: Out of scope, but I realize we really aren't testing any of these operator mappings in our create table with tests - might make sense to add those in later on if we intend to add back support for these operators in a follow up

dask_planner/src/parser.rs
```diff
@@ -375,4 +437,48 @@ impl<'a> DaskParser<'a> {
...
+
+                assert!(identities.len() <= 2 && !identities.is_empty());
+
+                if identities.len() == 1 {
+                    ("".to_string(), identities[0].clone())
+                } else if identities.len() == 2 {
+                    (identities[0].clone(), identities[1].clone())
+                } else {
+                    ("".to_string(), "".to_string())
+                }
```
andygrove, line 460: The `else` seems to be unreachable here, and I think this is a better fit for a `match` statement:


```suggestion
                match identities.len() {
                    1 => ("".to_string(), identities[0].clone()),
                    2 => (identities[0].clone(), identities[1].clone()),
                    _ => panic!("TableFactor name only supports 1 or 2 elements")
                }
```

andygrove (approved):
LGTM. Left one suggestion.

jdye64:
@charlesbluca what do you think about the `DaskParserUtils` I moved them to?

charlesbluca (approved):

PR #715, opened 2022-08-23
Support for DISTRIBUTE BY

jdye64:


dask_planner/src/sql.rs
```diff
@@ -241,10 +241,6 @@ impl DaskSQLContext {
                 for statement in k {
                     statements.push(statement.into());
                 }
-                assert!(
```
jdye64, line 244: I had added this long ago. My understanding of how this works has since changed so I removed this snippet as it is completely not necessary and in fact breaks running multiple `;` delimited queries

PR #717, opened 2022-08-24
Add support for CREATE MODE and DROP MODEL

ChrisJar:
This add support for `CREATE MODEL` and `DROP MODEL` queries.

Note: pytests require `PREDICT`


jdye64:
@ChrisJar It seems like the PR I opened contained a lot of these changes, I used the branch we were sharing yesterday so they are your commits and code. Any chance you could update this branch with upstream so we can see any differences? Sorry about that confusion

PR #718, opened 2022-08-24
[DF] Add support for `CREATE [OR REPLACE] TABLE [IF NOT EXISTS] WITH`

charlesbluca:
Updates the custom parsing logic for `CREATE TABLE WITH` statements to support `IF NOT EXISTS` and `OR REPLACE`; we could probably do the same for `CREATE MODEL`, not sure if we want to do this now or wait until we start rewriting the Python plugin.

### DIFF
--- a/dask_planner/src/parser.rs
+++ b/dask_planner/src/parser.rs
@@ -358,7 +358,7 @@ impl<'a> DaskParser<'a> {
         Ok(DaskStatement::CreateModel(Box::new(create)))
     }
 
-    /// Parse Dask-SQL CREATE TABLE ... WITH statement
+    /// Parse Dask-SQL CREATE [OR REPLACE] TABLE ... statement
     fn parse_create_table(&mut self, or_replace: bool) -> Result<DaskStatement, ParserError> {
         // parse [IF NOT EXISTS] `table_name` AS|WITH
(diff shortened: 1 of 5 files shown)


jdye64 (approved):
Thanks @charlesbluca LGTM

dask_planner/src/parser.rs
```diff
@@ -381,14 +394,23 @@ impl<'a> DaskParser<'a> {
...
+                            if_not_exists,
+                            or_replace,
                             with_options,
                         };
                         Ok(DaskStatement::CreateTable(Box::new(create)))
                     }
+                    _ => self.expected("'as' or 'with'", self.parser.peek_token()),
                 }
             }
             _ => {
```
charlesbluca, line 406: Wondering what this catchall would be catching at this point - are there other forms of `CREATE TABLE` statements not captured above, or can we just raise a parser error at this point?

ayushdg (approved):

dask_planner/src/parser.rs
ayushdg: Thinking about this a bit more, do we need the `as` branch or could we just pass the catchall case back to the parser to figure things out/error as needed.

PR #719, opened 2022-08-24
Support all boolean operations

sarahyurick:
With the recent updates to DataFusion and minor Dask SQL changes, "IS TRUE/FALSE" and "IS [NOT] UNKNOWN" are functional.

"IS NOT TRUE" and "IS NOT FALSE" should be functional once https://github.com/apache/arrow-datafusion/pull/3252 is merged.

UPDATE: As long as all checks pass, this should be ready to merge.

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -625,7 +625,7 @@ impl PyExpr {
     pub fn bool_value(&mut self) -> PyResult<bool> {
         match &self.expr {
             Expr::Literal(scalar_value) => match scalar_value {
-                ScalarValue::Boolean(iv) => Ok(iv.unwrap()),
+                ScalarValue::Boolean(Some(iv)) => Ok(*iv),
                 _ => Err(py_type_err("getValue<T>() - Unexpected value")),
             },
             _ => Err(py_type_err("getValue<T>() - Non literal value encountered")),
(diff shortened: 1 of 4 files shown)


dask_sql/physical/rex/core/literal.py
```diff
@@ -101,8 +101,12 @@ def convert(
...
         # type name back to a SQL type
         if literal_type == "Boolean":
-            literal_type = SqlTypeName.BOOLEAN
-            literal_value = rex.getBoolValue()
+            try:
+                literal_type = SqlTypeName.BOOLEAN
+                literal_value = rex.getBoolValue()
+            except TypeError:
+                literal_type = SqlTypeName.NULL
+                literal_value = None
```
charlesbluca, line 109: Out of interest, what type error are we catching here?

dask_planner/src/expression.rs
```diff
@@ -625,7 +625,7 @@ impl PyExpr {
     pub fn bool_value(&mut self) -> PyResult<bool> {
         match &self.expr {
             Expr::Literal(scalar_value) => match scalar_value {
-                ScalarValue::Boolean(iv) => Ok(iv.unwrap()),
+                ScalarValue::Boolean(Some(iv)) => Ok(*iv),
```
sarahyurick, line 628: @charlesbluca your question is related to the change here - originally, "IS UNKNOWN" would panic here when the value is null, then throws the type error on the next line.

Not sure if we can come up with a case where it wouldn't be `Some(true)`, `Some(false)`, or `None`

sarahyurick:
Hi @charlesbluca @ayushdg @galipremsagar @jdye64 this is ready for review now

jdye64 (approved):
Outside of suggestions already made by @charlesbluca the rest looks good to me, will go ahead and approve from my end

charlesbluca (approved):

PR #720, opened 2022-08-24
[DF] Bump DataFusion to rev 076b42

andygrove:
This picks up recent support for:

- `IS [NOT] TRUE / FALSE / UNKNOWN`
- `DISTRIBUTE BY`


jdye64 (approved):
Thanks Andy

PR #722, opened 2022-08-25
Implment PREDICT parsing and python wiring

jdye64:


dask_planner/src/parser.rs
```diff
@@ -164,6 +182,44 @@ impl<'a> DaskParser<'a> {
...
+                                }
+                                _ => {
+                                    // Keep looking for PREDICT
+                                    cnt += 1;
+                                    continue;
+                                }
+                            }
+                        }
+
+                        // Reset the parser back to where we started
```
jdye64, line 213: Wanted to use `self.parser.index = 0` but `index` is private so had to resort to this abomination

dask_planner/src/parser.rs
```diff
@@ -319,41 +384,70 @@ impl<'a> DaskParser<'a> {
...
+    /// Parse a SQL PREDICT statement
+    pub fn parse_predict_model(&mut self) -> Result<DaskStatement, ParserError> {
+        // PREDICT(
+        //     MODEL model_name,
+        //     SQLStatement
+        // )
+        self.parser.expect_token(&Token::LParen)?;
+
+        // Expect `MODEL`
+        // This doesn't work? Somethign about Keyword in the equality comparison, therefore always fails.
```
jdye64, line 396: @andygrove when I tried to use `expect_token` it would fail with error saying `MODEL, not equal MODEL` I believe it was because the `Keyword` field was not set there? Either way would like to use this syntax it just wasn't working

dask_planner/src/parser.rs
```diff
@@ -319,41 +384,70 @@ impl<'a> DaskParser<'a> {
...
+        //     quote_style: None,
+        //     keyword: Keyword::NONE,
+        // }))?;
+
+        let is_model = match self.parser.next_token() {
+            Token::Word(w) => matches!(w.value.to_lowercase().as_str(), "model"),
+            _ => false,
+        };
+        if !is_model {
+            panic!("Failed to parse expected model but not found!")
```
andygrove, line 408: Can we return an `Err` here rather than causing a panic?

dask_planner/src/sql/logical/predict_model.rs
```diff
@@ -0,0 +1,104 @@
...
+    }
+
+    fn schema(&self) -> &DFSchemaRef {
+        self.input.schema()
+    }
+
+    fn expressions(&self) -> Vec<Expr> {
+        // there is no need to expose any expressions here since DataFusion would
+        // not be able to do anything with expressions that are specific to
+        // CREATE TABLE
```
andygrove, line 44:
```suggestion
        // PREDICT MODEL
```

PR #723, opened 2022-08-25
[DF] Add `TypeCoercion` optimizer rule

andygrove:
I am working on adding this rule to DataFusion (see https://github.com/apache/arrow-datafusion/pull/3222 and https://github.com/apache/arrow-datafusion/pull/3254), but we may as well add it directly to the project for now so we can iterate faster on the rules.

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -4,6 +4,7 @@ pub mod function;
 pub mod logical;
 pub mod optimizer;
 pub mod parser_utils;
+pub mod rules;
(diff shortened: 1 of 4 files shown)


andygrove:
@charlesbluca fyi

jdye64 (approved):
Yeah, this is 100% awesome! Had already reviewed in DataFusion. Think it makes complete sense to iterate on it here for a bit and then can back it out once it is finalized in DataFusion. We have done that several times already anyway.

PR #726, opened 2022-08-26
Allows for compound identifiers in the `qualifier`/schema field

jdye64:
Previously if a compound identifier like `root.df` is encountered the index will fail to be located. This causes the results of the operation to be incorrect. This addition ensures that the compound identifier is honored.

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -625,7 +625,7 @@ impl PyExpr {
     pub fn bool_value(&mut self) -> PyResult<bool> {
         match &self.expr {
             Expr::Literal(scalar_value) => match scalar_value {
-                ScalarValue::Boolean(iv) => Ok(iv.unwrap()),
+                ScalarValue::Boolean(Some(iv)) => Ok(*iv),
                 _ => Err(py_type_err("getValue<T>() - Unexpected value")),
             },
             _ => Err(py_type_err("getValue<T>() - Non literal value encountered")),
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -158,34 +158,23 @@ impl PyExpr {
                     let name: Result<String> = self.expr.name(input_plans[0].schema());
                     match name {
                         Ok(fq_name) => {
-                            let mut idx: usize = 0;
-                            for schema in input_plans[0].all_schemas() {
-                                let (schema_name_opt, column_name) =
-                                    self._schema_column_from_fqn(&fq_name);
-                                match schema_name_opt {
-                                    Some(e) => {
-                                        match schema
-                                            .index_of_column_by_name(Some(e.as_str()), column_name)
-                                        {
-                                            Ok(e) => {
-                                                idx = e;
-                                                break;
-                                            }
-                                            Err(e) => py_runtime_err(e),
-                                        }
+                            let schema = input_plans[0].schema();
+                            let (schema_name_opt, column_name) =
+                                self._schema_column_from_fqn(&fq_name);
+                            match schema_name_opt {
+                                Some(e) => {
+                                    match schema
+                                        .index_of_column_by_name(Some(e.as_str()), column_name)
+                                    {
+                                        Ok(e) => Ok(e),
+                                        Err(e) => Err(py_runtime_err(e)),
                                     }
-                                    None => {
-                                        match schema.index_of_column_by_name(None, column_name) {
-                                            Ok(e) => {
-                                                idx = e;
-                                                break;
-                                            }
-                                            Err(e) => py_runtime_err(e),
-                                        }
-                                    }
-                                };
+                                }
+                                None => match schema.index_of_column_by_name(None, column_name) {
+                                    Ok(e) => Ok(e),
+                                    Err(e) => Err(py_runtime_err(e)),
+                                },
                             }
-                            Ok(idx)
                         }
                         Err(e) => Err(py_runtime_err(e)),
                     }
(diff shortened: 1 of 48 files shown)


jdye64:
@andygrove am I doing anything dense here?

dask_planner/src/expression.rs
```diff
@@ -146,14 +160,30 @@ impl PyExpr {
                         Ok(fq_name) => {
                             let mut idx: usize = 0;
                             for schema in input_plans[0].all_schemas() {
```
andygrove, line 162: I have a high level question about the logic here. It looks like we are iterating over multiple schemas but if we don't find a match on the first schema we will panic, so it looks like it can never look in more than one schema?

jdye64: indeed, that is not good. Let me adjust that

andygrove: I also don't think we should be using `all_schemas` here. I think we should just be looking at `input_plans[0].schema()`. I am not too familiar with `all_schema`, but for a join, it would return both the left scheme and the right schema for its inputs, so if we get an index into the right schema, that wouldn't make sense when applied to the output of the plan.

dask_planner/src/expression.rs
```diff
@@ -146,14 +160,30 @@ impl PyExpr {
...
+                                match schema_name_opt {
+                                    Some(e) => {
+                                        match schema
+                                            .index_of_column_by_name(Some(e.as_str()), column_name)
+                                        {
+                                            Ok(e) => {
+                                                idx = e;
+                                                break;
+                                            }
+                                            Err(_e) => panic!("Didn't find in index ..."),
```
andygrove, line 174:
```suggestion
                                            Err(e) => py_runtime_err(e),
```

dask_planner/src/expression.rs
```diff
@@ -146,14 +160,30 @@ impl PyExpr {
...
                                     }
-                                    Err(_e) => (),
-                                }
+                                    None => {
+                                        match schema.index_of_column_by_name(None, column_name) {
+                                            Ok(e) => {
+                                                idx = e;
+                                                break;
+                                            }
+                                            Err(_e) => panic!("Didn't find in index ..."),
```
andygrove, line 183:
```suggestion
                                            Err(e) => py_runtime_err(e),
```

dask_planner/src/expression.rs
jdye64: So I'm not sure why, however when I remove `all_schema` to use `input_plans[0].schema()` a lot of test start failing with field not found errors from DataFusion? Could it be possible that maybe we do need `all_schemas()` like you that doesn't quite make sense to me but clearly something in there is needed.

jdye64: I'll make a commit with that change just so the changes and errors can be vizualized.

andygrove: I'm getting test failures locally before the commit that removes `all_schema`. Not sure if there is something up with my local env.

dask_planner/src/expression.rs
```diff
@@ -146,13 +179,30 @@ impl PyExpr {
...
+                                    {
+                                        Ok(e) => {
+                                            idx = e;
+                                            break;
+                                        }
+                                        Err(_e) => (),
+                                    },
                                 }
                             }
                             Ok(idx)
```
andygrove, line 208: If no match is found, then we return `Ok(0)` here, which I assume is not intended.

dask_planner/src/expression.rs
```diff
@@ -146,13 +179,30 @@ impl PyExpr {
...
+                                match schema_name_opt {
+                                    Some(e) => {
+                                        match schema.index_of_column_by_name(
+                                            Some(e.as_str()),
+                                            column_name.as_str(),
+                                        ) {
+                                            Ok(e) => {
+                                                idx = e;
+                                                break;
+                                            }
```
andygrove, line 193: We can simplify the flow and just return here:

```suggestion
                                            Ok(idx) => return Ok(idx),
```

dask_planner/src/expression.rs
```diff
@@ -146,13 +179,30 @@ impl PyExpr {
...
+                                        }
                                     }
-                                    Err(_e) => (),
+                                    None => match schema
+                                        .index_of_column(&Column::from_qualified_name(&fq_name))
+                                    {
+                                        Ok(e) => {
+                                            idx = e;
+                                            break;
+                                        }
```
andygrove, line 203:
```suggestion
                                        Ok(idx) => return Ok(idx),
```

andygrove:
@jdye64 The code we are building on already contains a serious bug, so I think we need to step back and fix that first. Here is my proposed fix: https://github.com/dask-contrib/dask-sql/pull/732

dask_planner/src/expression.rs
jdye64: Resolved in https://github.com/dask-contrib/dask-sql/pull/732

dask_planner/src/expression.rs
jdye64: Resolved in https://github.com/dask-contrib/dask-sql/pull/732

dask_planner/src/expression.rs
jdye64: Resolved in https://github.com/dask-contrib/dask-sql/pull/732

jdye64:
This PR is no longer needed due to https://github.com/dask-contrib/dask-sql/pull/732

PR #727, opened 2022-08-26
{CREATE | USE | DROP} Schema support

jdye64:


dask_planner/src/expression.rs
```diff
@@ -146,14 +160,36 @@ impl PyExpr {
...
+                                            Ok(e) => {
+                                                idx = e;
+                                                break;
+                                            }
+                                            Err(_e) => {
+                                                println!(
+                                                    "Schema_name_opt: {:?} - Column_Name: {:?}",
+                                                    schema_name_opt, column_name
+                                                );
+                                                panic!("Didn't find in index ...")
```
charlesbluca, line 179: Since this is accessible from Python, should we return a Python error here instead of panicking?

dask_planner/src/expression.rs
```diff
@@ -146,14 +160,36 @@ impl PyExpr {
...
                                     }
-                                    Err(_e) => (),
-                                }
+                                    None => {
+                                        match schema.index_of_column_by_name(None, column_name) {
+                                            Ok(e) => {
+                                                idx = e;
+                                                break;
+                                            }
+                                            Err(_e) => panic!("Didn't find in index ..."),
```
charlesbluca, line 189: Same question as above

dask_planner/src/parser.rs
```diff
@@ -255,37 +294,43 @@ impl<'a> DaskParser<'a> {
...
+                        ]);
+
+                        // use custom parsing
+                        self.parse_create_schema(if_not_exists, or_replace)
+                    }
                     "table" => {
                         // move one token forward
                         self.parser.next_token();
                         // use custom parsing
                         self.parse_create_table(or_replace)
```
charlesbluca, line 331: For sake of consistency, think we could modify `parse_create_table` to accept `if_not_exists` so this could match the other cases?

```suggestion
                        self.parser.next_token();

                        let if_not_exists = self.parser.parse_keywords(&[
                            Keyword::IF,
                            Keyword::NOT,
                            Keyword::EXISTS,
                        ]);

                        // use custom parsing
                        self.parse_create_table(if_not_exists, or_replace)
```

dask_planner/src/parser.rs
```diff
@@ -255,37 +294,43 @@ impl<'a> DaskParser<'a> {
...
-                            // Go back three tokens if IF NOT EXISTS was consumed
-                            self.parser.prev_token();
-                            self.parser.prev_token();
-                            self.parser.prev_token();
-                        }
-                        if or_replace {
-                            // Go back two tokens if OR REPLACE was consumed
-                            self.parser.prev_token();
-                            self.parser.prev_token();
-                        }
```
charlesbluca, line 288: Unless I'm missing something, don't we need to keep this block since we are parsing for `OR REPLACE` before we enter the match?

If we follow the above review comment, maybe we could move the `OR REPLACE` parsing to each match case, and wrap it up along with the `IF NOT EXISTS` parsing in a utility function.

dask_planner/src/sql/logical/create_catalog_schema.rs
```diff
@@ -0,0 +1,113 @@
...
+
+    #[pyo3(name = "getIfNotExists")]
+    fn get_if_not_exists(&self) -> PyResult<bool> {
+        Ok(self.create_catalog_schema.if_not_exists)
+    }
+
+    #[pyo3(name = "getReplace")]
+    fn get_replace(&self) -> PyResult<bool> {
+        Ok(self.create_catalog_schema.or_replace)
+    }
```
charlesbluca, line 89: Nitpicky but do you prefer `getOrReplace` or `getReplace` for this Python-facing function (along with its other appearance in the drop and switch schema code)? Asking because I ended up going with `getOrReplace` for `CREATE TABLE AS | WITH`, and I feel like it makes sense to have some consistency here

jdye64:
@charlesbluca thank you for the review, all very valid things. Working on some upstream things that might alter a few of these already. Once I merge that into this branch I will address the remaining issues.

dask_planner/src/parser.rs
```diff
@@ -333,6 +400,21 @@ impl<'a> DaskParser<'a> {
...
+                        // move one token forward
+                        self.parser.next_token();
+                        // use custom parsing
+
+                        let if_exists = self.parser.parse_keywords(&[Keyword::IF, Keyword::EXISTS]);
+
+                        let schema_name = self.parser.parse_identifier();
+
+                        let drop_schema = DropSchema {
+                            schema_name: schema_name?.value,
```
andygrove, line 413: minor nit but I think it is better to have the `?` after the function call

```suggestion
                        let schema_name = self.parser.parse_identifier()?;

                        let drop_schema = DropSchema {
                            schema_name: schema_name.value,
```

dask_planner/src/parser.rs
```diff
@@ -410,6 +492,33 @@ impl<'a> DaskParser<'a> {
...
+            Token::Word(w) => {
+                match w.value.to_lowercase().as_str() {
+                    "schema" => {
+                        // move one token forward
+                        self.parser.next_token();
+                        // use custom parsing
+                        let schema_name = self.parser.parse_identifier();
+
+                        let use_schema = UseSchema {
+                            schema_name: schema_name?.value,
```
andygrove, line 507:
```suggestion
                        let schema_name = self.parser.parse_identifier()?;

                        let use_schema = UseSchema {
                            schema_name: schema_name.value,
```

dask_planner/src/sql/logical/create_catalog_schema.rs
```diff
@@ -0,0 +1,113 @@
...
+
+#[pyclass(name = "CreateCatalogSchema", module = "dask_planner", subclass)]
+pub struct PyCreateCatalogSchema {
+    pub(crate) create_catalog_schema: CreateCatalogSchemaPlanNode,
+}
+
+#[pymethods]
+impl PyCreateCatalogSchema {
+    #[pyo3(name = "getSchemaName")]
+    fn get_table_name(&self) -> PyResult<String> {
```
andygrove, line 77: Would it be better for the Rust method to be named `get_schema_name` here?

dev-tests/report.html
```diff
@@ -0,0 +1,522 @@
+<!DOCTYPE html>
```
andygrove, line 1: Not sure if this was intentionally added?

andygrove (approved):
LGTM. Left a couple of code style comments.

dev-tests/report.html
jdye64: dang, keep accidentally getting this file after running pytests report. I'm gonna add it to .gitignore

PR #733, opened 2022-08-30
[DF] Implement `ANALYZE TABLE`

charlesbluca:
Still some additional tasks to be done:

- [ ] Add support for specific columns instead of `ALL COLUMNS`
- [ ] Open sqlparser issue around potentially unexpected behavior in `parse_table_factor`

### DIFF
--- a/dask_planner/src/parser.rs
+++ b/dask_planner/src/parser.rs
@@ -5,7 +5,7 @@
 use crate::dialect::DaskDialect;
 use crate::sql::parser_utils::DaskParserUtils;
 use datafusion_sql::sqlparser::{
-    ast::{Expr, Statement as SQLStatement},
+    ast::{Expr, SelectItem, Statement as SQLStatement},
     dialect::{keywords::Keyword, Dialect},
     parser::{Parser, ParserError},
     tokenizer::{Token, Tokenizer},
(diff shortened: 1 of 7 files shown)


dask_planner/src/parser.rs
```diff
@@ -577,4 +620,33 @@ impl<'a> DaskParser<'a> {
             },
         })))
     }
+
+    /// Parse Dask-SQL ANALYZE TABLE <table>
+    fn parse_analyze_table(&mut self) -> Result<DaskStatement, ParserError> {
+        let table_factor = self.parser.parse_table_factor()?;
+        // parse_table_factor parses the following keyword as an alias, so we need to go back a token
+        // TODO: open an issue in sqlparser around this when possible
+        self.parser.prev_token();
```
charlesbluca, line 629: @jdye64 and I noticed that for a statement like `ANALYZE TABLE df COMPUTE STATISTICS FOR ALL COLUMNS`, sqlparser consumes and interprets `COMPUTE` as an alias for `df` even though there is no `AS` keyword. Not sure if this is expected behavior for `parse_table_factor`, so I added this comment in case we want to open up an issue around that later on.

cc @andygrove

dask_sql/physical/rel/custom/analyze.py
```diff
@@ -54,7 +56,9 @@ def convert(
         statistics = statistics.append(
             pd.Series(
                 {
-                    col: str(python_to_sql_type(df[mapping(col)].dtype)).lower()
+                    col: str(python_to_sql_type(df[mapping(col)].dtype).getSqlType())
+                    .rpartition(".")[2]
+                    .lower()
```
charlesbluca, line 61: Is there any way to get the string name of a `SqlTypeName` variant? Right now this code is used in a lot of places to turn something like `datafusion.SqlTypeName.DOUBLE` into `'DOUBLE'` and am wondering if there's a cleaner alternative.

dask_planner/src/parser.rs
andygrove: This does seem like a bug since `COMPUTE` is defined as a keyword in sqlparser-rs.

andygrove: Also, ANALYZE TABLE should really just be looking for an identifier for the table name and not call `parse_table_factor`.

dask_sql/physical/rel/custom/analyze.py
ayushdg: we can probably override the `__str__` method that's autp generated by pyo3 in the bindings to just return the name instead.

jdye64 (approved):
Looks good to me. I think there are lots of things we are learning as we go with the parsing and all of the part of the code base will be due for a refactor once we analyze all the elements we need to implement.

PR #738, opened 2022-08-31
Test function `test_aggregate_function`

sarahyurick:
In `test_function.py`, the only remaining test is `test_aggregate_function()`. This PR allows this test to pass by introducing the `aggregation` bool to the `DaskFunction` struct (following a similar convention to `FunctionDescription`) and implementing the `add_or_overload_aggregation` function.

Builds upon changes from #698

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -165,6 +165,7 @@ impl ContextProvider for DaskSQLContext {
                     if function.aggregation.eq(&true) {
                         return None;
                     }
+                    let sig = {
                         Signature::one_of(
                             function
                                 .return_types
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -16,8 +16,9 @@ use arrow::datatypes::{DataType, Field, Schema};
 use datafusion_common::{DFSchema, DataFusionError};
 use datafusion_expr::logical_plan::Extension;
 use datafusion_expr::{
-    AggregateUDF, LogicalPlan, PlanVisitor, ReturnTypeFunction, ScalarFunctionImplementation,
-    ScalarUDF, Signature, TableSource, TypeSignature, Volatility,
+    AccumulatorFunctionImplementation, AggregateUDF, LogicalPlan, PlanVisitor, ReturnTypeFunction,
+    ScalarFunctionImplementation, ScalarUDF, Signature, StateTypeFunction, TableSource,
+    TypeSignature, Volatility,
 };
 use datafusion_sql::{
     parser::Statement as DFStatement,
@@ -160,6 +161,9 @@ impl ContextProvider for DaskSQLContext {
         for schema in self.schemas.values() {
             for (fun_name, function) in &schema.functions {
                 if fun_name.eq(name) {
+                    if function.lock().unwrap().aggregation.eq(&true) {
+                        return None;
+                    }
                     let sig = {
                         let function = function.lock().unwrap();
                         Signature::one_of(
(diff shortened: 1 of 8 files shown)


dask_sql/context.py
```diff
@@ -752,12 +752,11 @@ def _prepare_schemas(self):
...
-                    # rust_schema.add_or_overload_aggregation(
-                    #     name,
-                    #     [param[1].getDataType() for param in sql_parameters],
-                    #     sql_return_type.getDataType(),
-                    # )
+                    rust_schema.add_or_overload_aggregation(
+                        name,
+                        [param[1].getDataType() for param in sql_parameters],
+                        sql_return_type.getDataType(),
+                    )
```
charlesbluca, line 759: I realize that I'm the one who left the suggestion to implement `add_or_overload_aggregation` 😅 but since at this point we've validated that there should be no UDF/UDAFs sharing the same name, would it make sense to consolidate this logic into a single function that also takes an `aggregation` boolean?

dask_planner/src/sql.rs
```diff
@@ -195,7 +199,73 @@ impl ContextProvider for DaskSQLContext {
...
+                            Volatility::Immutable,
+                        )
+                    };
+                    let function = function.clone();
+                    let rtf: ReturnTypeFunction = Arc::new(move |input_types| {
+                        let function = function.lock().unwrap();
+                        match function.return_types.get(&input_types.to_vec()) {
+                            Some(return_type) => Ok(Arc::new(return_type.clone())),
+                            None => Err(DataFusionError::Plan(format!(
+                                "UDF signature not found for input types {:?}",
```
charlesbluca, line 253:
```suggestion
                                "UDAF signature not found for input types {:?}",
```

dask_sql/context.py
sarahyurick: Yeah that sounds good to me! I'll make those changes now.

dask_planner/src/sql.rs
```diff
@@ -160,6 +161,9 @@ impl ContextProvider for DaskSQLContext {
         for schema in self.schemas.values() {
             for (fun_name, function) in &schema.functions {
                 if fun_name.eq(name) {
+                    if function.lock().unwrap().aggregation.eq(&true) {
+                        return None;
+                    }
```
jdye64, line 166: This mutex needs to be manually `unlocked` since the RAII guard will still be in scope at this point. https://doc.rust-lang.org/std/sync/struct.Mutex.html#method.lock I think it best to refactor this snippet like this

jdye64, line 166:
```suggestion
                    let mut guard = mutex.lock().unwrap();
                    if guard.unwrap().aggregation.eq(&true) {
                        return None;
                    }
```

dask_planner/src/sql.rs
```diff
@@ -160,6 +161,9 @@ impl ContextProvider for DaskSQLContext {
...
                     let sig = {
                         let function = function.lock().unwrap();
```
jdye64, line 168:
```suggestion
                        let function = guard.unwrap();
```

dask_planner/src/sql.rs
```diff
@@ -195,7 +199,67 @@ impl ContextProvider for DaskSQLContext {
         None
```
jdye64, line 199: No need to unlock the mutex since Rust guarantees once a RAII mutex guard variable goes out of scope the mutex will be unlocked by Rust itself. That will happen here in return since the `guard` mutex variable we made will no longer be in scope

dask_planner/src/sql.rs
```diff
@@ -195,7 +199,67 @@ impl ContextProvider for DaskSQLContext {
...
     }
 
-    fn get_aggregate_meta(&self, _name: &str) -> Option<Arc<AggregateUDF>> {
+    fn get_aggregate_meta(&self, name: &str) -> Option<Arc<AggregateUDF>> {
+        let acc: AccumulatorFunctionImplementation =
+            Arc::new(|| Err(DataFusionError::NotImplemented("".to_string())));
```
jdye64, line 204:
```suggestion
            Arc::new(|_| Err(DataFusionError::NotImplemented("".to_string())));
```

dask_planner/src/sql.rs
```diff
@@ -195,7 +199,67 @@ impl ContextProvider for DaskSQLContext {
...
+                return Some(Arc::new(AggregateUDF::new(name, &sig, &rtf, &acc, &st)));
+            }
+            _ => (),
+        }
+
+        // Loop through all of the user defined functions
+        for schema in self.schemas.values() {
+            for (fun_name, function) in &schema.functions {
+                if fun_name.eq(name) {
+                    if function.lock().unwrap().aggregation.eq(&false) {
```
jdye64, line 233: Same refactor as mutex variable suggestion above

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -327,9 +327,10 @@ def _collect_aggregations(
...
-            assert (
-                expr.getExprType() == "AggregateFunction"
-            ), "Do not know how to handle this case!"
+            assert expr.getExprType() in {
+                "AggregateFunction",
+                "AggregateUDF",
+            }, "Do not know how to handle this case!"
 
             # TODO: Generally we need a way to capture the current SQL schema here in case this is a custom aggregation function
             schema_name = "root"
```
jdye64, line 336: This is my fault. I am the one who originally made this naughty bit. Figure we can change it now before we forget to however

```suggestion
            schema_name = context.schema_name
```

jdye64 (changes requested):
Really cool Sarah! Nice to have aggregate udfs. Had a few small suggestions.

dask_planner/src/sql.rs
```diff
@@ -160,6 +161,9 @@ impl ContextProvider for DaskSQLContext {
         for schema in self.schemas.values() {
             for (fun_name, function) in &schema.functions {
```
jdye64, line 162:
```suggestion
            for (fun_name, func_mutex) in &schema.functions {
```

dask_planner/src/sql.rs
```diff
@@ -160,6 +161,9 @@ impl ContextProvider for DaskSQLContext {
...
                 if fun_name.eq(name) {
+                    if function.lock().unwrap().aggregation.eq(&true) {
+                        return None;
+                    }
                     let sig = {
```
jdye64, line 167:
```suggestion
                    let function = func_mutex.lock().unwrap();
                    if function.aggregation.eq(&true) {
                        return None;
                    }
```

dask_planner/src/sql.rs
jdye64, line 168:
```suggestion
```

dask_planner/src/sql.rs
```diff
@@ -195,7 +199,67 @@ impl ContextProvider for DaskSQLContext {
...
+                let rtf: ReturnTypeFunction =
+                    Arc::new(|input_types| Ok(Arc::new(input_types[0].clone())));
+                return Some(Arc::new(AggregateUDF::new(name, &sig, &rtf, &acc, &st)));
+            }
+            _ => (),
+        }
+
+        // Loop through all of the user defined functions
+        for schema in self.schemas.values() {
+            for (fun_name, function) in &schema.functions {
```
jdye64, line 231:
```suggestion
            for (fun_name, func_mutex) in &schema.functions {
```

dask_planner/src/sql.rs
```diff
@@ -195,7 +199,67 @@ impl ContextProvider for DaskSQLContext {
...
+                if fun_name.eq(name) {
+                    if function.lock().unwrap().aggregation.eq(&false) {
+                        return None;
+                    }
```
jdye64, line 235:
```suggestion
                    let function = func_mutex.lock().unwrap();
                    if function.aggregation.eq(&false) {
                        return None;
                    }
```

dask_planner/src/sql.rs
```diff
@@ -195,7 +199,67 @@ impl ContextProvider for DaskSQLContext {
...
+                    let sig = {
+                        let function = function.lock().unwrap();
```
jdye64, line 237:
```suggestion
```

dask_planner/src/sql.rs
```diff
@@ -195,7 +199,67 @@ impl ContextProvider for DaskSQLContext {
...
+                                .return_types
+                                .keys()
+                                .map(|v| TypeSignature::Exact(v.to_vec()))
+                                .collect(),
+                            Volatility::Immutable,
+                        )
+                    };
+                    let function = function.clone();
+                    let rtf: ReturnTypeFunction = Arc::new(move |input_types| {
+                        let function = function.lock().unwrap();
```
jdye64, line 249:
```suggestion
```

sarahyurick:
Thanks @jdye64 !

dask_planner/src/sql.rs
sarahyurick: I got `Arc::new(|| Err(DataFusionError::NotImplemented("".to_string())));` from #698 and it looks like it expects 0 arguments, so I'm going to revert this change.

jdye64 (approved):
LGTM

PR #741, opened 2022-09-01
test_groupby's `test_group_by_all()`

sarahyurick:
Here is a very simple solution to make something like `SUM(2) AS "X"` work. Essentially, we just create a new column called "2" that contains only 2s and then call the aggregation as usual.

I guess one thing to consider is the likelihood that the user already would have a column called "2", etc., since in that case the existing column might be overwritten?

I plan to work on the `SUM(a+b) AS mix_2` and `AVG(a+b) AS mix_3` portions of the test next.

### DIFF
--- a/dask_sql/datacontainer.py
+++ b/dask_sql/datacontainer.py
@@ -161,7 +161,11 @@ class ColumnContainer:
             # Make sure to get the backend column name here by using [1] instead of [0]. [0] is frontend
             backend_column = list(self._frontend_backend_mapping.items())[0][1]
         else:
-            backend_column = self._frontend_backend_mapping[column]
+            try:
+                backend_column = self._frontend_backend_mapping[column]
+            except KeyError:
+                backend_column = column[column.find("(")+1:column.find(")")]
+
         return backend_column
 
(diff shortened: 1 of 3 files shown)


dask_sql/datacontainer.py
```diff
@@ -161,7 +161,11 @@ def get_backend_by_frontend_name(self, column: str) -> str:
             # Make sure to get the backend column name here by using [1] instead of [0]. [0] is frontend
             backend_column = list(self._frontend_backend_mapping.items())[0][1]
         else:
-            backend_column = self._frontend_backend_mapping[column]
+            try:
+                backend_column = self._frontend_backend_mapping[column]
+            except KeyError:
+                backend_column = column[column.find("(")+1:column.find(")")]
```
sarahyurick, line 167: For the `SUM(2) AS "X"` example, `column` is "Int64(2)", so this line of code is extracting 2 from the parentheses.

ayushdg: cc: @charlesbluca since some of the things you are looking into may/may not handle this case

PR #742, opened 2022-09-02
[DF] DataFusion upgrade

andygrove:
Upgrade to new DataFusion. Key changes:

- `Expr.name()` no longer requires a schema
- New `Expr` variants for `IsTrue`, `IsFalse`, etc.
- An optimizer rule was renamed


jdye64 (approved):
Thanks Andy!

PR #744, opened 2022-09-06
Upgrade to latest DataFusion

andygrove:


dask_planner/src/sql/logical.rs
```diff
@@ -281,6 +281,7 @@ impl PyLogicalPlan {
             LogicalPlan::CreateExternalTable(_create_external_table) => "CreateExternalTable",
             LogicalPlan::CreateMemoryTable(_create_memory_table) => "CreateMemoryTable",
             LogicalPlan::DropTable(_drop_table) => "DropTable",
+            LogicalPlan::DropView(_drop_view) => "DropView",
```
charlesbluca, line 284: Do we make use of `DROP VIEW` anywhere? From my understanding, tables and views are both stored as tables and would be removed with `DROP TABLE`.

I'm okay with updating this in the future though if it makes sense.

andygrove:
I pushed another commit now that the type coercion rule has been merged into DataFusion, so we can delete our copy.

dask_planner/src/sql/logical.rs
andygrove: `DROP VIEW` was just recently added in DataFusion, so we are definitely not using this yet.

charlesbluca (approved):

PR #747, opened 2022-09-07
[DF] Generalize `CREATE | PREDICT MODEL` to accept non-native `SELECT` statements

charlesbluca:
Generalizes `CREATE | PREDICT MODEL` to accept `DaskStatements` instead of `SQLStatements`, so that we can do things like:

```sql
SELECT * FROM PREDICT (
    MODEL my_model,
    SELECT x, y FROM PREDICT (
        ...
    )
)
```

This is something that would've happened in working on `FILTER WHERE` support, but figured it might make sense to break this out into its own PR.

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -381,9 +381,7 @@ impl DaskSQLContext {
             DaskStatement::CreateModel(create_model) => Ok(LogicalPlan::Extension(Extension {
                 node: Arc::new(CreateModelPlanNode {
                     model_name: create_model.name,
-                    input: self._logical_relational_algebra(DaskStatement::Statement(Box::new(
-                        create_model.select,
-                    )))?,
+                    input: self._logical_relational_algebra(create_model.select)?,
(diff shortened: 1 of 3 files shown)


dask_planner/src/sql.rs
```diff
@@ -381,9 +381,7 @@ impl DaskSQLContext {
             DaskStatement::CreateModel(create_model) => Ok(LogicalPlan::Extension(Extension {
                 node: Arc::new(CreateModelPlanNode {
                     model_name: create_model.name,
-                    input: self._logical_relational_algebra(DaskStatement::Statement(Box::new(
-                        create_model.select,
-                    )))?,
+                    input: self._logical_relational_algebra(create_model.select)?,
```
charlesbluca, line 384: Ideally, we should be doing a match on `create_model.select` to make sure it's a relevant variant - the main ones I had in mind were `Statements` and `PredictModels`, but think it could realistically be any variant that outputs a table?

ayushdg (approved):

galipremsagar (approved):

PR #716, opened 2022-08-24
Optimize away distinct aggregate operations - eliminate_agg_distinct

jdye64:

Issue #725, opened 2022-08-26
[DF] Incorrect results for select count(distinct(col)) in datafusion-sql-plannet


randerzander:
Dask-SQL version: 2022.8.0+204.gca504f4

Both cudf and pandas DFs get incorrect count(distinct(col)) results:
```
>>> from dask_sql import Context
>>> import cudf

>>> c = Context()
>>> df = cudf.DataFrame({"id": [0, 1, 2, 2]})
>>> c.create_table("df", df)
>>> c.sql("select count(distinct(id)) from df").compute()
   COUNT(DISTINCT df.id)
0                      0

>>> import pandas as pd
>>> df = pd.DataFrame({"id": [0, 1, 2, 2]})
>>> c.create_table("df", df)
>>> c.sql("select count(distinct(id)) from df").compute()
   COUNT(DISTINCT df.id)
0                      4
```

charlesbluca:
Believe this is the same issue @sarahyurick brought up in #532, and should be resolved with #716

sarahyurick:
Should this also be linked to #759?

randerzander:
> Should this also be linked to #759?

Yes, this is now resolved :)

PR #748, opened 2022-09-08
[DF] Optimize away `COUNT DISTINCT` aggregate operations - eliminate_agg_distinct

andygrove:
This is a continuation of https://github.com/dask-contrib/dask-sql/pull/716

This PR only handles the simple case of `SELECT COUNT(DISTINCT expr)` and there is a follow on PR https://github.com/dask-contrib/dask-sql/pull/759 to handle more complex cases


tests/integration/test_compatibility.py
```diff
@@ -158,6 +158,29 @@ def test_order_by_no_limit():
     )
 
 
+@pytest.mark.skip(reason="WIP DataFusion")
```
andygrove, line 161: Jeremy had added this test but is not complete. The SQL references tables that do not exists. I am not sure if we should just remove this test from the PR?

dask_planner/src/sql/logical/aggregate.rs
```diff
@@ -67,8 +68,21 @@ impl PyAggregate {
...
+    #[pyo3(name = "isAggExprDistinct")]
+    pub fn distinct_agg_expr(&self, expr: PyExpr) -> PyResult<bool> {
+        Ok(match expr.expr {
+            Expr::AggregateFunction {
+                fun: _,
+                args: _,
+                distinct,
+                filter: _
+            } => distinct,
+            _ => panic!("Encountered a non Aggregate type in agg_func_name"),
```
ayushdg, line 80: We could raise a py_type_err instead of panic here.

tests/integration/test_compatibility.py
ayushdg: Yeah we should be good to remove this test and just verify that https://github.com/dask-contrib/dask-sql/issues/725 is working as expected

ayushdg (approved):
Minor suggestions but lgtm! Thanks

dask_planner/src/sql/logical/aggregate.rs
andygrove: Fixed

PR #751, opened 2022-09-09
[DF] Update DataFusion to pick up SQL support for LIKE, ILIKE, SIMILAR TO with escape char

andygrove:
There is one test failure after this update:

```
FAILED tests/integration/test_sqlite.py::test_filter - AssertionError: DataFrame are different
```

The reason for this is that the representation of `Like` and `NotLike` has changed in the logical plan and the corresponding Python code now needs updating.

Previously these were modeled as `Expr::Binary` with `Operator::Like/NotLike` but now they are represented as a new top-level `Expr::Like` expression instead.

### DIFF
--- a/tests/integration/test_rex.py
+++ b/tests/integration/test_rex.py
@@ -239,6 +239,8 @@ def test_like(c, input_table, gpu, request):
 
     assert len(df) == 0
 
+    # TODO: uncomment when sqlparser adds parsing support for non-standard escape characters
+    # https://github.com/dask-contrib/dask-sql/issues/754
     # df = c.sql(
     #     f"""
     #     SELECT * FROM {input_table}
--- a/tests/integration/test_rex.py
+++ b/tests/integration/test_rex.py
@@ -202,9 +202,6 @@ def test_operators(c, df):
     assert_eq(result_df, expected_df)
 
 
-@pytest.mark.skip(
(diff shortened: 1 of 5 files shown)


andygrove:
@ayushdg Could you help me out with the Python changes for this PR?

ayushdg:
> @ayushdg Could you help me out with the Python changes for this PR?

Yeah I can take a look

ayushdg (approved):
Changes lgtm! @charlesbluca Could you take a look if you get the chance?

tests/integration/test_rex.py
```diff
@@ -242,14 +239,14 @@ def test_like(c, input_table, gpu, request):
...
-    )
+    # df = c.sql(
+    #     f"""
+    #     SELECT * FROM {input_table}
+    #     WHERE a LIKE 'Ä%Ä_Ä%' ESCAPE 'Ä'
+    # """
+    # )
 
-    assert_eq(df, string_table.iloc[[1]])
+    # assert_eq(df, string_table.iloc[[1]])
```
charlesbluca, line 249: Do we have an associated issue / PR open to address this case? If so, could we add a link to that here?

ayushdg: Will open an issue and link here

charlesbluca (approved):
LGTM!

PR #755, opened 2022-09-13
Upgrade to Datafusion 12.0.0 RC1

andygrove:


dask_planner/src/expression.rs
```diff
@@ -421,19 +421,21 @@ impl PyExpr {
                 | Operator::RegexMatch
                 | Operator::RegexIMatch
                 | Operator::RegexNotMatch
-                | Operator::RegexNotIMatch
-                | Operator::BitwiseAnd
-                | Operator::BitwiseOr => "BOOLEAN",
```
andygrove, line 426: I was adding the new `BitwiseXor` operator when I noticed that this logic was incorrect so I combined handling of all bitwise operators to return an error,

charlesbluca (approved):

PR #758, opened 2022-09-13
[REVIEW] Resolve `test_table_schema`

sarahyurick:
It looks like #727 implemented schema support. This PR should resolve another schema test.

So far, I've made small changes so that we can do a basic `"SELECT * FROM root.df"`. The latter portion of the test, which I'm still working on, consists of creating tables within and across schemas.

### DIFF
--- a/dask_sql/physical/rel/custom/create_table_as.py
+++ b/dask_sql/physical/rel/custom/create_table_as.py
@@ -39,6 +39,10 @@ class CreateTableAsPlugin(BaseRelPlugin):
         create_memory_table = rel.create_memory_table()
 
         schema_name, table_name = context.schema_name, create_memory_table.getName()
+        if "." in table_name:
+            raise KeyError(
+                f"Cannot create table from {table_name} due to schema error."
+            )
 
         if table_name in context.schema[schema_name].tables:
             if create_memory_table.getIfNotExists():
--- a/tests/integration/test_schema.py
+++ b/tests/integration/test_schema.py
@@ -26,7 +26,7 @@ def test_table_schema(c, df):
     with pytest.raises(KeyError):
         # TODO: https://github.com/dask-contrib/dask-sql/issues/771
         # c.sql("CREATE TABLE other.bar AS TABLE df")
-        c.sql("CREATE TABLE other.bar AS (SELECT * FROM df)")
+        c.sql("CREATE TABLE other.bar AS (SELECT * FROM root.df)")
 
     c.sql('USE SCHEMA "root"')
     assert_eq(original_df, c.sql("SELECT * FROM foo.bar"))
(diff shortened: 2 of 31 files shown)


tests/integration/test_schema.py
```diff
@@ -32,7 +34,7 @@ def test_table_schema(c, df):
 
     c.sql("DROP SCHEMA foo")
 
-    with pytest.raises(ParsingException):
+    with pytest.raises(KeyError):
         c.sql("SELECT * FROM foo.bar")
```
sarahyurick, line 38: This raises a `KeyError: 'foo'` instead of a `ParsingException: Plan("Table 'dask_sql.foo.bar' not found")`. I could add extra logic to make it match the other ParsingException if desired, but for now just let it throw the KeyError.

dask_sql/physical/rel/custom/create_table_as.py
```diff
@@ -39,6 +39,10 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
         create_memory_table = rel.create_memory_table()
 
         schema_name, table_name = context.schema_name, create_memory_table.getName()
+        if "." in table_name:
+            raise KeyError(
+                f"Cannot create table from {table_name} due to schema error."
+            )
```
sarahyurick, line 45: This logic basically makes it impossible to create a table when the schema is specified.

Without it, if we were using schema `foo` and tried to create table `other.bar` (regardless of whether schema `other` exists), `other.bar` is added to `foo`, but then it is impossible to access it (e.g., with `select`) because we cannot parse `foo.other.bar`.

Since this test doesn't cover `create table [schema].[name] ...`, I just made it error regardless of whether or not a valid schema is specified.

tests/integration/test_schema.py
```diff
@@ -18,11 +17,16 @@ def test_table_schema(c, df):
...
+    # For now, we create table bar another (equivalent) way
+    # c.sql("CREATE TABLE bar AS TABLE root.df")
+    c.sql("CREATE TABLE bar AS (SELECT * FROM root.df)")
     assert_eq(original_df, c.sql("SELECT * FROM bar"))
 
     with pytest.raises(KeyError):
-        c.sql("CREATE TABLE other.bar AS TABLE df")
+        # TODO: https://github.com/dask-contrib/dask-sql/issues/771
+        # c.sql("CREATE TABLE other.bar AS TABLE df")
+        c.sql("CREATE TABLE other.bar AS (SELECT * FROM df)")
```
sarahyurick, line 29: This currently errors because there is no `foo.df`, and the nested statement is parsed before we look at `other.bar`. I could fix this by changing it to `(SELECT * FROM root.df)`

PR #759, opened 2022-09-13
[DF] Support complex queries with multiple DISTINCT aggregates

andygrove:
Improves support for `COUNT(DISTINCT expr)` queries:

- Supports GROUP BY
- Supports COUNT DISTINCT on multiple expressions

### DIFF
--- a/dask_planner/Cargo.toml
+++ b/dask_planner/Cargo.toml
@@ -9,19 +9,20 @@ edition = "2021"
 rust-version = "1.62"
 
 [dependencies]
-tokio = { version = "1.0", features = ["macros", "rt", "rt-multi-thread", "sync", "fs", "parking_lot"] }
-rand = "0.7"
-pyo3 = { version = "0.16.5", features = ["extension-module", "abi3", "abi3-py38"] }
 arrow = { version = "22.0.0", features = ["prettyprint"] }
-datafusion-sql = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-expr = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-common = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-optimizer = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-uuid = { version = "0.8", features = ["v4"] }
+async-trait = "0.1.41"
+datafusion-common = { git = "https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
+datafusion-expr = { git = "https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
+datafusion-optimizer = { git = "https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
+datafusion-sql = { git = "https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
+env_logger = "0.9"
+log = "^0.4"
 mimalloc = { version = "*", default-features = false }
 parking_lot = "0.12"
-async-trait = "0.1.41"
-
+pyo3 = { version = "0.16.5", features = ["extension-module", "abi3", "abi3-py38"] }
+rand = "0.7"
+tokio = { version = "1.0", features = ["macros", "rt", "rt-multi-thread", "sync", "fs", "parking_lot"] }
+uuid = { version = "0.8", features = ["v4"] }
 
 [lib]
 crate-type = ["cdylib"]
--- a/dask_planner/Cargo.toml
+++ b/dask_planner/Cargo.toml
@@ -11,10 +11,10 @@ rust-version = "1.62"
 [dependencies]
 arrow = { version = "22.0.0", features = ["prettyprint"] }
 async-trait = "0.1.41"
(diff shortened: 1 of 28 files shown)


dask_planner/Cargo.toml
```diff
@@ -9,19 +9,20 @@ edition = "2021"
...
-tokio = { version = "1.0", features = ["macros", "rt", "rt-multi-thread", "sync", "fs", "parking_lot"] }
-rand = "0.7"
-pyo3 = { version = "0.16.5", features = ["extension-module", "abi3", "abi3-py38"] }
 arrow = { version = "22.0.0", features = ["prettyprint"] }
-datafusion-sql = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-expr = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-common = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-optimizer = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-uuid = { version = "0.8", features = ["v4"] }
+async-trait = "0.1.41"
```
andygrove, line 13: I ran `cargo tomlfmt` here and added new dependencies on `log` and `env_logger`

dask_planner/src/sql/logical/aggregate.rs
```diff
@@ -45,47 +45,64 @@ impl PyAggregate {
...
     #[pyo3(name = "getAggregationFuncName")]
     pub fn agg_func_name(&self, expr: PyExpr) -> PyResult<String> {
-        match expr.expr {
-            Expr::AggregateFunction { fun, .. } => Ok(fun.to_string()),
-            Expr::AggregateUDF { fun, .. } => Ok(fun.name.clone()),
-            _ => Err(py_type_err(
-                "Encountered a non Aggregate type in agg_func_name",
-            )),
-        }
+        _agg_func_name(&expr.expr)
```
andygrove, line 48: The refactoring here is to support aliased aggregate expressions

dask_planner/Cargo.toml
```diff
@@ -9,19 +9,20 @@ edition = "2021"
...
-tokio = { version = "1.0", features = ["macros", "rt", "rt-multi-thread", "sync", "fs", "parking_lot"] }
-rand = "0.7"
-pyo3 = { version = "0.17.1", features = ["extension-module", "abi3", "abi3-py38"] }
 arrow = { version = "22.0.0", features = ["prettyprint"] }
-datafusion-sql = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-expr = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-common = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-datafusion-optimizer = { git="https://github.com/apache/arrow-datafusion/", rev = "12.0.0-rc1" }
-uuid = { version = "0.8", features = ["v4"] }
+async-trait = "0.1.41"
```
andygrove, line 13: I ran `cargo tomlfmt` hence the diff here. I also added `log` and `env_logger` dependencies

charlesbluca: Doesn't need to be done in this PR, but would we want a check like this to be run as part of the pre-commit hooks?

dask_planner/src/sql/logical/aggregate.rs
```diff
@@ -45,47 +45,64 @@ impl PyAggregate {
...
+fn _distinct_agg_expr(expr: &Expr) -> PyResult<bool> {
+    match expr {
+        Expr::Alias(expr, _) => _distinct_agg_expr(expr.as_ref()),
+        Expr::AggregateFunction {
+            fun: _,
+            args: _,
+            distinct,
+            filter: _,
+        } => Ok(*distinct),
+        _ => Err(py_type_err(
```
charlesbluca, line 103: Should we be returning some default value for `distinct` here when we encounter an `AggregateUDF`?

tests/integration/test_compatibility.py
```diff
@@ -276,7 +276,9 @@ def test_join_multi():
     )
 
 
-@pytest.mark.skip(reason="WIP DataFusion")
+@pytest.mark.skip(
+    reason="conflicting aggregation functions: [('count', 'a'), ('count', 'a')]"
+)
```
charlesbluca, line 281: Working on unblocking in #760

dask_planner/Cargo.toml
andygrove: Yes, ideally we would add this to the pre commit hook and add checks in CI as well

dask_planner/src/sql/logical/aggregate.rs
andygrove: That's a good point. I'll take a look.

tests/integration/test_compatibility.py
```diff
@@ -329,7 +348,7 @@ def test_agg_count_distinct_no_group_by():
 
 
 @pytest.mark.skip(
-    reason="WIP DataFusion - https://github.com/dask-contrib/dask-sql/issues/532"
+    reason="conflicting aggregation functions: [('count', 'c'), ('count', 'c')]"
```
ayushdg, line 351: Out of curiosity is this coming from 
```
COUNT(c) AS c_c,
COUNT(DISTINCT c) AS cd_c
```
where the optimization creates a `count(c)` for the latter case as well?

PR #746, opened 2022-09-06
Resolve `test_stats_aggregation`

sarahyurick:
Currently working on `test_stats_aggregation()` in `test_groupby.py`... so far have `REGR_COUNT` functionality and working on others.

PR #760, opened 2022-09-13
[DF] Add support for filtered groupby aggregations

charlesbluca:
Adds back support for filtered aggregations with `FILTER (WHERE ...)`, which allows us to unxfail `test_group_by_filtered` and unblocks several related tests in #746 and #759.


dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -364,9 +346,20 @@ def _collect_aggregations(
...
-            # Gather information about the input column
+            # Gather information about input columns
             inputs = agg.getArgs(expr)
 
+            # Compute any input columns that don't yet exist
+            for input_expr in inputs:
+                input_col = input_expr.column_name(input_rel)
+                if input_col in cc._frontend_backend_mapping:
+                    continue
+                random_name = new_temporary_column(df)
```
charlesbluca, line 357: A potential issue with `new_temporary_column` that came to mind while working on this (I say potential because I'm unsure of the behavior of `uuid.uuid4()`):

Since we're using the table's `columns` attribute to check that a random column name hasn't been used yet, and we don't actually assign any of these random columns names until several of them have been generated, it is technically possible (though rare) to accidentally assign multiple input / filter columns to the same random backend name, which will certainly cause issues.

Since `assign` calls are expensive and we ideally want to be adding all required backend columns in a single go, it might make sense to refactor `new_temporary_column` to instead look at some attribute of the `DataContainer` or `ColumnContainer` to check for duplicates, which are both cheaper to update on the fly.

Don't intend to block this PR, but could be worthwhile to open an issue / TODO to handle this down the line.

ayushdg: Agreed that moving the check to dc or cc should make things significantly cheaper. 
Based on the article [here](https://en.wikipedia.org/wiki/Universally_unique_identifier#Collisions) probability of name collision is almost negligible, especially at the scale at which we generate new columns

galipremsagar (approved):

PR #763, opened 2022-09-14
Fix upstream failures in `test_groupby_split_out`

### LINKED ISSUE #761: ⚠️ Upstream CI failed ⚠️
[Workflow Run URL](https://github.com/dask-contrib/dask-sql/actions/runs/3049617054)

charlesbluca:
Looks like with https://github.com/dask/dask/pull/9453 Dask's `aggregate` codepath has changed such that passing through a null `split_out` will error where it once didn't; not sure if this was something that was once supported, but looking at the codepath before this commit it seems like it would've errored in some other cases too.

Closes #761

### DIFF
--- a/tests/integration/test_groupby.py
+++ b/tests/integration/test_groupby.py
@@ -376,7 +376,7 @@ def test_groupby_split_out(c, input_table, split_out, request):
         FROM {input_table}
         GROUP BY user_id
         """,
-        config_options={"sql.groupby.split_out": split_out},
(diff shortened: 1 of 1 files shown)


ayushdg (approved):

ayushdg:
Thanks for the quick fix. Linking to the relevant dask issue where there's some discussion on dropping the `None` default. https://github.com/dask/dask/issues/9490

PR #765, opened 2022-09-15
[DF] Switch back to architectured builds

charlesbluca:
Turns out that our Rust binary is compiled for the Python version & platform of the build platform - there is probably a way to get cross-compiled packages so this isn't an issue, but a short-term unblocker is just to build & publish packages for all our supported Python versions.

### DIFF
--- a/.github/workflows/conda.yml
+++ b/.github/workflows/conda.yml
@@ -58,4 +58,4 @@ jobs:
           # install anaconda for upload
           mamba install anaconda-client
 
-          anaconda upload --label $LABEL noarch/*.tar.bz2
+          anaconda upload --label $LABEL linux-64/*.tar.bz2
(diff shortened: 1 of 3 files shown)


.github/workflows/conda.yml
```diff
@@ -5,6 +5,12 @@ on:
       - main
       - datafusion-sql-planner
   pull_request:
+    paths:
+      - setup.py
+      - continuous_integration/recipe/**
+      - .github/workflows/conda.yml
+  schedule:
+    - cron: '0 0 * * 0'
```
charlesbluca, line 13: Since these changes make conda builds take up much more time, this makes it so they only run when relevant files are impacted in PRs (and weekly to make sure things are working as expected)

ayushdg (approved):

PR #769, opened 2022-09-19
[DF] Use correct schema in TableProvider

andygrove:
When the SQL parser is trying to resolve a table, we were ignoring the schema in qualified names and using the default schema instead.

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -86,7 +86,10 @@ impl ContextProvider for DaskSQLContext {
             name.resolve(&self.default_catalog_name, &self.default_schema_name);
         if reference.catalog != self.default_catalog_name {
             // there is a single catalog in Dask SQL
-            return Err(DataFusionError::Plan(format!("Cannot resolve catalog '{}'", reference.catalog)));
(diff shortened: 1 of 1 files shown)


galipremsagar (approved):

ayushdg (approved):

dask_planner/src/sql.rs
```diff
@@ -84,7 +84,11 @@ impl ContextProvider for DaskSQLContext {
     ) -> Result<Arc<dyn TableSource>, DataFusionError> {
         let reference: ResolvedTableReference =
             name.resolve(&self.default_catalog_name, &self.default_schema_name);
-        match self.schemas.get(&self.default_schema_name) {
+        if reference.catalog != self.default_catalog_name {
+            // there is a single catalog in Dask SQL
+            return Err(DataFusionError::Plan(format!("Cannot resolve catalog '{}'", reference.catalog)));
```
galipremsagar, line 89:
```suggestion
            return Err(DataFusionError::Plan(format!(
                "Cannot resolve catalog '{}'",
                reference.catalog
            )));
```

PR #779, opened 2022-09-20
[DF] WIP: Upgrade to DataFusion rev 8ea59a

andygrove:
Creating this for a pairing session

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -550,6 +550,19 @@ impl PyExpr {
         }
     }
 
+    #[pyo3(name = "getDecimal128Value")]
+    pub fn decimal_128_value(&mut self) -> PyResult<(Option<i128>, u8, u8)> {
+        match &self.expr {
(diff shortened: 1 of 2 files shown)


dask_planner/src/expression.rs
```diff
@@ -529,152 +529,206 @@ impl PyExpr {
...
+                ScalarValue::TimestampNanosecond(iv, tz)
+                | ScalarValue::TimestampMicrosecond(iv, tz)
+                | ScalarValue::TimestampMillisecond(iv, tz)
+                | ScalarValue::TimestampSecond(iv, tz) => {
+                    if let Some(iv) = iv {
+                        println!("{}", iv);
+                    }
+                    if let Some(tz) = tz {
+                        println!("{}", tz);
+                    }
```
andygrove, line 687: If we want to keep this debug logging, we should use `trace!` rather than `println!`

ayushdg: nice catch. accidentally left it in

PR #783, opened 2022-09-21
Handle nullable types and empty partitions before Dask-ML predict

sarahyurick:
Tagging @randerzander and @VibhuJawa 

Changes in `create_model.py` handle nullable types, such as for this example:

```
df = pd.DataFrame({
    "rough_day_of_year": pd.Series([0, 1, 2, 3], dtype='Int32'),
    "prev_day_inches_rained": pd.Series([0, 1, 2, 3], dtype='float32'),
    "rained": pd.Series([False, False, False, True])
})
c.create_table("train_set", df)

model_class = ".linear_model.LogisticRegression'"
if GPU:
    model_class = "'cuml" + model_class
else:
    model_class = "'sklearn" + model_class

c.sql(f"""
CREATE OR REPLACE MODEL model WITH (
    model_class = {model_class},
     wrap_predict = True,
     wrap_fit = False,
    target_column = 'rained'
) AS (
    SELECT *
    FROM train_set
)
""")

c.sql("""
SELECT * FROM PREDICT(
  MODEL model,
  SELECT * FROM train_set
)
""").compute()
```

Changes in `predict.py` handle empty partitions, modeled based on this Dask-ML PR: https://github.com/dask/dask-ml/pull/912

### DIFF
--- a/dask_sql/physical/rel/custom/create_model.py
+++ b/dask_sql/physical/rel/custom/create_model.py
@@ -1,4 +1,5 @@
 import logging
+import numpy as np
 from typing import TYPE_CHECKING
 
 from dask import delayed
@@ -183,7 +184,13 @@ class CreateModelPlugin(BaseRelPlugin):
 
             delayed_model = [delayed(model.fit)(x_p, y_p) for x_p, y_p in zip(X_d, y_d)]
             model = delayed_model[0].compute()
-            model = ParallelPostFit(estimator=model)
+            output_meta = np.array([])
+            model = ParallelPostFit(
+                estimator=model,
+                predict_meta=output_meta,
+                predict_proba_meta=output_meta,
+                transform_meta=output_meta,
+            )
 
         else:
             model.fit(X, y, **fit_kwargs)
--- a/dask_sql/physical/rel/custom/create_model.py
+++ b/dask_sql/physical/rel/custom/create_model.py
@@ -1,7 +1,7 @@
 import logging
-import numpy as np
 from typing import TYPE_CHECKING
 
+import numpy as np
 from dask import delayed
 
 from dask_sql.datacontainer import DataContainer
--- a/dask_sql/physical/rel/custom/predict.py
+++ b/dask_sql/physical/rel/custom/predict.py
@@ -1,4 +1,5 @@
 import logging
+import numpy as np
 import uuid
 from typing import TYPE_CHECKING
 
@@ -59,7 +60,13 @@ class PredictModelPlugin(BaseRelPlugin):
 
         model, training_columns = context.schema[schema_name].models[model_name]
         df = context.sql(sql_select)
-        prediction = model.predict(df[training_columns])
+        part = df[training_columns]
+        output_meta = model.predict_meta
+        if part.shape[0].compute() == 0 and output_meta is not None:
+            empty_output = self.handle_empty_partitions(output_meta)
+            if empty_output is not None:
+                return empty_output
+        prediction = model.predict(part)
         predicted_df = df.assign(target=prediction)
 
         # Create a temporary context, which includes the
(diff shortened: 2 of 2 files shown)


dask_sql/physical/rel/custom/create_model.py
```diff
@@ -183,7 +184,13 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
 
             delayed_model = [delayed(model.fit)(x_p, y_p) for x_p, y_p in zip(X_d, y_d)]
             model = delayed_model[0].compute()
-            model = ParallelPostFit(estimator=model)
+            output_meta = np.array([])
```
sarahyurick, line 187: With this, `output_meta` is always `[]`

dask_sql/physical/rel/custom/predict.py
```diff
@@ -59,7 +60,13 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
 
         model, training_columns = context.schema[schema_name].models[model_name]
         df = context.sql(sql_select)
-        prediction = model.predict(df[training_columns])
+        part = df[training_columns]
+        output_meta = model.predict_meta
+        if part.shape[0].compute() == 0 and output_meta is not None:
```
sarahyurick, line 65: `compute` is needed on the Delayed object. I believe that right now, `output_meta` will always be `[]`

dask_sql/physical/rel/custom/predict.py
```diff
@@ -59,7 +60,13 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
 
         model, training_columns = context.schema[schema_name].models[model_name]
         df = context.sql(sql_select)
-        prediction = model.predict(df[training_columns])
+        part = df[training_columns]
+        output_meta = model.predict_meta
```
sarahyurick, line 64: AttributeError: 'KMeans' object has no attribute 'predict_meta'

dask_sql/physical/rel/custom/create_model.py
VibhuJawa: I dont think we can just hardcode the meta to be `output_meta` to be `np.array([])` .  We also use cuML for this case and that outputs a cuDF Series.

dask_sql/physical/rel/custom/predict.py
VibhuJawa: You dont need to compute for this to  do this, we can do it lazily too.

dask_sql/physical/rel/custom/predict.py
```diff
@@ -59,7 +61,16 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
+        part = df[training_columns]
+        try:
+            output_meta = model.predict_meta
+        except AttributeError:
+            output_meta = None
+        if part.shape[0].compute() == 0 and output_meta is not None:
+            empty_output = self.handle_empty_partitions(output_meta)
+            if empty_output is not None:
+                return empty_output
+        prediction = model.predict(part)
```
VibhuJawa, line 73: We should wrap the predict like the following for cases only for when we have a  `ParallelPostFit`  model.  


```python
  if isinstance(model, ParallelPostFit):
      output_meta = model.predict_meta
      if predict_meta is None:
         predict_meta = model.estimator.predict(part._meta_nonempty)
         
      prediction  = part.map_partitions(_predict, predict_meta, model.estimator, meta=predict_meta)

def _pedict(part, predict_meta, estimator):
    if part.shape[0] == 0 and predict_meta is not None:
        empty_output = handle_empty_partitions(output_meta)
        return empty_output
    return estimator.predict(part)
  ```

VibhuJawa (changes requested):
We should not hard code any meta values and should only handle cases when model is `ParallelPostFit ` .

VibhuJawa (changes requested):
Please add more tests

PR #796, opened 2022-09-22
Add support for `CREATE EXPERIMENT`, expand support for `WITH` kwargs

charlesbluca:
A continuation of @ChrisJar's work in https://github.com/dask-contrib/dask-sql/pull/752; implements the basic plumbing for `CREATE EXPERIMENT`, and adds a `PySqlArg` struct to parser.rs that should allow us to parse nested kwargs in statements that use `WITH`.


dask_planner/src/parser.rs
```diff
@@ -18,6 +22,170 @@ macro_rules! parser_err {
...
+    pub fn get_sql_value(&self) -> PyResult<String> {
+        match &self.custom {
+            None => match &self.expr {
+                Some(Expr::Identifier(Ident { value, .. })) => Ok(value.to_string()),
+                Some(Expr::Value(scalar)) => match scalar {
+                    Value::Boolean(value) => Ok(if *value {
+                        "1".to_string()
+                    } else {
+                        "".to_string()
+                    }),
```
charlesbluca, line 158: This is somewhat unclear mapping of booleans to string motivated by the fact that in Python, only empty strings are falsey (e.g. `bool("False") == True`); not sure if this serves as motivation to break up `getSqlValue` into several functions for the various different literal types.

charlesbluca:
gpuCI failures here are unrelated - working on a fix in https://github.com/dask-contrib/dask-sql/pull/802

dask_planner/src/parser.rs
```diff
@@ -18,6 +22,189 @@ macro_rules! parser_err {
...
+            None => match &self.expr {
+                Some(Expr::Identifier(Ident { value, .. })) => value.to_string(),
+                Some(Expr::Value(scalar)) => match scalar {
+                    Value::Boolean(value) => {
+                        if *value {
+                            "1".to_string()
+                        } else {
+                            "".to_string()
+                        }
+                    }
```
andygrove, line 169: nit: we can make this more concise 

```suggestion
                    Value::Boolean(true) => "1".to_string(),
                    Value::Boolean(false) => "".to_string(),

```

dask_planner/src/parser.rs
```diff
@@ -18,6 +22,189 @@ macro_rules! parser_err {
...
+                },
+                Some(Expr::UnaryOp {
+                    op: UnaryOperator::Minus,
+                    expr,
+                }) => match &**expr {
+                    Expr::Value(Value::Number(value, false)) => format!("-{}", value),
+                    unexpected => {
+                        return Err(py_type_err(format!(
+                            "Expected string, got {:?}",
```
andygrove, line 186:
```suggestion
                            "Expected number, got {:?}",
```

dask_planner/src/parser.rs
```diff
@@ -18,6 +22,189 @@ macro_rules! parser_err {
...
+                            "1".to_string()
+                        } else {
+                            "".to_string()
+                        }
+                    }
+                    Value::SingleQuotedString(string) => string.to_string(),
+                    Value::Number(value, false) => value.to_string(),
+                    unexpected => {
+                        return Err(py_type_err(format!(
+                            "Expected string, got {:?}",
```
andygrove, line 174:
```suggestion
                            "Expected scalar value, got {:?}",
```

dask_planner/src/parser.rs
```diff
@@ -18,6 +22,189 @@ macro_rules! parser_err {
...
+                        "Expected scalar value, got {:?}",
+                        unexpected
+                    )))
+                }
+            },
+            unexpected => {
+                return Err(py_type_err(format!(
+                    "Expected scalar value, got {:?}",
+                    unexpected
+                )))
```
andygrove, line 202: It seems like we could add a helper function for these errors to reduce duplication

dask_planner/src/parser.rs
```diff
@@ -18,6 +22,144 @@ macro_rules! parser_err {
...
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub enum CustomExpr {
+    Map(Vec<Expr>),
+    Multiset(Vec<Expr>),
+    Nested(Vec<(String, PySqlArg)>),
+}
+
+#[pyclass(name = "SqlArg", module = "datafusion")]
+#[derive(Debug, Clone, PartialEq, Eq)]
+pub struct PySqlArg {
```
andygrove, line 34: This looks like it should be an enum instead of a struct?

andygrove (approved):
LGTM. It might be worth considering changing `PySqlArg` from a struct to an enum in the future, as this would simplify some of the pattern matching.

ayushdg:
rerun tests

ayushdg (approved):
Changes generally lgtm! We can followup in another pr to simplify some structs as suggested by Andy and potentially consolidate of the pysqlArg methods

PR #800, opened 2022-09-23
Resolve bare requirement failures in upstream workflow

charlesbluca:
Installs `setuptools-rust` as part of the upstream dev bare requirements test, which is required to do a source install of the package. Also makes some modifications to the bare requirements job to make it consistent between the PR and scheduled workflows.


charlesbluca:
rerun tests

PR #802, opened 2022-09-23
Check `split_out` to decide on sorted groupby in aggregate.py

charlesbluca:
With https://github.com/dask/dask/pull/9486, Dask now defaults to `sort=True` for groupbys, which is incompatible with `split_out>1` and raises an [error](https://github.com/rapidsai/cudf/blob/204a09cb9de58485df2c931db9a56d19a6eda356/python/dask_cudf/dask_cudf/groupby.py#L527-L534) in dask-cuDF, causing failures in our `split_out` tests.

This PR resolves these failures by implicitly setting `sort=False` when we see that `split_out>1`. Another approach here could be to expose `sort` to users through the Dask configuration in the same way we do `split_out` and `split_every`.

### DIFF
--- a/dask_sql/physical/rel/logical/aggregate.py
+++ b/dask_sql/physical/rel/logical/aggregate.py
@@ -533,11 +533,14 @@ class DaskAggregatePlugin(BaseRelPlugin):
             for group_name in group_columns
         ]
 
+        # if split_out > 1, we cannot do a sorted groupby
+        sort = False if groupby_agg_options.get("split_out", 1) > 1 else True
+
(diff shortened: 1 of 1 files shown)


galipremsagar (approved):

ayushdg:
Given that dask now defaults to sort=True, it might be worth exploring if this impacts the perf of dask-cudf groupby's in anyway and if dask-cudf needs to update the default behavior. 

For context https://github.com/dask/dask/pull/5801 added a `sort` param in dask to allow for some perf improvements on the dask-cudf side. 

cc: @rjzamora and @galipremsagar

PR #803, opened 2022-09-23
Refactor `getValue<T>` code to reduce duplication

andygrove:
This is the first step in cleaning up some of the `PyExpr` code.

Changes in this PR:

- Introduce a new `DaskPlannerError` that we can use in non-Pythonic Rust methods instead of using DataFusion's error types
- Add `PyExpr::get_scalar_value`
- Refactor all `PyExpr::<type>_value` methods to call `PyExpr::get_scalar_value`
- Adds a few Rust unit tests

I can make this code much more concise with a macro, but I didn't want to go that far since we haven't discussed whether we want to start using macros in this project. I will follow up with a separate PR where we can have that discussion.

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -558,7 +558,7 @@ impl PyExpr {
     }
 
     #[pyo3(name = "getFloat32Value")]
-    pub fn float_32_value(&mut self) -> PyResult<Option<f32>> {
+    pub fn float_32_value(&self) -> PyResult<Option<f32>> {
         match self.get_scalar_value()? {
             ScalarValue::Float32(value) => Ok(*value),
             other => Err(unexpected_literal_value(other)),
@@ -566,7 +566,7 @@ impl PyExpr {
     }
 
     #[pyo3(name = "getFloat64Value")]
-    pub fn float_64_value(&mut self) -> PyResult<Option<f64>> {
+    pub fn float_64_value(&self) -> PyResult<Option<f64>> {
         match self.get_scalar_value()? {
             ScalarValue::Float64(value) => Ok(*value),
             other => Err(unexpected_literal_value(other)),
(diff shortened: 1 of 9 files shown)


dask_planner/src/expression.rs
```diff
@@ -821,10 +778,27 @@ impl PyExpr {
...
+            _ => Err(DaskPlannerError::Internal(
+                "get_scalar_value() called on non-literal expression".to_string(),
+            )),
+        }
+    }
+}
+
+fn unexpected_literal_value() -> PyErr {
+    DaskPlannerError::Internal("getValue<T>() - Unexpected value".to_string()).into()
+}
```
charlesbluca, line 796: Would it make sense to display the invalid scalar type received as part of this error message?

dask_planner/src/expression.rs
```diff
@@ -557,214 +557,156 @@ impl PyExpr {
...
-            Expr::Literal(scalar_value) => match scalar_value {
-                ScalarValue::Float32(iv) => Ok(*iv),
-                _ => Err(py_type_err("getValue<T>() - Unexpected value")),
-            },
-            _ => Err(py_type_err("getValue<T>() - Non literal value encountered")),
+    pub fn float_32_value(&self) -> PyResult<Option<f32>> {
+        match self.get_scalar_value()? {
+            ScalarValue::Float32(value) => Ok(*value),
+            other => Err(unexpected_literal_value(other)),
         }
```
charlesbluca, line 565: Is it possible/worthwhile to make a helper function (parametrized on the specific `ScalarValue` variant) for this basic match pattern?

andygrove: That would be easy to do with a macro, but there are pros and cons to using macros, so I planned on doing that in a follow-on PR.

andygrove: We could define this macro:

```rust
macro_rules! extract_scalar_value {
    ($self: expr, $variant: ident) => {
        match $self.get_scalar_value()? {
            ScalarValue::$variant(value) => Ok(*value),
            other => Err(unexpected_literal_value(other)),
        }
    }
}
```

and then most of the get_value methods become one-liners:

```rust
    #[pyo3(name = "getFloat32Value")]
    pub fn float_32_value(&self) -> PyResult<Option<f32>> {
        extract_scalar_value!(self, Float32)
    }
```

The main downside is that there is now another language for the maintainers to become familiar with.

charlesbluca: Thanks for the detailed response, in that case I'm okay with pushing this off to a follow-up PR, and we can have a large discussion around the use of macros.

charlesbluca (approved):

PR #804, opened 2022-09-23
Resolve Docker build failures, update core dependency constraints

charlesbluca:
This should resolve the Docker build issues we've been seeing since switching to the DataFusion planner; I also propagated the explicit constraints on `prompt_toolkit` and `pygments` to the rest of our packaging / CI environments, as I imagine having a min version on those dependencies is a good thing.

Also should note that having the Docker CI @ayushdg was working on in #583 would help to catch issues like this moving forward, though IIRC that got blocked so doesn't need to block this PR unless we run into more issues


ayushdg (approved):
Thanks!

charlesbluca:
Just want to wait for 3.8 CI to run to verify that the min constraints I added work as expected

PR #812, opened 2022-09-27
Resolve `test_literals()`

sarahyurick:
To do `SELECT TIME '08:08:00.091' AS "T"`, we just need to add logic for `Time64`. Since `Time32` is also an [Arrow Data Type](https://arrow.apache.org/docs/python/api/datatypes.html), I added that too.


dask_sql/mappings.py
```diff
@@ -137,6 +139,10 @@ def sql_to_python_value(sql_type: "SqlTypeName", literal_value: Any) -> Any:
             literal_value = literal_value.rstrip("'")
             literal_value = literal_value.encode(encoding=encoding)
             return literal_value.decode(encoding=encoding)
+        # If we have a datetime without a date,
+        # then we should offset relative to the Unix epoch
+        elif ":" in literal_value and len(literal_value.split(" ")) == 1:
+            return "1970-01-01 " + literal_value
```
sarahyurick, line 145: I found that hardcoding this was easier than trying to subtract today's date with a timedelta.

dask_sql/mappings.py
```diff
@@ -168,6 +174,8 @@ def sql_to_python_value(sql_type: "SqlTypeName", literal_value: Any) -> Any:
     elif (
         sql_type == SqlTypeName.TIMESTAMP
         or sql_type == SqlTypeName.TIME
+        or sql_type == SqlTypeName.TIME32
+        or sql_type == SqlTypeName.TIME64
```
sarahyurick, line 178: I mostly just added this to match the existing logic - the pytest still functions without it.

dask_sql/mappings.py
sarahyurick: Should also maybe add some sort of `is_datetime` check.

ayushdg: is there a specific case/test which was failing without this?

ayushdg (commented):
Thanks for this! Changes generally look good to me. 

Could you also add an implementation for `getTime32Value` similar to time64 and potentially add a test case for time32 if possible

dask_sql/mappings.py
sarahyurick: I found that without the `1970-01-10`, it automatically displays `SELECT TIME '08:08:00.091' AS "T"` with the current date, such as `2022-10-03 08:08:00.091`, even though the test is expecting a datetime relative to the Unix epoch.

I do think I should probably add some sort of check that the value is convertible to a datetime, just on the chance that `literal_value` is some random string that has a colon and space in it.

sarahyurick: Actually, I think the combination of these changes and Andy's changes in #825 removes the need for this `elif` block. I'll remove it and we can check again after his PR is merged.

dask_sql/physical/rex/core/literal.py
```diff
@@ -151,6 +152,9 @@ def convert(
         elif literal_type == "Date64":
             literal_type = SqlTypeName.DATE
             literal_value = np.datetime64(rex.getDate64Value(), "ms")
+        elif literal_type == "Time32":
+            literal_value = np.datetime64(rex.getTime32Value(), "ms")
```
sarahyurick, line 156: To match https://arrow.apache.org/docs/python/generated/pyarrow.time32.html

sarahyurick:
> Thanks for this! Changes generally look good to me.
> 
> Could you also add an implementation for `getTime32Value` similar to time64 and potentially add a test case for time32 if possible

Upon further inspection, I don't think much is needed in terms of Time32/Time64 stuff, and we can just use the existing `SqlTypeName::TIME`. Since DataFusion doesn't support `ScalarValue::Time32`, I don't think `getTime32Value` is necessary and can't be tested.

After @andygrove 's PR (#825) is merged, I think this PR will be very small.

dask_planner/src/sql/types.rs
```diff
@@ -246,6 +246,8 @@ impl SqlTypeName {
             DataType::Float16 => Ok(SqlTypeName::REAL),
             DataType::Float32 => Ok(SqlTypeName::FLOAT),
             DataType::Float64 => Ok(SqlTypeName::DOUBLE),
+            DataType::Time32(_unit) => Ok(SqlTypeName::TIME),
+            DataType::Time64(_unit) => Ok(SqlTypeName::TIME),
```
andygrove, line 250: We can combine these into one match:

```suggestion
            DataType::Time32(_) | DataType::Time64(_) => Ok(SqlTypeName::TIME),
```

andygrove (approved):
LGTM

ayushdg (approved):

PR #91, opened 2020-11-29
Feature/datetime functions

Yuhuishishishi:
Noob here.

I am trying to help with #11.
I have done a few remaining datetime related functions:
* TIMESTAMPADD
* LASTDAY
* CEIL (time unit=YEAR, WEEK, MONTH, QUARTER currently do not work)
* FLOOR (time unit=YEAR, WEEK, MONTH, QUARTER currently do not work)

I am still working on:
* TIMESTAMPDIFF

Would mind helping me review the diff and provide a few directions? Thanks

PR #813, opened 2022-09-27
Resolve `test_date_functions()`

sarahyurick:
- Extract microseconds/millennium/millisecond
- Last_day
- Timestampadd
- Ceil
- Floor


tests/integration/test_rex.py
```diff
@@ -549,39 +546,39 @@ def test_date_functions(c):
             EXTRACT(DOY FROM d) AS "doy",
             EXTRACT(HOUR FROM d) AS "hour",
             EXTRACT(MICROSECONDS FROM d) AS "microsecond",
-            EXTRACT(MILLENNIUM FROM d) AS "millennium",
+            -- EXTRACT(MILLENNIUM FROM d) AS "millennium",
```
ayushdg, line 549: sqlparser uses a different spelling with one `n` instead of two. Feel free to open up a pr on sqlparser with support for this spelling as well

sarahyurick: Thanks! I've opened it here: https://github.com/sqlparser-rs/sqlparser-rs/pull/633

tests/integration/test_rex.py
```diff
@@ -556,32 +553,32 @@ def test_date_functions(c):
...
-            FLOOR(d TO DAY) as floor_to_day,
-            FLOOR(d TO HOUR) as floor_to_hour,
-            FLOOR(d TO MINUTE) as floor_to_minute,
-            FLOOR(d TO SECOND) as floor_to_seconds,
-            FLOOR(d TO MILLISECOND) as floor_to_millisec
+            EXTRACT(YEAR FROM d) AS "year",
+
+            LAST_DAY(d) as "last_day"
+
+            -- TIMESTAMPADD(YEAR, 2, d) as "plus_1_year",
```
sarahyurick, line 560: https://www.w3resource.com/mysql/date-and-time-functions/mysql-timestampadd-function.php

tests/integration/test_rex.py
```diff
@@ -556,32 +553,32 @@ def test_date_functions(c):
...
+            -- TIMESTAMPADD(WEEK, 1, d) as "plus_1_week",
+            -- TIMESTAMPADD(DAY, 1, d) as "plus_1_day",
+            -- TIMESTAMPADD(HOUR, 1, d) as "plus_1_hour",
+            -- TIMESTAMPADD(MINUTE, 1, d) as "plus_1_min",
+            -- TIMESTAMPADD(SECOND, 1, d) as "plus_1_sec",
+            -- TIMESTAMPADD(MICROSECOND, 999*1000, d) as "plus_999_millisec",
+            -- TIMESTAMPADD(MICROSECOND, 999, d) as "plus_999_microsec",
+            -- TIMESTAMPADD(QUARTER, 1, d) as "plus_1_qt",
+
+            -- CEIL(d TO DAY) as ceil_to_day,
```
sarahyurick, line 571: Waiting on https://github.com/sqlparser-rs/sqlparser-rs/pull/635

AugustoFKL: @sarahyurick is there a reference to this format `<expr> TO <type>`, please? I can't find anything about that

sarahyurick: I'm having trouble finding it myself. For the Dask-SQL project, it was first implemented with #91, although not much context was provided.

AugustoFKL: @sarahyurick I'll keep looking but unless there's a source where we can find it, I don't think it's implementable at all.

It looks like the `CAST` from standard, but with smaller syntax, but I'm not quite sure as it seems really specific to a random dialect or something like that.

dask_planner/src/dialect.rs
```diff
@@ -37,4 +43,110 @@ impl Dialect for DaskDialect {
     fn supports_filter_during_aggregation(&self) -> bool {
         true
     }
+
+    /// override expression parsing
+    fn parse_prefix(&self, parser: &mut Parser) -> Option<Result<Expr, ParserError>> {
+        fn parse_expr(parser: &mut Parser) -> Result<Option<Expr>, ParserError> {
+            match parser.peek_token() {
+                Token::Word(w) if w.value.to_lowercase() == "timestampadd" => {
```
sarahyurick, line 51: From #857

dask_planner/src/sql.rs
```diff
@@ -152,6 +152,31 @@ impl ContextProvider for DaskSQLContext {
                 let rtf: ReturnTypeFunction = Arc::new(|_| Ok(Arc::new(DataType::Int64)));
                 return Some(Arc::new(ScalarUDF::new(name, &sig, &rtf, &fun)));
             }
+            "last_day" => {
+                let sig = Signature::variadic(
```
charlesbluca, line 156: Think we're going to want to amend the `Signature` variants we're using for these date functions to match up with the expected inputs; `variadic` implies that we could have an arbitrary amount of `Timestamps` passed through as input, when in this case we would only want one. For reference, [here's](https://docs.rs/datafusion-expr/13.0.0/datafusion_expr/struct.Signature.html) a list of the various signatures we could use.

I think that the signatures used here may be the cause for the datetime casting we later need to do on the Python end, as it looks like for the floor/ceil operations DataFusion is inserting some unnecessary casting to `Utf8` on the datetime columns.

Also want to note that this is a larger issue with almost all of our custom function signatures, and it would probably be worthwhile to comb through them and fix them where possible

dask_sql/physical/rex/core/call.py
```diff
@@ -99,6 +100,12 @@ def __init__(
         self.false_route = false_route
 
     def apply(self, *operands, **kwargs):
+        if (
+            isinstance(self, CeilFloorOperation)
+            and not isinstance(operands[0], datetime)
+            and len(operands) == 2
+        ):
+            operands = (operands[0].astype("datetime64[ns]"), operands[1])
```
charlesbluca, line 108: Like I mentioned above, think that if we modify the signatures for `timestampceil`/`timestampfloor` on the Rust end we should be able to remove this conditional casting.

That being said, if it needs to stick around, I think it would make more sense to override `apply` in `CeilFloorOperation`, with something like

```python
    def apply(self, *operands, **kwargs):
        if not isinstance(operands[0], datetime):
            operands = (operands[0].astype("datetime64[ns]"), operands[1])
        return super().apply(*operands, **kwargs)
```

sarahyurick: You're right, changing `variadic` to `exact` removes the need for this check. Thanks!

dask_planner/src/sql.rs
sarahyurick: Interesting, I changed `variadic` from the custom functions for this PR. I've opened #868 for tracking the functions outside of this PR.

dask_sql/physical/rex/core/call.py
```diff
@@ -614,22 +615,62 @@ def __init__(self):
         super().__init__(self.timestampadd)
 
     def timestampadd(self, unit, interval, df: SeriesOrScalar):
-        df = convert_to_datetime(df)
-
-        if unit in {"DAY", "SQL_TSI_DAY"}:
-            return df + np.timedelta64(interval, "D")
+        interval = int(interval)
```
charlesbluca, line 618: Like `date_part`, we'll want to normalize our `unit`'s casing here so that something like `select timestampadd(year, ...) ...` will work:

```suggestion
        unit = unit.upper()
        interval = int(interval)
```

dask_sql/physical/rex/core/call.py
```diff
@@ -614,22 +615,62 @@ def __init__(self):
...
-            return df + np.timedelta64(interval, "s")
-        elif unit in {"WEEK", "SQL_TSI_WEEK"}:
-            return df + np.timedelta64(interval * 7, "W")
+            return df + timedelta(seconds=interval)
+        elif unit in {"MILLISECOND", "MILLISECONDS"}:
+            return df + timedelta(miliseconds=interval)
+        elif unit in {"MICROSECOND", "MICROSECONDS"}:
+            return df + timedelta(microseconds=interval)
         else:
             raise NotImplementedError(f"Extraction of {unit} is not (yet) implemented.")
```
charlesbluca, line 675: We should probably update this error message to correspond to timestamp addition instead of date extraction

dask_planner/src/sql.rs
```diff
@@ -152,6 +152,31 @@ impl ContextProvider for DaskSQLContext {
...
+                    vec![DataType::Timestamp(TimeUnit::Nanosecond, None)],
+                    Volatility::Immutable,
+                );
+                let rtf: ReturnTypeFunction =
+                    Arc::new(|_| Ok(Arc::new(DataType::Timestamp(TimeUnit::Nanosecond, None))));
+                return Some(Arc::new(ScalarUDF::new(name, &sig, &rtf, &fun)));
+            }
+            "timestampadd" => {
+                let sig = Signature::exact(
+                    vec![DataType::Utf8, DataType::Int64, DataType::Date64],
```
charlesbluca, line 166: In the tests, I am getting the following optimizer warnings:

```
Skipping optimizer rule type_coercion due to unexpected error: Error during planning: Coercion from [Utf8, Int64, Timestamp(Nanosecond, None)] to the signature Exact([Utf8, Int64, Date64]) failed.
```

I would've assumed that we would be able to coerce a `Timestamp` into a `Date64`, but in general think that since they both have roughly the same corresponding Python type that we should probably allow for both signatures.

@andygrove is there any way we can easily define signatures that parametrize on the type of one or more inputs? In this case, we would probably want this signature to be `OneOf` with several signatures for the various int / date types we can accept

dask_sql/physical/rex/core/call.py
```diff
@@ -812,11 +853,11 @@ def date_part(self, what, df: SeriesOrScalar):
             return df.dayofyear
         elif what == "HOUR":
             return df.hour
-        elif what == "MICROSECOND":
+        elif what in {"MICROSECOND", "MICROSECONDS"}:
```
charlesbluca, line 856: Nitpicky, but if we add support for singular and plural units, we should probably apply this to all units where it's applicable (minutes, hours, days, etc.); trying to think of if there's a general way we can handle this as part of validating `unit`

dask_sql/physical/rex/core/call.py
```diff
@@ -812,11 +853,11 @@ def date_part(self, what, df: SeriesOrScalar):
...
             return df.microsecond
-        elif what == "MILLENNIUM":
+        elif what in {"MILLENIUM", "MILLENNIUM"}:
```
charlesbluca, line 858: Is `millenium` supported in specific grammars? I assumed it was a spelling variant of millennium but it seems like it's a misspelling?

dask_planner/src/dialect.rs
```diff
@@ -37,4 +43,110 @@ impl Dialect for DaskDialect {
...
+                    let args = vec![
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(Expr::Value(
+                            Value::SingleQuotedString(time_unit.to_string()),
+                        ))),
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(n)),
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(expr)),
+                    ];
+
+                    Ok(Some(Expr::Function(Function {
+                        name: ObjectName(vec![Ident::new("TIMESTAMPADD")]),
```
sarahyurick, line 72:
```suggestion
                        name: ObjectName(vec![Ident::new("timestampadd")]),
```

dask_sql/physical/rex/core/call.py
sarahyurick: I mainly added it because originally sqlparser-rs only supported `millenium` until support for both spellings was added with https://github.com/sqlparser-rs/sqlparser-rs/pull/633. I could change it back to just `millennium` though if desired.

charlesbluca: Thanks for the context! In that case, I'm okay with leaving both spellings in 🙂

dask_planner/src/dialect.rs
```diff
@@ -47,6 +48,70 @@ impl Dialect for DaskDialect {
     fn parse_prefix(&self, parser: &mut Parser) -> Option<Result<Expr, ParserError>> {
         fn parse_expr(parser: &mut Parser) -> Result<Option<Expr>, ParserError> {
             match parser.peek_token() {
+                Token::Word(w) if w.value.to_lowercase() == "ceil" => {
+                    // CEIL(d TO DAY)
+                    parser.next_token(); // skip ceil
+                    parser.expect_token(&Token::LParen)?;
+                    let expr = parser.parse_expr()?;
+                    let keyword_to = parser.parse_keyword(Keyword::TO);
+                    if !keyword_to {
```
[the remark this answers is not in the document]
sarahyurick: Thanks!

dask_sql/physical/rex/core/call.py
```diff
@@ -682,24 +682,67 @@ def __init__(self):
...
+        if unit in {"YEAR", "YEARS"}:
+            result = []
+            for date in df:
+                year = date.year + interval
+                # Check leap day
+                if year % 4 != 0 and date.month == 2 and date.day == 29:
+                    result.append(date.replace(year=year, month=3, day=1))
+                else:
+                    result.append(date.replace(year=year))
+            return pd.Series(result)
```
ayushdg, line 700: Is there an alternative approach that can achieve this without iterating over each row individually and specifically return a pandas series?

PR #817, opened 2022-09-28
Switch docs/CI away from conda-installed Rust

charlesbluca:
After some discussion with @andygrove and @ayushdg, I think it makes sense to encourage developers to install Rust through `rustup` rather than conda for a few reasons:

- better integration with dev tools like VSCode's `rust-analyzer`
- more flexibility in configuring `rustfmt` using nightly features
- less confusion in installing additional build requirements on Windows

This PR switches CI to use the Rust package included in the standard GHA runners, and modifies developer docs to add instructions on installing Rust.


.github/workflows/rust.yml
```diff
@@ -1,37 +1,32 @@
...
+  # "1" means line tables only, which is useful for panic tracebacks.
+  RUSTFLAGS: "-C debuginfo=1"
+
 jobs:
   # Check crate compiles
   linux-build-lib:
     name: cargo check
     runs-on: ubuntu-latest
-    container:
-      image: amd64/rust
```
charlesbluca, line 17: Shouldn't need to use this container as stable Rust is already included in the GHA runners

.github/workflows/rust.yml
```diff
@@ -1,37 +1,32 @@
...
         with:
-          # these represent dependencies downloaded by cargo
-          # and thus do not depend on the OS, arch nor rust version.
-          path: /github/home/.cargo
-          key: cargo-cache-
-      - name: Setup Rust toolchain
-        uses: ./.github/actions/setup-builder
-        with:
-          rust-version: stable
+          path: /home/runner/.cargo
```
charlesbluca, line 28: For reference, this is where the `.cargo` is located in an Ubuntu runner

ayushdg (approved):
Changes lgtm! 
Installing rust via rustup seems to simplify a lot of our workflows and makes it compatible with vscode extensions.

Just to clarify, for following releases will rust still be included as a pre-requisite to the dask-sql conda package (if it's even needed)?

charlesbluca:
> Just to clarify, for following releases will rust still be included as a pre-requisite to the dask-sql conda package (if it's even needed)?

Current releases don't have Rust listed as a runtime dependency, so shouldn't be any changes there. The biggest change I've made to the conda packages here is unpinning the `rust_compiler_version` so that they pull the latest stable Rust compiler, which is roughly equivalent to what's happening in CI (assuming the latest stable Rust is available on conda-forge).

continuous_integration/gpuci/environment.yaml
```diff
@@ -28,7 +28,6 @@ dependencies:
 - pytest-xdist
 - pytest
 - python=3.9
-- rust=1.62.1
```
charlesbluca, line 31: Switching over to `rustup` in gpuCI with https://github.com/rapidsai/dask-build-environment/pull/48

PR #825, opened 2022-09-30
Update DataFusion and change order of optimization rules

andygrove:
There have been a lot of improvements to the optimization rules in DataFusion recently, particularly related to type coercion. This PR picks up those improvements.

Closes https://github.com/dask-contrib/dask-sql/issues/844


andygrove:
`test_literals` is failing with a timezone-related issue ... seems to be off by 6 hours

```
E           AssertionError: numpy array are different
E           
E           numpy array values are different (100.0 %)
E           [index]: [0]
E           [left]:  [1649288001000000000]
E           [right]: [1649266401000000000]
```

dask_sql/physical/rex/core/literal.py
```diff
@@ -167,18 +167,17 @@ def convert(
...
+                "TimestampSecond": "s",
+                "TimestampMillisecond": "ms",
+                "TimestampMicrosecond": "us",
+                "TimestampNanosecond": "ns",
             }
+            numpy_unit = unit_mapping.get(literal_type)
             literal_value, timezone = rex.getTimestampValue()
             if timezone and timezone != "UTC":
                 raise ValueError("Non UTC timezones not supported")
-            literal_type = SqlTypeName.TIMESTAMP
```
sarahyurick, line 178: after this `if` block, you can have:
```
elif timezone is None:
    literal_value = datetime.fromtimestamp(literal_value // 10**9)
```
(make sure to do a `from datetime import datetime`)

sarahyurick: btw, I chose 10**9 to convert to seconds, as suggested by https://stackoverflow.com/questions/45423917/pandas-errno-75-value-too-large-for-defined-data-type, but maybe milliseconds (or something else) would be generally better.

andygrove: Thanks @sarahyurick that fixes the regression locally for me. I have pushed this change.

tests/integration/test_join.py
```diff
@@ -329,11 +329,11 @@ def test_intersect(c):
...
-    select * from df_simple
+    select a, b from df_simple
     intersect
-    select * from df_simple
+    select a, b from df_simple
     intersect
-    select * from df_wide
+    select a, b from df_wide
     ) hot_item
     limit 100
```
andygrove, line 338: This query was invalid. `intersect` requires that all relations have the same number of columns. For example, postgres would fail with `ERROR:  each INTERSECT query must have the same number of columns`. DataFusion now has a check for this.

dask_sql/physical/rex/core/literal.py
```diff
@@ -167,18 +168,19 @@ def convert(
...
+                "TimestampMillisecond": "ms",
+                "TimestampMicrosecond": "us",
+                "TimestampNanosecond": "ns",
             }
+            numpy_unit = unit_mapping.get(literal_type)
             literal_value, timezone = rex.getTimestampValue()
             if timezone and timezone != "UTC":
                 raise ValueError("Non UTC timezones not supported")
+            elif timezone is None:
+                literal_value = datetime.fromtimestamp(literal_value // 10**9)
```
sarahyurick, line 181: If gpuCI still fails, you can try adding `literal_value = str(literal_value)` after this line (but still in the `elif` block).

andygrove: > If gpuCI still fails, you can try adding `literal_value = str(literal_value)` after this line (but still in the `elif` block).

Thanks Sarah. I tried that but this still fails. Here is more info on the failure:

```
self = <dask_sql.physical.rex.core.call.ReduceOperation object at 0x7f4717a097c0>
operands = ('2001-03-09', datetime.timedelta(days=90)), kwargs = {}

    def reduce(self, *operands, **kwargs):
        if len(operands) > 1:
            if any(
                map(
                    lambda op: is_frame(op) & pd.api.types.is_datetime64_dtype(op),
                    operands,
                )
            ):
>           return reduce(partial(self.operation, **kwargs), operands)
E           TypeError: can only concatenate str (not "datetime.timedelta") to str

dask_sql/physical/rex/core/call.py:137: TypeError
```

Perhaps `def reduce` or `def as_timelike` need updating to support the interval type that the Rust code is returning?

It looks like @charlesbluca may be familiar with these methods and maybe has some insight here.

charlesbluca (approved):
LGTM

dask_sql/physical/rex/core/literal.py
ayushdg: Curious if this change is still needed

ayushdg (approved):
Minor question but otherwise lgtm!

dask_sql/physical/rex/core/literal.py
charlesbluca: I don't think we need the string casting, but it seems like the `datetime.fromtimestamp` is necessary to get the proper timezone for the datetime

galipremsagar (approved):

PR #832, opened 2022-09-30
Replace `dask_ml.wrappers.ParallelPostFit` with custom `ParallelPostFit` class

sarahyurick:
As part of our initiative to move away from Dask-ML, I've migrated some code from Dask-ML into Dask-SQL to support ParallelPostFit.


dask_sql/physical/rel/custom/wrappers.py
```diff
@@ -0,0 +1,448 @@
...
+        else:
+            scoring = self.scoring
+
+        if scoring:
+            if not dask.is_dask_collection(X) and not dask.is_dask_collection(y):
+                scorer = sklearn.metrics.get_scorer(scoring)
+            else:
+                # TODO: implement Dask-ML's get_scorer() function
+                # scorer = get_scorer(scoring, compute=compute)
+                pass
```
sarahyurick, line 226: Dask-ML has a utility function called `get_scorer` that we still need to replicate. The toy example I was working with passed without it, so I'm mostly just curious to see if any of our tests fail without it :)

dask_sql/physical/rel/custom/wrappers.py
```diff
@@ -0,0 +1,467 @@
...
+    attrs = {k: v for k, v in vars(from_estimator).items() if k.endswith("_")}
+
+    for k, v in attrs.items():
+        setattr(to_estimator, k, v)
+
+
+@contextlib.contextmanager
+def _timer(name, _logger=None, level="info"):
+    """
+    Output execution time of a function to the given logger level
```
sarahyurick, line 450: Do we care about having this function?

sarahyurick:
Let me know if you think any tests from Dask-ML's `tests/test_parallel_post_fit.py` or from https://github.com/dask/dask-ml/pull/912 should be added.

tests/integration/test_model.py
```diff
@@ -924,3 +924,79 @@ def test_experiment_automl_regressor(c, client, training_df):
     ), "Best model was not registered"
 
     check_trained_model(c, "my_automl_exp2")
+
+
+# TODO - many ML tests fail on clusters without sklearn - can we avoid this?
```
VibhuJawa, line 929: @ayushdg , What will be the recommended way to test something that has a install dependency ?

tests/integration/test_model.py
```diff
@@ -924,3 +924,79 @@ def test_experiment_automl_regressor(c, client, training_df):
...
+@skip_if_external_scheduler
```
sarahyurick, line 930: @charlesbluca Wanted to ask you about this since we have it for several of the `test_model` pytests

sarahyurick: Opened an issue for this here: https://github.com/dask-contrib/dask-sql/issues/861

dask_sql/physical/rel/custom/wrappers.py
```diff
@@ -0,0 +1,493 @@
...
+"""Meta-estimators for parallelizing estimators using the scikit-learn API."""
+import logging
+import warnings
+
+import dask.array as da
+import dask.dataframe as dd
+import dask.delayed
+import numpy as np
+import sklearn.base
+import sklearn.metrics
```
charlesbluca, line 12: Should we wrap these in a `try`/`except` similar to what we were doing for dask-ml since scikit-learn isn't a core dependency?

VibhuJawa (approved):
Small Change, looks good other-wise

dask_sql/physical/rel/custom/wrappers.py
```diff
@@ -0,0 +1,497 @@
...
+        else:
+            scoring = self.scoring
+
+        if scoring:
+            if not dask.is_dask_collection(X) and not dask.is_dask_collection(y):
+                scorer = sklearn.metrics.get_scorer(scoring)
+            else:
+                # TODO: implement Dask-ML's get_scorer() function
+                # scorer = get_scorer(scoring, compute=compute)
+                pass
```
VibhuJawa, line 236: Can you actually raise an error here instead of passing. 
```suggestion
               raise NotImplementedError("Relevant error message")
```

VibhuJawa (changes requested):
Requesed Small change, Looks good other-wise.

VibhuJawa (approved):
LGTM

dask_sql/physical/rel/custom/wrappers.py
```diff
@@ -0,0 +1,497 @@
...
+import dask.array as da
+import dask.dataframe as dd
+import dask.delayed
+import numpy as np
+
+try:
+    import sklearn.base
+    import sklearn.metrics
+except ImportError:  # pragma: no cover
+    raise ValueError("sklearn must be installed")
```
ayushdg, line 16: Minor nit: I would prefer the raised error here to be an `importError` as well. 
Looking through the codebase it seems like the original errors were also `valueErrors` but `importError` might be cleaner

ayushdg (approved):
Thanks for this @sarahyurick. Changes generally look good to me. 

In addition to some of the integration tests added here which check the interaction of sql with our wrappers, it might make sense (in a followup) to also add unit tests that independently test the wrappers functionality outside of sql and make sure things are working as expected. 

We can probably adapt the tests from https://github.com/dask/dask-ml/blob/94e52613fc87abd4ef8c97510539ded1303ae6f4/tests/test_parallel_post_fit.py and add it to the `tests/unit` section

charlesbluca (approved):

PR #834, opened 2022-10-03
Add quoting around partition keys for Hive table inputs

### LINKED ISSUE #179: Dask-SQL not wrapping partition value in quotes
Hello, I am querying a Hive table that is partitioned using string values that look like a date -`YYYY-MM-DD`. My `create_table` command fails with a `HiveSQLException` because the `-` isn't being escaped.
For example:

```
[SQL: DESCRIBE FORMATTED schema.table PARTITION (partition_col=2019-01-21)]
```

this should instead read `partition_col="2019-01-21"`.

I am using `dask-sql==0.3.6`

### LINKED ISSUE #234: multiple partitions for hive not delimited correctly
To get a list of hive partitions the command `SHOW PARTITIONS` is used:

https://github.com/dask-contrib/dask-sql/blob/ece7ec7858b12b2b54b372c34fc79197e9f2fc43/dask_sql/input_utils/hive.py#L265-L277

For tables with multiple partition keys, it will return a list that is `/` delimited.  For example, if the table has two partition keys, `date` and `region`, one of the entries returned might be like this:

`date=20210101/region=south`

This string is used without modifications to get additional information about each partition using `DESCRIBE FORMATTED {table_name} PARTITION ({partition})`.  If multiple partition keys exist when using this command, the list should be `,` separated, not `/` separated.

https://github.com/dask-contrib/dask-sql/blob/ece7ec7858b12b2b54b372c34fc79197e9f2fc43/dask_sql/input_utils/hive.py#L200-L203

Because of this, when I try to read in a table with multiple hive partition keys dask-sql throws an error.

The issue https://github.com/dask-contrib/dask-sql/issues/179 might also be somewhat related to the example above.

randerzander:
Fixes #179 
Fixes #234

### DIFF
--- a/tests/integration/test_hive.py
+++ b/tests/integration/test_hive.py
@@ -76,7 +76,11 @@ def hive_cursor():
             hostname="hive-server",
             name="hive-server",
             network="dask-sql-hive",
-            volumes=[f"{tmpdir}:{tmpdir}", f"{tmpdir_parted}:{tmpdir_parted}"],
+            volumes=[
+                f"{tmpdir}:{tmpdir}",
+                f"{tmpdir_parted}:{tmpdir_parted}",
+                f"{tmpdir_multiparted}:{tmpdir_multiparted}",
+            ],
             environment={
                 "HIVE_CORE_CONF_javax_jdo_option_ConnectionURL": "jdbc:postgresql://hive-metastore-postgresql/metastore",
                 **DEFAULT_CONFIG,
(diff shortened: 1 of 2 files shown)


tests/integration/test_hive.py
```diff
@@ -196,3 +211,16 @@ def test_select_partitions(hive_cursor):
...
+
+def test_select_multipartitions(hive_cursor):
+    c = Context()
+    c.create_table("df_parts", hive_cursor)
+
+    result_df = c.sql("SELECT * FROM df_parts")
+    expected_df = pd.DataFrame({"i": [1, 2], "j": [1, 2], "k": ["a", "b"]})
+    expected_df["i"] = expected_df["i"].astype("int32")
+    expected_df["j"] = expected_df["j"].astype("int64")
+    expected_df["k"] = expected_df["k"].astype("object")
```
ayushdg, line 224: We can probably get away without explicitly casting these since they are already the same type, but I don't have a strong opinion either way.

ayushdg (approved):
changes lgtm! Thanks for looking into this @randerzander

tests/integration/test_hive.py
randerzander: These were largely copied from the pre-existing Hive partition case. If anything, I think it makes it clearer at a quick glance exactly what type will be returned.

PR #838, opened 2022-10-05
Support `to_timestamp`

### LINKED ISSUE #831: [ENH] Support for to_timestamp
I'd like to be able to convert data representing time since UNIX epoch to explicit timestamps format with `to_timestamp`, like I can in Spark SQL and PosgreSQL.

```python
from pyspark.sql import SparkSession
import pandas as pd

spark = SparkSession.builder \
    .master("local") \
    .getOrCreate()


df = pd.DataFrame({
    "a": [1406073600, 1406073600, 1406073600],
})

sdf = spark.createDataFrame(df)
sdf.createOrReplaceTempView("df")

query = """
SELECT to_timestamp(a) as date
FROM df
"""
spark.sql(query).show()
+-------------------+
|               date|
+-------------------+
|2014-07-22 20:00:00|
|2014-07-22 20:00:00|
|2014-07-22 20:00:00|
+-------------------+
```

```python
import pandas as pd
from dask_sql import Context
import dask

c = Context()

c.create_table("df", df)

query = """
SELECT to_timestamp(a) as date
FROM df
"""
print(c.sql(query).compute())
KeyError                                  Traceback (most recent call last)
File ~/miniconda3/envs/rapids-22.10-dasksql/lib/python3.9/site-packages/dask_sql/physical/rex/core/call.py:1000, in RexCallPlugin.convert(self, rel, expr, dc, context)
    999 try:
-> 1000     operation = self.OPERATION_MAPPING[operator_name]
   1001 except KeyError:

KeyError: 'totimestamp'

During handling of the above exception, another exception occurred:

KeyError                                  Traceback (most recent call last)
File ~/miniconda3/envs/rapids-22.10-dasksql/lib/python3.9/site-packages/dask_sql/physical/rex/core/call.py:1003, in RexCallPlugin.convert(self, rel, expr, dc, context)
   1002 try:
-> 1003     operation = context.schema[schema_name].functions[operator_name]
   1004 except KeyError:  # pragma: no cover

KeyError: 'totimestamp'

During handling of the above exception, another exception occurred:

NotImplementedError                       Traceback (most recent call last)
Input In [5], in <cell line: 13>()
      7 c.create_table("df", df)
      9 query = """
     10 SELECT to_timestamp(a) as date
     11 FROM df
     12 """
---> 13 print(c.sql(query).compute())

File ~/miniconda3/envs/rapids-22.10-dasksql/lib/python3.9/site-packages/dask_sql/context.py:501, in Context.sql(self, sql, return_futures, dataframes, gpu, config_options)
    496 else:
    497     raise RuntimeError(
    498         f"Encountered unsupported `LogicalPlan` sql type: {type(sql)}"
    499     )
--> 501 return self._compute_table_from_rel(rel, return_futures)

File ~/miniconda3/envs/rapids-22.10-dasksql/lib/python3.9/site-packages/dask_sql/context.py:830, in Context._compute_table_from_rel(self, rel, return_futures)
    829 def _compute_table_from_rel(self, rel: "LogicalPlan", return_futures: bool = True):
--> 830     dc = RelConverter.convert(rel, context=self)
    832     # Optimization might remove some alias projects. Make sure to keep them here.
    833     select_names = [field for field in rel.getRowType().getFieldList()]

File ~/miniconda3/envs/rapids-22.10-dasksql/lib/python3.9/site-packages/dask_sql/physical/rel/convert.py:61, in RelConverter.convert(cls, rel, context)
     55     raise NotImplementedError(
     56         f"No relational conversion for node type {node_type} available (yet)."
     57     )
     58 logger.debug(
     59     f"Processing REL {rel} using {plugin_instance.__class__.__name__}..."
     60 )
---> 61 df = plugin_instance.convert(rel, context=context)
     62 logger.debug(f"Processed REL {rel} into {LoggableDataFrame(df)}")
     63 return df

File ~/miniconda3/envs/rapids-22.10-dasksql/lib/python3.9/site-packages/dask_sql/physical/rel/logical/project.py:57, in DaskProjectPlugin.convert(self, rel, context)
     55 else:
     56     random_name = new_temporary_column(df)
---> 57     new_columns[random_name] = RexConverter.convert(
     58         rel, expr, dc, context=context
     59     )
     60     logger.debug(f"Adding a new column {key} out of {expr}")
     61     new_mappings[key] = random_name

File ~/miniconda3/envs/rapids-22.10-dasksql/lib/python3.9/site-packages/dask_sql/physical/rex/convert.py:73, in RexConverter.convert(cls, rel, rex, dc, context)
     65     raise NotImplementedError(
     66         f"No conversion for class {expr_type} available (yet)."
     67     )
     69 logger.debug(
     70     f"Processing REX {rex} using {plugin_instance.__class__.__name__}..."
     71 )
---> 73 df = plugin_instance.convert(rel, rex, dc, context=context)
     74 logger.debug(f"Processed REX {rex} into {LoggableDataFrame(df)}")
     75 return df

File ~/miniconda3/envs/rapids-22.10-dasksql/lib/python3.9/site-packages/dask_sql/physical/rex/core/call.py:1005, in RexCallPlugin.convert(self, rel, expr, dc, context)
   1003         operation = context.schema[schema_name].functions[operator_name]
   1004     except KeyError:  # pragma: no cover
-> 1005         raise NotImplementedError(f"{operator_name} not (yet) implemented")
   1007 logger.debug(
   1008     f"Executing {operator_name} on {[str(LoggableDataFrame(df)) for df in operands]}"
   1009 )
   1011 kwargs = {}

NotImplementedError: totimestamp not (yet) implemented
```

sarahyurick:
Closes #831.


dask_sql/physical/rex/core/call.py
```diff
@@ -596,6 +596,16 @@ def extract(self, what, df: SeriesOrScalar):
             raise NotImplementedError(f"Extraction of {what} is not (yet) implemented.")
 
 
+class ToTimestampOperation(Operation):
+    def __init__(self):
+        super().__init__(self.to_timestamp)
+
+    def to_timestamp(self, df, format="%Y-%m-%d %H:%M:%S"):
+        df = df.astype("datetime64[s]")
+        df = df.dt.strftime(format)
```
sarahyurick, line 605: In addition to basic functionality on columns, PySpark allows an [optional `format` input](https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.to_timestamp.html). Right now, I'm having trouble parsing an additional string input.

tests/integration/test_rex.py
```diff
@@ -639,3 +639,30 @@ def test_date_functions(c):
...
+    df = pd.DataFrame(
+        {
+            "a": [1203073300, 1406073600, 2806073600],
+        }
+    )
+    c.create_table("df", df)
+
+    df = c.sql(
+        """
+        SELECT to_timestamp(a) AS date FROM df
```
charlesbluca, line 654: Maybe I'm missing something, but based on the `OPERATION_MAPPING` shouldn't this be

```suggestion
        SELECT totimestamp(a) AS date FROM df
```

dask_sql/physical/rex/core/call.py
sarahyurick: Hi @beckernick I was wondering what the desired input for the `format` parameter would look like? For reference, these are the variations I've tried and the errors generated on the DataFusion side:

`SELECT to_timestamp(a, format="%Y-%m-%d %H:%M:%S") AS date FROM df` gives a `ParsingException: SchemaError(FieldNotFound { qualifier: None, name: "format", valid_fields: Some(["df.a"]) })`

`SELECT to_timestamp(a, "%Y-%m-%d %H:%M:%S") AS date FROM df` gives a `ParsingException: SchemaError(FieldNotFound { qualifier: None, name: "%Y-%m-%d %H:%M:%S", valid_fields: Some(["df.a"]) })`

`SELECT to_timestamp(a, '%Y-%m-%d %H:%M:%S') AS date FROM df` gives a `ParsingException: Plan("Coercion from [Int64, Utf8] to the signature Uniform(1, [Int64, Timestamp(Nanosecond, None), Timestamp(Microsecond, None), Timestamp(Millisecond, None), Timestamp(Second, None), Utf8]) failed.")`

`SELECT to_timestamp(a, %Y-%m-%d %H:%M:%S) AS date FROM df` gives a `ParsingException: ParserError("Expected an expression:, found: %")`

dask_planner/src/dialect.rs
```diff
@@ -37,4 +43,93 @@ impl Dialect for DaskDialect {
     fn supports_filter_during_aggregation(&self) -> bool {
         true
     }
+
+    /// override expression parsing
+    fn parse_prefix(&self, parser: &mut Parser) -> Option<Result<Expr, ParserError>> {
+        fn parse_expr(parser: &mut Parser) -> Result<Option<Expr>, ParserError> {
+            match parser.peek_token() {
+                Token::Word(w) if w.value.to_lowercase() == "timestampadd" => {
```
sarahyurick, line 51: From #857

dask_planner/src/dialect.rs
```diff
@@ -37,4 +42,93 @@ impl Dialect for DaskDialect {
...
+                            ))),
+                        ];
+
+                        Ok(Some(Expr::Function(Function {
+                            name: ObjectName(vec![Ident::new("dsql_totimestamp")]),
+                            args,
+                            over: None,
+                            distinct: false,
+                            special: false,
+                        })))
```
charlesbluca, line 122: Think we could condense this by just setting `time_format` conditionally on the value of `comma`:

```suggestion
                    let time_format = if comma {
                        parser.next_token().to_string()
                    } else {
                        "%Y-%m-%d %H:%M:%S".to_string()
                    };
                    parser.expect_token(&Token::RParen)?;

                    // convert to function args
                    let args = vec![
                        FunctionArg::Unnamed(FunctionArgExpr::Expr(expr)),
                        FunctionArg::Unnamed(FunctionArgExpr::Expr(Expr::Value(
                            Value::SingleQuotedString(time_format),
                        ))),
                    ];

                    Ok(Some(Expr::Function(Function {
                        name: ObjectName(vec![Ident::new("dsql_totimestamp")]),
                        args,
                        over: None,
                        distinct: false,
                        special: false,
                    })))
```

dask_planner/src/sql.rs
```diff
@@ -152,6 +152,14 @@ impl ContextProvider for DaskSQLContext {
                 let rtf: ReturnTypeFunction = Arc::new(|_| Ok(Arc::new(DataType::Int64)));
                 return Some(Arc::new(ScalarUDF::new(name, &sig, &rtf, &fun)));
             }
+            "dsql_totimestamp" => {
+                let sig = Signature::variadic(
+                    vec![DataType::Date64, DataType::Utf8],
+                    Volatility::Immutable,
+                );
```
charlesbluca, line 159: We'll probably want this to be either `Signature::one_of` if we want to allow a variety of numeric types as input; cc @andygrove in case you know of an intuitive way to write out this function signature as it seems like it might be cumbersome to enumerate all the different supportable numeric types here

dask_sql/physical/rex/core/call.py
```diff
@@ -600,6 +600,21 @@ def extract(self, what, df: SeriesOrScalar):
...
+class ToTimestampOperation(Operation):
+    def __init__(self):
+        super().__init__(self.to_timestamp)
+
+    def to_timestamp(self, df, format):
+        # Remove double and single quotes from string
+        format = format.replace('"', "")
+        format = format.replace("'", "")
+
+        df = pd.to_datetime(df, unit="s")
```
charlesbluca, line 612: Depending on the outcome of the above modifications to `dsql_totimestamp`'s signature, we may not need this casting to datetimes

dask_planner/src/dialect.rs
```diff
@@ -37,4 +42,93 @@ impl Dialect for DaskDialect {
...
+                    }
```
sarahyurick, line 123:
```suggestion
```

dask_planner/src/sql.rs
charlesbluca: No, those should be fine 🙂 I was mostly thinking about if there was a simpler way to enumerate all the variations of ints, i.e. `[U]Int[8|16|32|64]`

charlesbluca (commented):
With #857 closed in favor of this, would you mind adding this [`parser.rs` test](https://github.com/dask-contrib/dask-sql/pull/857/files#diff-69e778a867a11000e6997abe3a2fa3a81d048f867f11ec504fc132917a1847fe) in?

dask_planner/src/sql.rs
```diff
@@ -152,6 +152,24 @@ impl ContextProvider for DaskSQLContext {
...
+                    vec![
+                        TypeSignature::Variadic(vec![DataType::Int8, DataType::Utf8]),
+                        TypeSignature::Variadic(vec![DataType::Int16, DataType::Utf8]),
+                        TypeSignature::Variadic(vec![DataType::Int32, DataType::Utf8]),
+                        TypeSignature::Variadic(vec![DataType::Int64, DataType::Utf8]),
+                        TypeSignature::Variadic(vec![DataType::UInt8, DataType::Utf8]),
+                        TypeSignature::Variadic(vec![DataType::UInt16, DataType::Utf8]),
+                        TypeSignature::Variadic(vec![DataType::UInt32, DataType::Utf8]),
+                        TypeSignature::Variadic(vec![DataType::UInt64, DataType::Utf8]),
+                        TypeSignature::Variadic(vec![DataType::Utf8, DataType::Utf8]),
```
sarahyurick, line 166:
```suggestion
                        TypeSignature::Exact(vec![DataType::Int8, DataType::Utf8]),
                        TypeSignature::Exact(vec![DataType::Int16, DataType::Utf8]),
                        TypeSignature::Exact(vec![DataType::Int32, DataType::Utf8]),
                        TypeSignature::Exact(vec![DataType::Int64, DataType::Utf8]),
                        TypeSignature::Exact(vec![DataType::UInt8, DataType::Utf8]),
                        TypeSignature::Exact(vec![DataType::UInt16, DataType::Utf8]),
                        TypeSignature::Exact(vec![DataType::UInt32, DataType::Utf8]),
                        TypeSignature::Exact(vec![DataType::UInt64, DataType::Utf8]),
                        TypeSignature::Exact(vec![DataType::Utf8, DataType::Utf8]),
```

sarahyurick:
I'm not quite sure why `test_fsql` is consistently failing at https://github.com/dask-contrib/dask-sql/blob/main/tests/integration/test_fugue.py#L67

ayushdg:
> I'm not quite sure why `test_fsql` is consistently failing at https://github.com/dask-contrib/dask-sql/blob/main/tests/integration/test_fugue.py#L67

I saw this pop up on a some tests unrelated to this pr: https://github.com/dask-contrib/dask-sql/actions/runs/3292477865/attempts/1. 

They seem to be presenting only on Mac-os so far and seems intermittent. Might need to investigate further if this keeps showing up.

dask_planner/src/dialect.rs
```diff
@@ -37,4 +42,74 @@ impl Dialect for DaskDialect {
...
+                    // convert to function args
+                    let args = vec![
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(expr)),
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(Expr::Value(
+                            Value::SingleQuotedString(time_format),
+                        ))),
+                    ];
+
+                    Ok(Some(Expr::Function(Function {
+                        name: ObjectName(vec![Ident::new("dsql_totimestamp")]),
```
ayushdg, line 99: Is the naming here to avoid collision with a datafusion function name?

dask_planner/src/parser.rs
```diff
@@ -1236,6 +1236,23 @@ impl<'a> DaskParser<'a> {
 mod test {
     use crate::parser::{DaskParser, DaskStatement};
 
+    #[test]
+    fn timestampadd() {
```
ayushdg, line 1240: We can probably add a `to_timestamp` test here as well

dask_sql/physical/rex/core/call.py
```diff
@@ -600,6 +600,25 @@ def extract(self, what, df: SeriesOrScalar):
...
+        format = format.replace("'", "")
+
+        if df.dtype == "object":
+            df = pd.to_datetime(df)
+            return df.strftime(format)
+        else:
+            df = pd.to_datetime(df, unit="s")
+            df = df.strftime(format)
+            result = [timestamp for timestamp in df]
+            return pd.Series(result)
```
ayushdg, line 619: is it possible to generalize this approach to support cudf dataframes as well?

ayushdg (changes requested):

dask_planner/src/dialect.rs
sarahyurick: Yes

dask_sql/physical/rex/core/call.py
```diff
@@ -35,6 +35,11 @@
     import dask_sql
     from dask_planner.rust import Expression, LogicalPlan
 
+try:
+    import cudf as pdlike
+except ImportError:
+    import pandas as pdlike
+
```
sarahyurick, line 42: Hi @ayushdg, let me know what you think about this generalization.

dask_sql/physical/rex/core/call.py
```diff
@@ -600,6 +605,25 @@ def extract(self, what, df: SeriesOrScalar):
...
+        if df.dtype == "object":
+            df = pdlike.to_datetime(df)
+            return df.strftime(format)
+        else:
+            df = pdlike.to_datetime(df, unit="s")
+            df = df.strftime(format)
+            result = [timestamp for timestamp in df]
+            return pdlike.Series(result)
+
```
ayushdg, line 625: qq: is one for string inputs like '2021-12-12' vs the other for integer inputs?

Do we need to do `[timestamp for timestamp in df]`. If we can avoid this we can avoid a host based list for gpu inputs.

tests/integration/test_rex.py
```diff
@@ -673,3 +673,81 @@ def test_date_functions(c):
             FROM df
             """
         )
+
+
+def test_totimestamp(c):
```
ayushdg, line 678: It would be good to generalize this for gpu+cpu tests

ayushdg (changes requested):

dask_sql/physical/rex/core/call.py
sarahyurick: Yes, the first block is for string '2021-12-12' inputs and the second for numeric inputs.

I can double check about other possible ways besides `[timestamp for timestamp in df]`, but from what I remember extra logic was needed to handle this case.

dask_sql/physical/rex/core/call.py
```diff
@@ -600,6 +605,25 @@ def extract(self, what, df: SeriesOrScalar):
...
+        format = format.replace("'", "")
+
+        if df.dtype == "object":
+            df = pdlike.to_datetime(df)
+            return df.strftime(format)
+        else:
+            df = pdlike.to_datetime(df, unit="s")
+            df = df.strftime(format)
+            result = [timestamp for timestamp in df]
+            return pdlike.Series(result)
```
sarahyurick, line 624:
```suggestion
            df = pdlike.to_datetime(df, unit="s")
            df = df.strftime(format)
            result = np.array([])
            for timestamp in df:
                result = np.append(result, timestamp)
            return pdlike.Series(result)
```

Hi @ayushdg would this be a more preferred way?

ayushdg: I think I understand, is it the case that `df.strftime` returns a string or list of strings that we need to reconvert into a series?

sarahyurick: Yes, but when I try to append directly to an empty series I got a `TypeError: cannot concatenate object of type '<class 'str'>'; only Series and DataFrame objs are valid`.

tests/integration/test_rex.py
```diff
@@ -677,3 +677,113 @@ def test_date_functions(c):
...
+    if gpu:
+        import cudf
+
+    df = pd.DataFrame(
+        {
+            "a": np.array([1203073300, 1406073600, 2806073600]),
+        }
+    )
+    if gpu:
+        df = cudf.from_pandas(df)
```
charlesbluca, line 693: Don't think we need this as the pandas frame should be getting wrapped in a `dask_cudf.DataFrame` through `create_table`:

```suggestion
```

tests/integration/test_rex.py
```diff
@@ -677,3 +677,113 @@ def test_date_functions(c):
             FROM df
             """
         )
+
+
+@pytest.mark.parametrize("gpu", [False, pytest.param(True, marks=pytest.mark.gpu)])
+def test_totimestamp(c, gpu):
+    if gpu:
+        import cudf
+
```
charlesbluca, line 686:
```suggestion
```

tests/integration/test_rex.py
```diff
@@ -677,3 +677,113 @@ def test_date_functions(c):
...
+        {
+            "date": [
+                datetime(2008, 2, 15),
+                datetime(2014, 7, 23),
+                datetime(2058, 2, 12),
+            ],
+        }
+    )
+    if not gpu:
+        assert_eq(df, expected_df, check_dtype=False)
```
charlesbluca, line 727: Does this currently fail on GPU? If so interested in what error message we see.

If this does fail currently, would it be possible to split this GPU case off into its own xfailing test so that we can better track it (related to what I'm doing in #867)

tests/integration/test_rex.py
```diff
@@ -677,3 +677,113 @@ def test_date_functions(c):
...
+    expected_df = pd.DataFrame(
+        {
+            "date": [
+                datetime(1997, 2, 28),
+                datetime(1997, 3, 28),
+            ],
+        }
+    )
+    if not gpu:
+        assert_eq(df, expected_df, check_dtype=False)
```
charlesbluca, line 767: Same comment as above

tests/integration/test_rex.py
sarahyurick: Yes, I was having trouble getting `dd.to_datetime()` (or `cudf.to_datetime()` or any other datetime functions) to be compatible with cuDF Series, which is fine if we're just using `to_timestamp()` with the default formatting string, but otherwise I couldn't figure out a good way to reformat timestamps without causing errors when `compute()` is called (hence the printed message [here](https://github.com/dask-contrib/dask-sql/blob/df688e7329f0caa8b53fee60130883586622a9e8/dask_sql/physical/rex/core/call.py#L616)).

Right now it's unclear to me whether this would need to be fixed on the Dask side, the cuDF side, or whether there's a Dask-SQL solution I missed. I can open an issue to look further into this if that works?

This is the error message I would get when trying to use a solution with `to_datetime` and `strftime` to format the timestamp:
```
c = <dask_sql.context.Context object at 0x7f3f3ea2b0a0>, gpu = True

    @pytest.mark.parametrize("gpu", [False, pytest.param(True, marks=pytest.mark.gpu)])
    def test_totimestamp(c, gpu):
        if gpu:
            import cudf
    
        df = pd.DataFrame(
            {
                "a": np.array([1203073300, 1406073600, 2806073600]),
            }
        )
        if gpu:
            df = cudf.from_pandas(df)
        c.create_table("df", df, gpu=gpu)
    
        df = c.sql(
            """
            SELECT to_timestamp(a) AS date FROM df
        """
        )
        expected_df = pd.DataFrame(
            {
                "date": [
                    datetime(2008, 2, 15, 11, 1, 40),
                    datetime(2014, 7, 23),
                    datetime(2058, 12, 2, 16, 53, 20),
                ],
            }
        )
>       assert_eq(df, expected_df, check_dtype=False)

tests/integration/test_rex.py:710: 
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 
tests/utils.py:13: in assert_eq
    return _assert_eq(*args, **kwargs)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/dataframe/utils.py:545: in assert_eq
    assert_divisions(a, scheduler=scheduler)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/dataframe/utils.py:620: in assert_divisions
    results = get(ddf.dask, ddf.__dask_keys__())
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/local.py:557: in get_sync
    return get_async(
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/local.py:500: in get_async
    for key, res_info, failed in queue_get(queue).result():
/opt/conda/envs/dask_sql/lib/python3.9/concurrent/futures/_base.py:439: in result
    return self.__get_result()
/opt/conda/envs/dask_sql/lib/python3.9/concurrent/futures/_base.py:391: in __get_result
    raise self._exception
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/local.py:542: in submit
    fut.set_result(fn(*args, **kwargs))
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/local.py:238: in batch_execute_tasks
    return [execute_task(*a) for a in it]
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/local.py:238: in <listcomp>
    return [execute_task(*a) for a in it]
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/local.py:229: in execute_task
    result = pack_exception(e, dumps)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/local.py:224: in execute_task
    result = _execute_task(task, data)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/core.py:119: in _execute_task
    return func(*(_execute_task(a, cache) for a in args))
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/optimization.py:990: in __call__
    return core.get(self.dsk, self.outkey, dict(zip(self.inkeys, args)))
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/core.py:149: in get
    result = _execute_task(task, cache)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/core.py:119: in _execute_task
    return func(*(_execute_task(a, cache) for a in args))
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/utils.py:71: in apply
    return func(*args, **kwargs)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/dask/dataframe/core.py:6770: in apply_and_enforce
    df = func(*args, **kwargs)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/pandas/core/tools/datetimes.py:1100: in to_datetime
    result = convert_listlike(argc, format)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/pandas/core/tools/datetimes.py:393: in _convert_listlike_datetimes
    return _to_datetime_with_unit(arg, unit, name, tz, errors)
/opt/conda/envs/dask_sql/lib/python3.9/site-packages/pandas/core/tools/datetimes.py:556: in _to_datetime_with_unit
    arg = np.asarray(arg)
_ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ _ 

self = 0    1203073300
1    1406073600
2    2806073600
Name: a, dtype: int64
dtype = None

    def __array__(self, dtype=None):
>       raise TypeError(
            "Implicit conversion to a host NumPy array via __array__ is not "
            "allowed, To explicitly construct a GPU matrix, consider using "
            ".to_cupy()\nTo explicitly construct a host matrix, consider "
            "using .to_numpy()."
        )
E       TypeError: Implicit conversion to a host NumPy array via __array__ is not allowed, To explicitly construct a GPU matrix, consider using .to_cupy()
E       To explicitly construct a host matrix, consider using .to_numpy().

/opt/conda/envs/dask_sql/lib/python3.9/site-packages/cudf/core/frame.py:447: TypeError
```

With my current changes, the error is just due to `expected_df` being a truncated version of `df`.

tests/integration/test_rex.py
```diff
@@ -677,3 +677,113 @@ def test_date_functions(c):
...
+
+    df = pd.DataFrame(
+        {
+            "a": np.array(["1997-02-28 10:30:00", "1997-03-28 10:30:01"]),
+        }
+    )
+    if gpu:
+        df = cudf.from_pandas(df)
```
sarahyurick, line 735:
```suggestion
```

tests/integration/test_rex.py
ayushdg: Looking a bit deeper it does seem like `dd.to_datetime` doesn't work completely with cudf backed dataframes, 
here's a small reproducer:
```python
import cudf
import dask_cudf

 a = cudf.Series(['2022-01-01', '2022-02-02'])
dser = dask_cudf.from_cudf(a, 1)
res = dd.to_datetime(dser).compute()
```

We could start off with an issue in dask/ dask-cudf and take it from there.

dask_sql/physical/rex/core/call.py
```diff
@@ -613,6 +614,37 @@ def extract(self, what, df: SeriesOrScalar):
...
+
+    def to_timestamp(self, df, format):
+        # Remove double and single quotes from string
+        format = format.replace('"', "")
+        format = format.replace("'", "")
+
+        # TODO: format timestamps for GPU tests
+        if "cudf" in str(type(df)):
+            if format != "%Y-%m-%d %H:%M:%S":
+                print("Formatting timestamps not supported on GPU")
```
ayushdg, line 629: based on my tests the gpu case fails for all cases when using `dd.to_datetime` and there's an additional bug with using some formats with `cudf.to_datetime`. It would be good to confirm that and replace this section with a error instead of a print.

dask_sql/physical/rex/core/call.py
```diff
@@ -613,6 +614,37 @@ def extract(self, what, df: SeriesOrScalar):
...
+            if df.dtype == "object":
+                return df
+            else:
+                return df * 10**9
```
ayushdg, line 633: which case is this handling?

dask_sql/physical/rex/core/call.py
```diff
@@ -613,6 +614,37 @@ def extract(self, what, df: SeriesOrScalar):
...
+            return datetime.strptime(df, format)
+        elif df.dtype == "object":
+            df = dd.to_datetime(df)
+        # Integer cases
+        elif np.isscalar(df):
+            df = datetime.utcfromtimestamp(df)
+            return df.strftime(format)
+        else:
+            df = dd.to_datetime(df, unit="s")
+        return df.dt.strftime(format)
```
ayushdg, line 645: The `to_timestamp` method is expected to return a datetime64 object rather than a string. The `format` argument here specified how the input may be formatted for strings but doesn't require the output to be formatted that way.

ayushdg (changes requested):
Added some comments below.

Also looking through the postgresql [docs](https://www.postgresql.org/docs/15/functions-datetime.html) the integer version of this method doesn't accept a `format` argument. Do you think it's possible to update the parser to handle this? If not we might need to either throw an error or ignore the format for integer cases.

Another note is that the `format` [argument](https://www.postgresql.org/docs/15/functions-formatting.html) for postgresql expects a different formatting pattern than python's datetime. For now we should be good to use the python format but add that to the docs

dask_sql/physical/rex/core/call.py
sarahyurick: This case handles a cuDF Series of integers - we multiply by 10**9 to convert from nanoseconds to seconds. It's essentially the same functionality as when we use `dd.to_datetime(df, unit="s")` for Pandas Series, except that `dd.to_datetime()` doesn't work with cuDF Series.

dask_sql/physical/rex/core/call.py
sarahyurick: Oh, I interpreted `format` to be the format of the timestamp that we want to be returned, i.e., setting `format="yyyy-mm-dd"` is telling Dask-SQL that we want the outputted column to be a datetime64 object to be formatted as `yyyy-mm-dd`. From the PostgreSQL docs, it seems more obvious that `format` is supposed to be telling Dask-SQL what format the user entered the timestamp as. I can update that.

I was also looking at PySpark docs when writing the formatting patterns: https://spark.apache.org/docs/latest/api/python/reference/pyspark.sql/api/pyspark.sql.functions.to_timestamp.html

sarahyurick:
> Also looking through the postgresql [docs](https://www.postgresql.org/docs/15/functions-datetime.html) the integer version of this method doesn't accept a `format` argument. Do you think it's possible to update the parser to handle this? If not we might need to either throw an error or ignore the format for integer cases.

Interesting, thanks! Yes, this should be easy to fix on the Rust side, so I can update this as well.

tests/integration/test_rex.py
```diff
@@ -677,3 +677,110 @@ def test_date_functions(c):
             FROM df
             """
         )
+
+
+@pytest.mark.parametrize(
+    "gpu", [False, pytest.param(True, marks=(pytest.mark.gpu, pytest.mark.skip))]
+)
```
ayushdg, line 684:
```suggestion
@pytest.mark.parametrize(
    "gpu",
    [
        False,
        pytest.param(
            True,
            marks=(
                pytest.mark.gpu,
                pytest.mark.xfail(
                    reason="Failing due to dask-cudf bug https://github.com/rapidsai/cudf/issues/12062"
                ),
            ),
        ),
    ],
)
```

dask_sql/physical/rex/core/call.py
```diff
@@ -613,6 +614,41 @@ def extract(self, what, df: SeriesOrScalar):
...
+            if format != default_format:
+                raise RuntimeError("Non-default timestamp formats not supported on GPU")
+            if df.dtype == "object":
+                return df
+            else:
+                nanoseconds_to_seconds = 10**9
+                return df * nanoseconds_to_seconds
+        # String cases
+        elif type(df) == str:
+            return datetime.strptime(df, format)
```
ayushdg, line 638: Is it possible to return a NumPy datetime64 here instead.

dask_sql/physical/rex/core/call.py
```diff
@@ -613,6 +614,41 @@ def extract(self, what, df: SeriesOrScalar):
...
+        elif df.dtype == "object":
+            return dd.to_datetime(df, format=format)
+        # Integer cases
+        elif np.isscalar(df):
+            if format != default_format:
+                raise RuntimeError("Integer input does not accept a format argument")
+            return datetime.utcfromtimestamp(df)
```
ayushdg, line 645: same as above

ayushdg (changes requested):

dask_sql/physical/rex/core/call.py
sarahyurick: From what I tried, `np.datetime64` can't use the `format` param the same way `datetime.strptime` can. I could wrap it in a NumPy datetime64, like `np.datetime64(datetime.strptime(df, format))`, if desired, though.

dask_sql/physical/rex/core/call.py
sarahyurick: Done, thanks!

dask_sql/physical/rex/core/call.py
ayushdg: Yeah I think that works. Just suggested this change to standardize on all python return types for user requesting sql timestamp to be inline with datetime64 as specified here: https://github.com/dask-contrib/dask-sql/blob/5440effe0d26c746650c4e801c7f846efd17f5c9/dask_sql/mappings.py#L79-L83

sarahyurick: Ok, updated. Thanks!

ayushdg (approved):

PR #840, opened 2022-10-06
Configure dependabot to ignore `arrow` and `datafusion`

charlesbluca:
Makes it so new releases of `arrow` or `datafusion` will not trigger a dependabot PR, as we intend to update these dependencies manually

### DIFF
--- a/.github/dependabot.yml
+++ b/.github/dependabot.yml
@@ -7,4 +7,6 @@ updates:
     ignore:
       # arrow and datafusion are bumped manually
       - dependency-name: "arrow"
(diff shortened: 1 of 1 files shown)


.github/dependabot.yml
```diff
@@ -4,3 +4,7 @@ updates:
     directory: "/dask_planner"
     schedule:
       interval: "daily"
+    ignore:
+      # arrow and datafusion are bumped manually
+      - dependency-name: "arrow"
```
charlesbluca, line 9: Are there any updates we would still want dependabot to look out for (i.e. patch version updates)?

PR #843, opened 2022-10-06
Unxfailing passing queries

randerzander:
Enabling more queries in test suite


charlesbluca:
Were these queries passing locally for you @randerzander? I've set `xfail_strict=true` so that xpassing tests should be counted as failures, but it is possible that the set of queries passing on CPU and GPU are different.

PR #854, opened 2022-10-11
Parse and use schema for plugins whenever possible

### LINKED ISSUE #841: [ENH] Generalize all custom plugins to accept a schema name
Currently, many of the custom plugins (such as `CREATE MODEL`) are not supported for non-default schemas, operating directly on `context.schema_name`.

It would be nice if all custom parsing functions in parser.rs parsed the schema name (if provided) for use in the Python plugins, using either `elements_from_tablefactor` or some other helper function.

An alternative is to use `USE SCHEMA` statements to change the underlying `context.schema_name`; however, this doesn't necessarily resolve the potential user confusion around the mixed support of schema names.

charlesbluca:
Generalizes plugins to accept and use a schema name whenever possible, instead of the currently set schema.

Closes #841


dask_planner/src/parser.rs
```diff
@@ -1174,14 +1182,23 @@ impl<'a> DaskParser<'a> {
...
+            table_name,
+        })))
+    }
+
+    /// Parse Dask-SQL SHOW MODEL [FROM <schema>]
+    fn parse_show_models(&mut self) -> Result<DaskStatement, ParserError> {
+        let mut schema_name: Option<String> = None;
+        if !self.parser.consume_token(&Token::EOF) {
+            self.parser.expect_keyword(Keyword::FROM)?;
+            schema_name = Some(self.parser.parse_identifier()?.value);
```
andygrove, line 1198: nit: it might be good to add a check for an expected EOF after the schema name

andygrove (approved):
Rust changes LGTM

dask_planner/src/parser.rs
andygrove: I guess this comment applies to all cases where we are taking over parsing. If we don't check for EOF after parsing then we are potentially silently ignoring any extra tokens.

charlesbluca: Yeah that's a good point - considering the diff of this PR is already pretty large, are you okay with me opening an issue to track this and we can address it in a follow up?

andygrove: Yes, absolutely.

charlesbluca: It looks like maybe a check for the end of statement is being done implicitly? Trying out the following with this branch:

```python
from dask_sql import Context

c = Context()

c.sql("show models from 'root' extra tokens")
```

I get a `ParserError("Expected end of statement, found: extra")`.

dask_planner/src/sql/logical/create_model.rs
```diff
@@ -80,6 +82,11 @@ impl PyCreateModel {
         Ok(self.create_model.input.clone().into())
     }
 
+    #[pyo3(name = "getSchemaName")]
+    fn get_schema_name(&self) -> PyResult<Option<String>> {
+        Ok(self.create_model.schema_name.clone())
+    }
+
```
ayushdg, line 89: I think this is not possible today, but do you happen to know if there's a way we could have a common utility or template that could be share across classes. 

I'm assuming it might not be straightforward since these are all pyo3 classes and the variable name (`create_model` and so on) wouldn't match the class name of the caller.

ayushdg (approved):
Thanks @charlesbluca changes lgtm and this significantly improves our handling of schema in a few places. 

I have a couple of comments inline but nothing that should be blocking

dask_planner/src/sql/table.rs
```diff
@@ -212,8 +215,8 @@ pub(crate) fn table_from_logical_plan(
             }
 
             Ok(Some(DaskTable {
-                schema: String::from("EmptySchema"),
-                name: String::from("EmptyRelation"),
+                schema_name: Some(String::from("EmptySchema")),
```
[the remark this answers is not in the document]
charlesbluca: I don't think it should have much impact? Looking through the codebase, it seems like the only place this information is directly used on the Python end is through `context.fqn`:

https://github.com/dask-contrib/dask-sql/blob/2a8e9582de7b9a2519244796574054f1866222c7/dask_sql/context.py#L719

Which in turn only ends up getting used to grab the fully qualified table name for the table scan plugin:

https://github.com/dask-contrib/dask-sql/blob/2a8e9582de7b9a2519244796574054f1866222c7/dask_sql/physical/rel/logical/table_scan.py#L44

Given this knowledge, I'm interested in if we could refactor `table_scan.rs` to grab the schema name directly and remove these Python bindings altogether.

dask_planner/src/sql/logical/create_model.rs
charlesbluca: Would need to do some digging to see (cc @andygrove in case there's an obvious solution I'm missing here), but as long as the attribute the schema name is stored under is consistent we could probably consolidate this somewhat with a macro?

dask_sql/physical/rel/custom/create_memory_table.py
```diff
@@ -38,7 +38,11 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
         # Rust create_memory_table instance handle
         create_memory_table = rel.create_memory_table()
 
-        schema_name, table_name = context.schema_name, create_memory_table.getName()
+        # TODO: can we avoid hardcoding the schema name?
+        schema_name, table_name = (
+            context.schema_name,
+            create_memory_table.getTableName(),
```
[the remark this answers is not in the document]
andygrove: I reviewed `LogicalPlan.schema(),` and it looks like the implementation for `CreateMemoryTable` is incorrect because it returns the schema of the input rather than the schema of the newly created table. I recommend filing a bug against DataFusion for this.

dask_planner/src/sql/table.rs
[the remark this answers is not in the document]
ayushdg: Might be worth filing in issue to explore further and leaving it outside the scope of this pr

dask_sql/physical/rel/custom/drop_table.py
```diff
@@ -25,13 +25,23 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
+            raise RuntimeError(
+                f"Expected unqualified or fully qualified table name, got {qualified_table_name}."
+            )
+
+        schema_name = context.schema_name if not schema_name else schema_name[0]
+
+        if (
+            schema_name not in context.schema
+            or table_name not in context.schema[schema_name].tables
+        ):
```
charlesbluca, line 41: Ended up going with roughly the same Python schema name parsing for DropTable and CreateMemoryTable since we're able to pass through a qualified table name as input.

dask_sql/physical/rel/custom/create_memory_table.py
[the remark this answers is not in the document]
charlesbluca: Thanks for looking into this @andygrove! Opened https://github.com/apache/arrow-datafusion/issues/5004 to track this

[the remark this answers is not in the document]
qrilka: @andygrove I was looking into apache/arrow-datafusion#5004 but I don't quite understand the motivation here: the difference between the schema of the input and the schema of the newly create table is only the table name. Is the idea to make it override the qualifiers of `DFField`s from the input?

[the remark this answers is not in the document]
ayushdg: cc: @charlesbluca

PR #857, opened 2022-10-12
Add parser support for `timestampadd`

andygrove:
DataFusion's parser cannot parse `timestampadd(year, 2, d)` because `year` is not a keyword, so it assumes that this is an identifier and looks for a column with this name.

This PR works around the issue by providing a custom prefix parser in the `DaskSqlDialect` where we can override parsing of just this expression and then fall back to DataFusion. This is a new feature that was added to sqlparser relatively recently.

### DIFF
--- a/dask_planner/src/dialect.rs
+++ b/dask_planner/src/dialect.rs
@@ -1,6 +1,11 @@
 use core::iter::Peekable;
 use core::str::Chars;
+use datafusion_sql::sqlparser::ast::{
+    Expr, Function, FunctionArg, FunctionArgExpr, Ident, ObjectName, Value,
+};
 use datafusion_sql::sqlparser::dialect::Dialect;
+use datafusion_sql::sqlparser::parser::{Parser, ParserError};
+use datafusion_sql::sqlparser::tokenizer::Token;
 
 #[derive(Debug)]
 pub struct DaskDialect {}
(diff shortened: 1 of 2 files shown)


dask_planner/src/dialect.rs
```diff
@@ -37,4 +42,46 @@ impl Dialect for DaskDialect {
...
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(n)),
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(expr)),
+                    ];
+
+                    Ok(Some(Expr::Function(Function {
+                        name: ObjectName(vec![Ident::new("TIMESTAMPADD")]),
+                        args,
+                        over: None,
+                        distinct: false,
+                        special: false,
```
sarahyurick, line 75: This is so cool! Definitely more convenient than having to go through SQLParser and DataFusion, I can already think of a couple more functions that I'll do this for.

dask_planner/src/dialect.rs
```diff
@@ -37,4 +42,46 @@ impl Dialect for DaskDialect {
...
+                    let args = vec![
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(Expr::Value(
+                            Value::SingleQuotedString(time_unit.to_string()),
+                        ))),
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(n)),
+                        FunctionArg::Unnamed(FunctionArgExpr::Expr(expr)),
+                    ];
+
+                    Ok(Some(Expr::Function(Function {
+                        name: ObjectName(vec![Ident::new("TIMESTAMPADD")]),
```
charlesbluca, line 71: For sake of consistency with other custom functions, could we give this an all lowercase name?

```suggestion
                        name: ObjectName(vec![Ident::new("timestampadd")]),
```

andygrove:
Closing this now that these changes are part of https://github.com/dask-contrib/dask-sql/pull/838

PR #860, opened 2022-10-12
Check for `np.timedelta64` in `as_timelike`

sarahyurick:
When I try:

```
from dask_sql import Context
import pandas as pd
import dask.dataframe as dd
from datetime import datetime

c = Context()

date = datetime(2021, 10, 3, 15, 53, 42, 47)
date2 = datetime(2021, 2, 3, 15, 53, 42, 47)

df = dd.from_pandas(pd.DataFrame({"d": [date, date2]}), npartitions=1)
c.register_dask_table(df, "df")

c.sql(
"""
SELECT d + INTERVAL '5 days' FROM df
"""
)
```

It fails with a `Don't know how to make <class 'numpy.timedelta64'> timelike`. This is because we are adding 5 days (432000000 milliseconds) but not checking if it's a `np.timedelta64` type (in which case, it's already in the desired format and can be returned as is). This small PR fixes the bug.

### DIFF
--- a/dask_sql/physical/rex/core/call.py
+++ b/dask_sql/physical/rex/core/call.py
@@ -44,9 +44,7 @@ def as_timelike(op):
         return np.timedelta64(op, "D")
     elif isinstance(op, str):
         return np.datetime64(op)
-    elif pd.api.types.is_datetime64_dtype(op):
-        return op
-    elif isinstance(op, np.timedelta64):
+    elif pd.api.types.is_datetime64_dtype(op) or isinstance(op, np.timedelta64):
         return op
     else:
         raise ValueError(f"Don't know how to make {type(op)} timelike")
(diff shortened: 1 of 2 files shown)


charlesbluca (commented):
Thanks for the fix @sarahyurick 😄 could we add a test for this case?

dask_sql/physical/rex/core/call.py
```diff
@@ -46,6 +46,8 @@ def as_timelike(op):
         return np.datetime64(op)
     elif pd.api.types.is_datetime64_dtype(op):
         return op
+    elif isinstance(op, np.timedelta64):
+        return op
```
ayushdg, line 50: we can probably combine the two conditions since they return the same result.
```suggestion
    elif pd.api.types.is_datetime64_dtype(op) or isinstance(op, np.timedelta64):
        return op
```

ayushdg (approved):

charlesbluca (approved):

PR #867, opened 2022-10-18
Use `xfail` consistently for known failing tests

charlesbluca:
The intent of this PR is to make the way we indicate failing tests more consistent, by:

- uncommenting any tests that have been commented out in their entirety, xfailing with a reason if they fail
- uncommenting any sections of tests that are partially failing, splitting these sections off into separate tests if they fail

### DIFF
--- a/tests/integration/test_compatibility.py
+++ b/tests/integration/test_compatibility.py
@@ -594,72 +594,71 @@ def test_window_row_number_partition_by():
     )
 
 
-# TODO: Except not implemented so far
-# def test_window_ranks():
-#     a = make_rand_df(100, a=int, b=(float, 50), c=(str, 50))
-#     eq_sqlite(
-#         """
-#         SELECT *,
-#             RANK() OVER (PARTITION BY a ORDER BY b DESC NULLS FIRST, c) AS a1,
-#             DENSE_RANK() OVER (ORDER BY a ASC, b DESC NULLS LAST, c DESC) AS a2,
-#             PERCENT_RANK() OVER (ORDER BY a ASC, b ASC NULLS LAST, c) AS a4
-#         FROM a
-#         """,
-#         a=a,
-#     )
+def test_window_ranks():
+    a = make_rand_df(100, a=int, b=(float, 50), c=(str, 50))
+    eq_sqlite(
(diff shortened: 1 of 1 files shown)


continuous_integration/environment-3.10-dev.yaml
```diff
@@ -4,7 +4,8 @@ channels:
 - nodefaults
 dependencies:
 - dask>=2022.3.0
-- fastapi>=0.69.0
+# FIXME: handling is needed for httpx-based fastapi>=0.87.0
+- fastapi>=0.69.0,<0.87.0
```
charlesbluca, line 8: `fastapi>=0.87.0` pulls in `starlette>=0.21.0`, which switched to using `httpx` instead of `requests` for its `TestClient` API (https://github.com/encode/starlette/pull/1376); it's not immediately obvious to me how to support both versions of the API, so this pins us to the older `starlette` packages for now.

tests/integration/fixtures.py
```diff
@@ -332,15 +332,15 @@ def gpu_client(gpu_cluster):
 # client for all computations. otherwise, only connect to a client
 # when specified.
 @pytest.fixture(
-    scope="function" if SCHEDULER_ADDR is None else "session",
+    scope="function",
```
charlesbluca, line 335: When we use a session-wide client for external cluster testing, it ends up getting blown away with `dask.config.refresh()` while running `test_config.py`. Since there wasn't any immediate way to avoid this, I made this change so that we reconnect to the external cluster on each test, so that tests run after `test_config.py` are still able to use the external cluster as expected.

ayushdg (approved):
Thanks a lot for this overhaul of the pytests and moving over to marking everything as xfail and removing commented subparts of tests into their own xfailing test. This will help us catch bugs and issues much quicker.

tests/integration/fixtures.py
ayushdg: I also wonder if `test_config` can be improved somehow to save the state of configs before the tests start and restore that state at the end of the test. 

Either way I didn't see a noticeable change in overall test runtimes with this change so it might be okay to improve later.

PR #874, opened 2022-10-19
Install `twine` in `cibuildwheel` environment

charlesbluca:
It looks like the `twine` installed prior to running `cibuildwheel` is not picked up in the Mac builds, potentially due to the venv that `cibuildwheel` creates.

This PR attempts to resolve this by installing `twine` as part of the `cibuildwheel` builds.


charlesbluca:
Looks like even with the Python setup action, `twine` isn't getting installed properly on macOS - going to try using conda instead of pip to see if that resolves

.github/workflows/release.yml
```diff
@@ -2,13 +2,14 @@ name: Upload Python package
 on:
   release:
     types: [created]
+  pull_request:
```
ayushdg, line 5: Commenting as a reminder to remove before merging

ayushdg (approved):
Changes lgtm! Should be good to merge when tests pass (after removing the pr trigger)

PR #876, opened 2022-10-19
Add `TIMESTAMPDIFF` support

### LINKED ISSUE #852: [ENH] Support `TIMESTAMPDIFF`
After #813 is resolved, we should also add support for the `TIMESTAMPDIFF` function, as described [here](https://www.w3resource.com/mysql/date-and-time-functions/mysql-timestampdiff-function.php).

The implementation should look similar to that of `TIMESTAMPADD`. Previous support was added with #495, which we can grab the pytests from.

None.

**Additional context**
I plan to open a PR for this after #813 is merged.

sarahyurick:
Closes #852


dask_sql/physical/rex/core/call.py
```diff
@@ -634,6 +634,132 @@ def timestampadd(self, unit, interval, df: SeriesOrScalar):
...
+
+                day1 = df1[ind].dt.day.compute()[ind]
+                day2 = df2[ind].dt.day.compute()[ind]
+
+                if diff > 0 and day1 > day2:
+                    diff = diff - 1
+                elif diff < 0 and day2 > day1:
+                    diff = diff + 1
+
+                result.append(diff)
```
ayushdg, line 686: Trying to think of a way we can do this without compute. 
Is it possible to somehow divide the result by `np.timedelta64(1, "M")`.

Happy to discuss offline if this approach works

sarahyurick: Thanks! Initially, I was having rounding issues with months and quarters, (e.g., just doing # of days / 30 was not enough to produce an exactly correct result), but I've updated my implementation to divide by a more precise number (e.g., the average number of days in a month is `((30 * 4) + 28 + (31 * 7)) / 12`), which has removed the need for `compute()`.

For years, I initially had extra logic to deal with leap days, particularly the cases when we are comparing February 28/29 and/or March 1, which required `compute()` to extract the months and days to compare with each other. This is pretty nitpicky, though, so I removed that logic and just calculate the years with # of days / 365.

Let me know if the logic for these changes seems reasonable 🙂

ayushdg: The changes make sense to me. Even libraries like numpy/mysql have differing behavior here and diff between months/leap years are handled differently in many cases.

ayushdg (approved):

ayushdg (approved):

PR #885, opened 2022-10-24
Replace `variadic` with `exact` where appropriate

### LINKED ISSUE #868: Use appropriate `TypeSignature` variants in custom function signatures
In [dask_planner/src/sql.rs](https://github.com/dask-contrib/dask-sql/blob/main/dask_planner/src/sql.rs), we define several custom function signatures ("year", "atan2", "every", "bit_and", etc.) by using `Signature::variadic`. This may not always be appropriate, as `Variadic` implies that we could have an arbitrary amount of the specified datatypes passed through as input.

For reference, the various signatures that we could use are found [here](https://docs.rs/datafusion-expr/13.0.0/datafusion_expr/struct.Signature.html). We should comb through all uses of `Signature::variadic` and replace them with different `TypeSignature` variants where appropriate.

sarahyurick:
Closes #868

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -148,12 +148,12 @@ impl ContextProvider for DaskSQLContext {
 
         match name {
             "year" => {
-                let sig = Signature::variadic(vec![DataType::Int64], Volatility::Immutable);
+                let sig = Signature::exact(vec![DataType::Int64], Volatility::Immutable);
                 let rtf: ReturnTypeFunction = Arc::new(|_| Ok(Arc::new(DataType::Int64)));
                 return Some(Arc::new(ScalarUDF::new(name, &sig, &rtf, &fun)));
             }
             "atan2" | "mod" => {
-                let sig = Signature::variadic(
+                let sig = Signature::exact(
                     vec![DataType::Float64, DataType::Float64],
                     Volatility::Immutable,
                 );
@@ -161,7 +161,7 @@ impl ContextProvider for DaskSQLContext {
                 return Some(Arc::new(ScalarUDF::new(name, &sig, &rtf, &fun)));
             }
             "cbrt" | "cot" | "degrees" | "radians" | "sign" | "truncate" => {
-                let sig = Signature::variadic(vec![DataType::Float64], Volatility::Immutable);
+                let sig = Signature::exact(vec![DataType::Float64], Volatility::Immutable);
                 let rtf: ReturnTypeFunction = Arc::new(|_| Ok(Arc::new(DataType::Float64)));
                 return Some(Arc::new(ScalarUDF::new(name, &sig, &rtf, &fun)));
             }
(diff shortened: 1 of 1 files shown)


sarahyurick:
The only custom functions I keep as `variadic` are `rand` and `rand_integer` because `rand` doesn't necessarily need an input and `rand_integer` doesn't necessarily need 2 inputs, as demonstrated here: https://github.com/dask-contrib/dask-sql/blob/e104cc5e0e85e3cce9dff5dd2d445bbebb723521/tests/integration/test_rex.py#L164

dask_planner/src/sql.rs
```diff
@@ -148,20 +148,20 @@ impl ContextProvider for DaskSQLContext {
 
         match name {
             "year" => {
-                let sig = Signature::variadic(vec![DataType::Int64], Volatility::Immutable);
+                let sig = Signature::exact(vec![DataType::Int64], Volatility::Immutable);
```
charlesbluca, line 151: Interested in how this would work out with non `Int64` integers - I think we would get some warnings raised by the optimizer, but ultimately not end up doing extra casting operations on the Python side because the types are considered similar by `similar_type`.

Might be worth opening an issue to see if we can suppress optimizer warnings by default, as they come up quite a lot during normal use cases.

dask_planner/src/sql.rs
```diff
@@ -148,20 +148,20 @@ impl ContextProvider for DaskSQLContext {
...
                 let rtf: ReturnTypeFunction = Arc::new(|_| Ok(Arc::new(DataType::Int64)));
                 return Some(Arc::new(ScalarUDF::new(name, &sig, &rtf, &fun)));
             }
             "atan2" | "mod" => {
-                let sig = Signature::variadic(
+                let sig = Signature::exact(
                     vec![DataType::Float64, DataType::Float64],
```
charlesbluca, line 157: Maybe we should allow `Int64` as input here? If we don't, we end up forcing an unnecessary cast to floats:

```python
In [10]: c.explain("select mod(10, 30)")
WARNING:dask_sql.context:No tables are registered.
Skipping optimizer rule simplify_expressions due to unexpected error: This feature is not implemented: 
Skipping optimizer rule simplify_expressions due to unexpected error: This feature is not implemented: 
Skipping optimizer rule simplify_expressions due to unexpected error: This feature is not implemented: 
Out[10]: 'Projection: mod(CAST(Int64(10) AS Float64), CAST(Int64(30) AS Float64))\n  EmptyRelation'
```

This illustrates some of the difficulty I had when trying to switch away from `variadic`, as essentially, we would want to allow any length-2 combination of floats and ints, which can get cumbersome to write out - some workarounds could be making a utility function that can generate a `TypeSignature::OneOf` given some parameters, or maybe even the introduction of a new `TypeSignature` variant similar to `Variadic`, but with a fixed length.

Also, trying out `atan2` I notice that it is seemingly being handled implicitly on the DataFusion end, which seems to imply that we no longer need to define it here:

```python
In [11]: c.explain("select atan2(10, 30)")
Out[11]: 'Projection: Float32(0.32175055) AS atan2(Int64(10),Int64(30))\n  EmptyRelation'
```

Doesn't need to be done in this PR, but might be worthwhile going through the custom UDF/UDAFs to see if any have been implemented upstream and can be removed.

charlesbluca (commented):
Thanks for getting this started @sarahyurick 😄 a couple comments

Wonder if for `rand` and `rand_integer` it would make more sense to use `Signature::one_of` since we know the signature should either take 0/1 and 1/2 int arguments, respectively?

dask_planner/src/sql.rs
sarahyurick: I like the idea of making a utility function that generates all combinations of type inputs! I'll try and get it working.

I can also check out which custom functions are no longer needed.

dask_planner/src/sql.rs
```diff
@@ -576,3 +571,55 @@ impl PlanVisitor for OptimizablePlanVisitor {
         Ok(true)
     }
 }
+
+fn generate_numeric_signatures(n: i32) -> Signature {
```
charlesbluca, line 575: This works well for uniform signatures that accept int/float inputs; however, as we add in timestamp functions, it would be nice to have a utility like this be flexible for an arbitrary list of valid input types.

I was thinking maybe a function that accepts a vector of vectors of `DataTypes` and returns the corresponding cartesian product, i.e. something like:

```rust
let sig = generate_signatures(vec![
    vec![DataType::Int64, DataType::Float64],
    vec![DataType::Utf8, DataType::Int64]
]);
// would be equivalent to
// let sig = Signature::one_of(
//     vec![
//         TypeSignature::Exact(vec![DataType::Int64, DataType::Utf8]),
//         TypeSignature::Exact(vec![DataType::Int64, DataType::Int64]),
//         TypeSignature::Exact(vec![DataType::Float64, DataType::Utf8]),
//         TypeSignature::Exact(vec![DataType::Float64, DataType::Int64]),
//     ],
//     Volatility::Immutable,
// );
```

dask_planner/src/sql.rs
```diff
@@ -576,3 +571,112 @@ impl PlanVisitor for OptimizablePlanVisitor {
...
+    let mut one_of_vector = vec![];
+    for vector in exact_vector.iter() {
+        one_of_vector.push(TypeSignature::Exact(vector.clone()));
+    }
+
+    Signature::one_of(one_of_vector.clone(), Volatility::Immutable)
+}
+
+#[allow(dead_code)]
+fn generate_signatures(cartesian_setup: Vec<Vec<DataType>>) -> Signature {
```
sarahyurick, line 628: I've marked this with `#[allow(dead_code)]` since it will be used with #813, #838, #876, which haven't been merged yet.

charlesbluca: Wonder if `generate_numeric_signatures` could be refactored to use this function to generate the list of numeric signatures?

charlesbluca (approved):
Thanks @sarahyurick this LGTM; one small comment but can be handled in a follow up

PR #903, opened 2022-11-03
Upgrade to DataFusion 14.0.0

andygrove:
DataFusion 14.0.0 is due to be released next week, so this PR will update to the release when it is available.

### DIFF
--- a/dask_planner/Cargo.lock
+++ b/dask_planner/Cargo.lock
@@ -351,9 +351,10 @@ dependencies = [
 [[package]]
 name = "datafusion-common"
 version = "13.0.0"
-source = "git+https://github.com/apache/arrow-datafusion/?rev=54d2870a56d8d8f914a617a7e2d52e387ef5dba2#54d2870a56d8d8f914a617a7e2d52e387ef5dba2"
+source = "git+https://github.com/apache/arrow-datafusion/?rev=c9442ce6ecf7efeb238c6c25c0519350d0ba3839#c9442ce6ecf7efeb238c6c25c0519350d0ba3839"
 dependencies = [
  "arrow",
+ "chrono",
  "ordered-float",
  "sqlparser",
 ]
@@ -361,7 +362,7 @@ dependencies = [
 [[package]]
 name = "datafusion-expr"
 version = "13.0.0"
-source = "git+https://github.com/apache/arrow-datafusion/?rev=54d2870a56d8d8f914a617a7e2d52e387ef5dba2#54d2870a56d8d8f914a617a7e2d52e387ef5dba2"
+source = "git+https://github.com/apache/arrow-datafusion/?rev=c9442ce6ecf7efeb238c6c25c0519350d0ba3839#c9442ce6ecf7efeb238c6c25c0519350d0ba3839"
 dependencies = [
  "ahash 0.8.0",
  "arrow",
@@ -373,7 +374,7 @@ dependencies = [
 [[package]]
 name = "datafusion-optimizer"
 version = "13.0.0"
-source = "git+https://github.com/apache/arrow-datafusion/?rev=54d2870a56d8d8f914a617a7e2d52e387ef5dba2#54d2870a56d8d8f914a617a7e2d52e387ef5dba2"
+source = "git+https://github.com/apache/arrow-datafusion/?rev=c9442ce6ecf7efeb238c6c25c0519350d0ba3839#c9442ce6ecf7efeb238c6c25c0519350d0ba3839"
 dependencies = [
  "arrow",
  "async-trait",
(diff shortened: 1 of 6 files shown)


andygrove:
There are regressions with window functions, Example:

```
>       sort_expressions = rel.window().getSortExprs(window)
E       TypeError: "Provided Expr SUM(a.b + a.aa.aa.b AS a.b + a.a) PARTITION BY [a.b] ORDER BY [a.a ASC NULLS FIRST] ROWS BETWEEN NULL PRECEDING AND CURRENT ROW AS SUM(a.b + a.a) PARTITION BY [a.b] ORDER BY [a.a ASC NULLS FIRST] ROWS BETWEEN UNBOUNDED PRECEDING AND CURRENT ROW is not a WindowFunction type"
```

ayushdg:
Some of the failures here come from the fact that with datafusion 14.0 the plans generated are slightly different leading to the addition of an additional `ddf.a=1` step before the filters are applied:
```python
SELECT a FROM parquet_ddf WHERE (b > 5 AND b < 10) OR a = 1

# dask-sql main (datafusion rev)
Projection: parquet_ddf.a, parquet_ddf.b, parquet_ddf.c, parquet_ddf.d
  Filter: parquet_ddf.b > Int64(5) AND parquet_ddf.b < Int64(10) OR parquet_ddf.a = Int64(1)
    TableScan: parquet_ddf projection=[a, b, c, d]

# df 14.0
Projection: parquet_ddf.a, parquet_ddf.b, parquet_ddf.c, parquet_ddf.d
  Filter: (parquet_ddf.b > Int64(5) OR parquet_ddf.a = Int64(1)Int64(1)parquet_ddf.a) AND (parquet_ddf.b < Int64(10) OR parquet_ddf.a = Int64(1)Int64(1)parquet_ddf.a)
    Projection: parquet_ddf.a = Int64(1) AS parquet_ddf.a = Int64(1)Int64(1)parquet_ddf.a, parquet_ddf.a, parquet_ddf.b, parquet_ddf.c, parquet_ddf.d
      TableScan: parquet_ddf projection=[a, b, c, d]
```
In this case it's safe to push down the filter to the IO since the `df.a=1` op is selecting the same value as the filter. In a more arbitrary case if we happened to have `df.a=other_val` it would not be safe to push the filter of `df.a=1` down to the IO. 

The way dask handles it today is not by looking at the val, but generally allowing a subset of operations (irrespective of the values) to appear between the IO and filter stage to push predicates down.

ayushdg:
Looking into this a bit more the step that introduces this additional projection after the table scan comes from the `common_sub_expression_eliminate` rule. I'm not sure if this is a case where `a=1` is a common sub expression so it might be erroneously being applied? 

Also while looking through this I realized that the `filter_push_down` optimizer rule might be able to push these filter down to the table scan for us and allow to pass these filter into the IO. I don't recall if there was a specific reason this rule was not added at the time but it might be worth exploring re-adding this rule? 

cc: @andygrove

andygrove:
Thanks @ayushdg. I am going to work on this today. I have updated this PR to use the official 14.0.0 release of DataFusion now.

ayushdg:
> I don't understand this failure with Python 3.8 / mac:
> 
> ```
> FAILED tests/integration/test_fugue.py::test_fsql - AssertionError: DataFrame are different
> 
> DataFrame shape mismatch
> [left]:  (0, 1)
> [right]: (1, 1)
> ```
> 
> @ayushdg any ideas?

This is a known flaky test that appears occasionally on Mac and windows 3.8. 

It should be safe to ignore for now

ayushdg:
A bunch of gpu tests seem to be failing though. @charlesbluca Could you take a look if you get the chance?

charlesbluca:
Yeah can take a look into this

charlesbluca:
From a quick glance, it looks like q4, q11, and q74 are failing because we are trying to concat a dask-cudf and dask CPU dataframe together; the underlying cause for this is that new optimizations mean we are beginning to use the `EmptyRelation` plugin more prominently, and it always creates a CPU dataframe:

https://github.com/dask-contrib/dask-sql/blob/9f97cc7413c1013217b89417c9d1be9d4e0226cc/dask_sql/physical/rel/logical/empty.py#L32-L35

q33, q56, q60, and q83 all seem to be failing when attempting to regenerate the HLG with predicate pushdown, which is a little harder to diagnose - will look into those queries more closely.

dask_sql/physical/utils/filter.py
```diff
@@ -91,7 +91,7 @@ def attempt_predicate_pushdown(ddf: dd.DataFrame) -> dd.DataFrame:
     try:
         return dsk.layers[name]._regenerate_collection(
             dsk,
-            new_kwargs={io_layer: {"filters": filters}},
+            new_kwargs={io_layer: {"filters": filters, "index": False}},
```
charlesbluca, line 94: Looks like the issues with predicate pushdown were stemming from the automatic setting of an index in `read_parquet` by default, which this kwarg override should disallow.

Chatting with @rjzamora, we agreed that this shouldn't be the default behavior, so we may be able to remove this override later on when changes are made upstream.

charlesbluca (approved):
Thanks @andygrove! We can iterate on the optimizer changes once #908 and #914 are resolved

PR #904, opened 2022-11-04
Add support for `ALTER TABLE` and `ALTER SCHEMA`

### LINKED ISSUE #884: [ENH] Add back support for `ALTER TABLE | SCHEMA`
Currently, `test_alter_schema` and `test_alter_schema` are failing because we haven't yet implemented parser support for `ALTER TABLE | SCHEMA`.

We should add in parsing support for `ALTER TABLE | SCHEMA`.

It looks like sqlparser is already able to parse `ALTER TABLE ... RENAME TO ...` statements, so we might just need to implement a corresponding plan node for this statement. On the other hand, `ALTER SCHEMA` is not supported, so we would need to add in custom parsing logic for those statements.

charlesbluca:
Closes #884

### DIFF
--- a/tests/unit/test_context.py
+++ b/tests/unit/test_context.py
@@ -412,5 +412,4 @@ def test_alter_table(c, df_simple):
 
     c.sql("ALTER TABLE IF EXISTS alien RENAME TO humans")
 
-    print(c.schema[c.schema_name].tables)
     del c.schema[c.schema_name].tables["physics"]
--- a/tests/unit/test_context.py
+++ b/tests/unit/test_context.py
@@ -391,28 +391,26 @@ def test_aggregation_adding():
     assert c.schema[c.schema_name].function_lists[1].aggregation
 
 
-# TODO: Alter schema is not yet implemented
-# def test_alter_schema(c):
-#     c.create_schema("test_schema")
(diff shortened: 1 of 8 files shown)


tests/unit/test_context.py
```diff
@@ -391,28 +391,25 @@ def test_aggregation_adding():
...
 
-# TODO: Alter table is not yet implemented
-# def test_alter_table(c, df_simple):
-#     c.create_table("maths", df_simple)
-#     c.sql("ALTER TABLE maths RENAME TO physics")
-#     assert "physics" in c.schema[c.schema_name].tables
+def test_alter_table(c, df_simple):
+    c.create_table("maths", df_simple)
+    c.sql("ALTER TABLE maths RENAME TO physics")
+    assert "physics" in c.schema[c.schema_name].tables
```
ayushdg, line 408: Do we also want to assert here than `maths` not in the tables list?

ayushdg (approved):
Changes lgtm! Thanks a lot @charlesbluca.

Minor comment around the tests.

tests/unit/test_context.py
charlesbluca: Yeah that makes sense, can make a similar alteration to `test_alter_schema`

PR #912, opened 2022-11-09
Remove `generate_numeric_signatures`

### LINKED ISSUE #902: Refactor `generate_numeric_signatures`
In https://github.com/dask-contrib/dask-sql/pull/885, we added the `generate_numeric_signatures` and `generate_signatures`, which both compute the Cartesian product. The difference is that the former assumes that our signature should consist of only numeric types (integers and floats) and thus has an integer input corresponding to the number of numeric fields we expect in the signature, while the latter allows the user to input a vector of vectors with which to compute the Cartesian product.

Since both are performing similar operations, just with different function inputs, we should look into finding a way to combine them.

sarahyurick:
Closes #902


ayushdg:
@charlesbluca Since you're slightly more familiar with this part of the codebase could you take a look at this pr when you get a chance?

charlesbluca (approved):
Thanks @sarahyurick this LGTM! One minor thing that could be worth changing in this PR is using `generate_signatures` to make the type signature for `dsql_totimestamp`, but okay with merging as is if that isn't possible

charlesbluca:
> Let me know whether you have a preference and I can add the changes if desired.

I think my preference would be to use `generate_signatures` here for sake of consistency, as it might cause some confusion seeing a single `one_of(exact(...))` type signature among the others that are generated using the util function.

charlesbluca:
The new clippy warnings seem to be unrelated and might just be a product of a newer toolchain getting pulled in? Regardless resolved things locally and pushed up, assuming CI is green we can go ahead and merge.

Thanks again @sarahyurick!

charlesbluca:
Seems like the gpuCI & RTD builds are stuck - given that they were passing in previous commits, I think this should be safe to merge in

PR #926, opened 2022-11-18
Clean up `wrap_predict` and `wrap_fit` flags

### LINKED ISSUE #909: Clean up `wrap_predict` and `wrap_fit` flags
Our `CreateModelPlugin` has two parameters `wrap_predict` and `wrap_fit` associated with the use of `ParallelPostFit` and `Incremental`, respectively.

As @VibhuJawa suggested, after #855 has been merged, we should clean up these arguments with better defaults.

sarahyurick:
Closes #909


tests/integration/test_model.py
```diff
@@ -527,7 +529,7 @@ def test_export_model(c, training_df, tmpdir):
     )
 
     assert (
-        pickle.load(open(str(temporary_file), "rb")).__class__.__name__
+        pickle.load(open(str(temporary_file), "rb")).estimator.__class__.__name__
```
sarahyurick, line 532: Since `wrap_predict` will default to true for this case, we have a `GradientBoostingClassifier` wrapped in a `ParallelPostFit`, so we have to grab the estimator.

dask_sql/physical/rel/custom/create_model.py
```diff
@@ -125,10 +125,35 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
+        wrap_predict = kwargs.pop("wrap_predict", None)
+        wrap_fit = kwargs.pop("wrap_fit", None)
         fit_kwargs = kwargs.pop("fit_kwargs", {})
 
+        try:
+            ModelClass = import_class(model_class)
+        except ImportError:
+            raise ValueError(
+                f"Can not import model {model_class}. Make sure you spelled it correctly and have installed all packages."
+            )
```
ayushdg, line 137: I realize this is a copy of the existing error we had here, but standardizing on raising the right error where possible.
```suggestion
            raise ImportError(
                f"Failed to import model {model_class}. Make sure it is spelled correctly and the relevant packages are installed."
            )
```

dask_sql/physical/rel/custom/create_model.py
```diff
@@ -125,10 +125,35 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
+                wrap_predict = False
+        if wrap_fit is None:
+            if (
+                "sklearn" in model_class
+                or ("cuml" in model_class and "cuml.dask" not in model_class)
+            ) and hasattr(model, "partial_fit"):
+                wrap_fit = True
+            else:
+                wrap_fit = False
+
```
ayushdg, line 156: Minor nit:
```suggestion
        if "sklearn" in model_class or (
            "cuml" in model_class and "cuml.dask" not in model_class
        ):
            wrap_predict = True if wrap_predict else False
            wrapt_fit = True if wrap_fit and hasattr(model, "partial_fit") else False
```

ayushdg (changes requested):
Minor suggestions but things generally lgtm

dask_sql/physical/rel/custom/create_model.py
sarahyurick: I think we still need the outer conditions `if wrap_predict is None` and `if wrap_fit is None`, otherwise if the user specifies either of these parameters then their specification will be overwritten. Let me know what you think?

ayushdg: Yeah you're right, we can scrap this suggestion.

ayushdg (approved):

PR #823, opened 2022-09-29
Implement basic COALESCE functionality

ChrisJar:

PR #951, opened 2022-12-02
Remove unnecessary `first` aggregation from groupby agg code

charlesbluca:
Right now, if group columns are specified for a groupby aggregation (e.g. `select name, sum(x) from df group by name`), we include a `first` aggregation to grab the relevant row values for each grouping. This shouldn't be necessary, as Dask groupby aggregations already provide this info in the resulting index.

This PR removes this `first` aggregation and retains the lost information by instead dropping the groupby agg index with `reset_index(drop=False)` when group columns are specified.

### DIFF
--- a/dask_sql/physical/rel/logical/aggregate.py
+++ b/dask_sql/physical/rel/logical/aggregate.py
@@ -229,18 +229,15 @@ class DaskAggregatePlugin(BaseRelPlugin):
             logger.debug("Performing full-table aggregation")
 
         # Do all aggregates
-        df_result, output_column_order, cc = self._do_aggregations(
+        df_agg, output_column_order, cc = self._do_aggregations(
             rel,
             dc,
             group_columns,
             context,
         )
 
-        # SQL does not care about the index, but we do not want to have any multiindices
-        df_agg = df_result.reset_index(drop=True)
-
-        # Fix the column names and the order of them, as this was messed with during the aggregations
-        df_agg.columns = df_agg.columns.get_level_values(-1)
+        # SQL does not care about the index, but if group columns were specified we'll want to keep those
+        df_agg = df_agg.reset_index(drop=(not group_columns))
 
         def try_get_backend_by_frontend_name(oc):
             try:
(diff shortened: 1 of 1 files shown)


dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -229,18 +229,15 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
             context,
         )
 
-        # SQL does not care about the index, but we do not want to have any multiindices
-        df_agg = df_result.reset_index(drop=True)
-
-        # Fix the column names and the order of them, as this was messed with during the aggregations
-        df_agg.columns = df_agg.columns.get_level_values(-1)
+        # SQL does not care about the index, but if group columns were specified we'll want to keep those
+        df_agg = df_agg.reset_index(drop=(not group_columns))
```
ayushdg, line 240: In cases where we have multiple groupby columns but are selecting only 1 does it make sense to handle that logic of dropping the rest here or do we rely on the `limit_to` projection mechanisms downstream to achieve this?

charlesbluca: Yeah right now we're relying on `limit_to` to exclude the unnecessary columns, not sure if it's worth it to manually drop these columns after the `reset_index`

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -526,8 +514,6 @@ def _perform_aggregation(
                 logger.debug(f"Using original output_col value of '{output_col}'")
 
             aggregations_dict[input_col][output_col] = aggregation_f
-            if not isinstance(aggregation_f, str):
-                fast_groupby = False
```
charlesbluca, line 530: I initially included this because we were seeing groupby test failures when using `dd.Aggregations` without our custom null sorting, but it seems like this is no longer the case?

ayushdg:
I tested this on a sample dataset:
```python
df = timeseries(freq="10ms")
c.create_table("df",df, persist=True)
 %timeit c.sql("SELECT id, SUM(x) from df GROUP BY id").compute()

dask_sql/physical/rel/logical/aggregate.py
ayushdg: I'm assuming that there are tests with nulls checking for dd.Aggregation type of aggregations but it might be worth verifying before adding this in.

charlesbluca: We actually have a lot more exposure for `dd.Aggregations` now with https://github.com/dask-contrib/dask-sql/pull/823 since that introduced a `custom_sum` aggregation in the place of standard `sum` aggregations on CPU; one test that is using this on a dataframe with nulls is `test_single_agg_count_no_group_by` (along with a few other tests in test_compatibility.py).

ayushdg (approved):

PR #956, opened 2022-12-05
`FilterColumnsPostJoin` Optimizer Rule

### LINKED ISSUE #898: [ENH] Optimizer rule to drop join key columns post join
Many queries follow a pattern where certain columns are only used as join keys and aren't really needed/selected in the query post join. In cases such as these it might be useful to be able to drop those columns post join rather than carrying them around for subsequent ops and only dropping them at the end during the projection.

Add an optimizer rule that either adds a projection step post join or adds some sort of column filter post join to only keep the needed columns. 

**Additional context**
Add any other context, code examples, or references to existing implementations about the feature request here.

Here are some example cases which showcase the issue:

```python
print(c.explain("SELECT SUM(df.a),df2.b FROM df INNER JOIN df2 ON df.c=df2.c GROUP BY df2.b"))
Projection: SUM(df.a), df2.b
  Aggregate: groupBy=[[df2.b]], aggr=[[SUM(df.a)]]
    Inner Join: df.c = df2.c
      Filter: df.c IS NOT NULL
        TableScan: df projection=[a, c], partial_filters=[df.c IS NOT NULL]
      Filter: df2.c IS NOT NULL
        TableScan: df2 projection=[b, c], partial_filters=[df2.c IS NOT NULL]
```
Here we carry `column c` into the aggregate step even though it can be dropped

sarahyurick:
Closes #898

### DIFF
--- a/dask_planner/src/sql/optimizer/filter_columns_post_join.rs
+++ b/dask_planner/src/sql/optimizer/filter_columns_post_join.rs
@@ -1,0 +1,182 @@
+//! Optimizer rule dropping join key columns post join
+//!
+//! Many queries follow a pattern where certain columns are only used as join keys and aren't
+//! really needed/selected in the query post join. In cases such as these it is useful to drop
+//! those columns post join rather than carrying them around for subsequent operations and only
+//! dropping them at the end during the projection.
+//!
+//! A query like
+//! ```text
+//! SELECT
+//!     SUM(df.a), df2.b
+//! FROM df
+//! INNER JOIN df2
+//!     ON df.c = df2.c
+//! GROUP BY df2.b
+//! ```
+//!
+//! Would typically produce a LogicalPlan like ...
+//! ```text
+//! Projection: SUM(df.a), df2.b\
+//!   Aggregate: groupBy=[[df2.b]], aggr=[[SUM(df.a)]]\
+//!     Inner Join: df.c = df2.c\
+//!       TableScan: df projection=[a, c], full_filters=[df.c IS NOT NULL]\
+//!       TableScan: df2 projection=[b, c], full_filters=[df2.c IS NOT NULL]\
+//! ```
+//!
+//! Where df.c and df2.c would be unnecessarily carried into the aggregate step even though it can
+//! be dropped.
+//! 
+//! To solve this problem, we insert a projection after the join step. In our example, the
+//! optimized LogicalPlan is
+//! ```text
+//! Projection: SUM(df.a), df2.b\
+//!   Aggregate: groupBy=[[df2.b]], aggr=[[SUM(df.a)]]\
+//!     Projection: df.a, df2.b\
+//!       Inner Join: df.c = df2.c\
+//!         TableScan: df projection=[a, c], full_filters=[df.c IS NOT NULL]\
+//!         TableScan: df2 projection=[b, c], full_filters=[df2.c IS NOT NULL]\
+//! ```
+//! 
+//! Which corresponds to rewriting the query as
+//! ```text
+//! SELECT
+//!     SUM(df.a), df2.b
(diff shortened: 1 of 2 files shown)


sarahyurick:
We can test whether or not the rule is working as intended by adding a `print(cc._frontend_columns)` on https://github.com/dask-contrib/dask-sql/blob/main/dask_sql/physical/rel/logical/aggregate.py#L211 (after `cc = dc.column_container`).

By using the optimization rule with the example
```
SELECT SUM(df.a),df2.b FROM df INNER JOIN df2 ON df.c=df2.c GROUP BY df2.b
```
we print `['df2.b', 'df.a']`, whereas before (without the optimization rule introduced in this PR) `df.c` and `df2.c` would be in the list as not being dropped.

dask_planner/src/sql/optimizer/filter_columns_post_join.rs
```diff
@@ -0,0 +1,678 @@
...
+        let mut next_step;
+        let mut return_plan = plan.clone();
+
+        for key in (0..counter).rev() {
+            next_step = new_plan.get(&key).unwrap();
+            match next_step {
+                LogicalPlan::Projection(p) => {
+                    return_plan = LogicalPlan::Projection(Projection {
+                        expr: p.expr.clone(),
+                        input: Arc::new(previous_step.clone()),
```
sarahyurick, line 324: Not sure if there's a better way to rebuild the plan? Essentially we need to replace the `input` (an `Arc<LogicalPlan>`) of every `LogicalPlan` enum in the plan. So we're iterating through the plan and matching it with an enum, which requires checking if it is any of the possible enums (that have an `input` parameter) in https://github.com/apache/arrow-datafusion/blob/master/datafusion/expr/src/logical_plan/plan.rs#L48

dask_planner/src/sql/optimizer/filter_columns_post_join.rs
```diff
@@ -0,0 +1,678 @@
...
+        if let Some(exprs) = get_column_name(column) {
+            for expr in exprs {
+                post_join_columns.insert(expr);
+            }
+        }
+    }
+}
+
+fn get_column_name(column: &Expr) -> Option<Vec<Expr>> {
+    // TODO: Can we use to_columns() or expr_to_columns() here?
```
sarahyurick, line 475: Right now we use DataFusion 14.0.0, but once we update to 15.0.0 we might be able to simplify this function by just using https://github.com/apache/arrow-datafusion/blob/4538912db3c913220bfa0b251b6ab62842886145/datafusion/expr/src/expr.rs#L701

dask_planner/src/sql/optimizer/filter_columns_post_join.rs
sarahyurick: Still don't like this. It's essentially just 100 lines of code to check if there's an `input` that needs to be replaced.

jdye64 (approved):
@sarahyurick thank you for this PR. This logic is very beneficial for large queries that carry around un-necessary columns after a join. I feel your pain on the "boiler plate code" sections but as you pointed out that is just unfortunately what it takes right now. I plan to write some utility methods in the coming for improving, or at least centralizing, all of that boiler plate code so that future optimizers need not re-implment and hopefully speed up reviews as well.

I know others work is waiting on this so I'm going to approve for merging now and we can come back and address those things later. The PR does work so think it is worth merging now.

Thanks again!

sarahyurick:
Thanks @jdye64 !

jdye64 (approved):
LGTM

andygrove (approved):

ayushdg (approved):

charlesbluca:
Thanks @sarahyurick!

PR #855, opened 2022-10-11
Replace `dask_ml.wrappers.Incremental` with custom `Incremental` class

sarahyurick:
Closes #839.

In addition to #832, we want to create a custom implementation for Dask-ML's `Incremental` class as well.

So as not to create any merge conflicts, I've only added a single file relating to the scorers used in Dask-ML's implementation. After #832 I will add the remaining functionality and changes needed in `dask_sql/physical/rel/custom/create_model.py`, `dask_sql/physical/rel/custom/wrappers.py` (created in #832), and `docs/source/sql/ml.rst`.

@VibhuJawa

PR #957, opened 2022-12-05
Resolve ML cluster failures

charlesbluca:
It looks like something changed in upstream Dask such that we're now seeing issues when trying to use the local ML wrapper functions introduced in #832 and #855 in an independent cluster; this PR wraps these relevant function usages with `make_pickable_without_dask_sql` as is done with other functions passed directly off to Dask, and adds the `skip_if_external_scheduler` marker to relevant tests that would otherwise need scikit-learn.

cc @sarahyurick


sarahyurick:
Interesting, I wasn't familiar with `make_pickable_without_dask_sql` but makes sense to me! Thanks

Issue #943, opened 2022-11-28
[BUG] `experiment_class` fails on GPU


sarahyurick:
In #886, we removed all dependencies on Dask-ML in favor of scikit-learn, cuML, and our own classes (ParallelPostFit and Incremental). Previously, when creating an experiment, `experiment_class` was expected to be a path to a `dask_ml` class, but `sklearn` classes were also found to be compatible. However, I couldn't get it to work with `cuml`, such as with `cuml.model_selection.GridSearchCV`. For example:

```
c.sql(
    """
    CREATE EXPERIMENT my_exp WITH (
    model_class = 'sklearn.ensemble.GradientBoostingClassifier',
    experiment_class = 'cuml.model_selection.GridSearchCV',
    tune_parameters = (n_estimators = ARRAY [16, 32, 2],learning_rate = ARRAY [0.1,0.01,0.001],
                       max_depth = ARRAY [3,4,5,10]),
    target_column = 'target'
) AS (
        SELECT x, y, x*y > 0 AS target
        FROM timeseries
        LIMIT 100
    )
    """
)
```
errors with:
```
INFO:dask_sql.physical.rel.custom.create_experiment:{'n_estimators': [16, 32, 2], 'learning_rate': [0.1, 0.01, 0.001], 'max_depth': [3, 4, 5, 10]}
INFO:dask_sql.physical.rel.custom.create_experiment:{}
---------------------------------------------------------------------------
TypeError                                 Traceback (most recent call last)
Cell In [8], line 1
----> 1 c.sql(
      2     """
      3     CREATE EXPERIMENT my_exp WITH (
      4     model_class = 'sklearn.ensemble.GradientBoostingClassifier',
      5     experiment_class = 'cuml.model_selection.GridSearchCV',
      6     tune_parameters = (n_estimators = ARRAY [16, 32, 2],learning_rate = ARRAY [0.1,0.01,0.001],
      7                        max_depth = ARRAY [3,4,5,10]),
      8     target_column = 'target'
      9 ) AS (
     10         SELECT x, y, x*y > 0 AS target
     11         FROM timeseries
     12         LIMIT 100
     13     )
     14     """
     15 )

File ~/miniconda3/envs/dsql_rapids-22.12/lib/python3.9/site-packages/dask_sql/context.py:501, in Context.sql(self, sql, return_futures, dataframes, gpu, config_options)
    496 else:
    497     raise RuntimeError(
[...]

PR #962, opened 2022-12-07
Backend agnostic machine learning models

### LINKED ISSUE #370: [ENH] Backend agnostic machine learning models
As we are working becoming more backend agnostic (GPU/CPU) , we should look into a way of supporting multiple ML backends with minimal code change .

We currently have to specify the cuml/sklearn model class.  We should look into a way of training models where we detect what the input type is and have the user just specify ‘RandomForest’ and have dask-sql handle inferring the rest of the classname. 

So if the training dataframe is on CPU we use `sklearn` and if it is on GPU we use `cuML`.

sarahyurick:
Closes #370 

Most of the additions come from listing out all class mappings in `ml_classes.py` and adding various tests in `test_model.py`. Obviously, it would be too much to test all mappings, so I just included the classes already tested elsewhere.


dask_sql/physical/rel/custom/ml_classes.py
```diff
@@ -0,0 +1,381 @@
...
+        "MDS": "sklearn.manifold.MDS",
+        "SpectralEmbedding": "sklearn.manifold.SpectralEmbedding",
+        "TSNE": "sklearn.manifold.TSNE",
+        # sklearn.mixture: Gaussian Mixture Models
+        "BayesianGaussianMixture": "sklearn.mixture.BayesianGaussianMixture",
+        "GaussianMixture": "sklearn.mixture.GaussianMixture",
+        # sklearn.model_selection: Model Selection
+        "GroupKFold": "sklearn.model_selection.GroupKFold",
+        "GroupShuffleSplit": "sklearn.model_selection.GroupShuffleSplit",
+        "KFold": "sklearn.model_selection.KFold",
```
sarahyurick, line 188: There's probably a lot of mappings here that aren't even compatible with `model_class` or `experiment_class`.

I pretty much just listed every class available with sklearn and cuml and allow it to be mapped no matter what, since these are just string mappings and it has to error eventually anyways if it's not compatible with `CREATE MODEL`/`CREATE EXPERIMENT`. But let me know if there are chunks of mappings that I should just remove.

tests/integration/test_model.py
```diff
@@ -1024,3 +1024,266 @@ def test_predict_with_nullable_types(c):
...
+            LIMIT 100
+        )
+        """
+    )
+    check_trained_model(c, "my_exp")
+
+    c.sql(
+        """
+        CREATE MODEL IF NOT EXISTS my_model_lightgbm WITH (
+            model_class = 'LGBMClassifier',
```
sarahyurick, line 1139: LightGBM does not seem to support `dask_cudf` DataFrames. I actually get the same error `TypeError: Implicit conversion to a NumPy array is not allowed` as from https://github.com/dask-contrib/dask-sql/issues/943

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -147,6 +156,19 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
         y = training_df[target_column]
 
         if model_class and experiment_class:
+            if type(training_df) == dd.core.DataFrame:
+                if model_class in cpu_classes:
+                    model_class = cpu_classes[model_class]
+                if experiment_class in cpu_classes:
+                    experiment_class = cpu_classes[experiment_class]
```
charlesbluca, line 163: Think this (and other similar patterns) can be consolidated to

```suggestion
                model_class = cpu_classes.get(model_class, model_class)
                experiment_class = cpu_classes.get(experiment_class, experiment_class)
```

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -147,6 +156,19 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
...
+            elif (
+                dask_cudf is not None and type(training_df) == dask_cudf.core.DataFrame
```
charlesbluca, line 165: To avoid a dask-cudf import in this file, we could do:

```suggestion
                "cudf" in str(training_df._partition_type)
```

The motive here being that in the past, cuDF has failed to import on CPU machines where it was installed, which would then cause errors in dask-sql when we attempt to import it (xref #473)

tests/integration/test_model.py
```diff
@@ -1024,3 +1046,130 @@ def test_predict_with_nullable_types(c):
         result,
         check_dtype=False,
     )
+
+
+# TODO - many ML tests fail on clusters without sklearn - can we avoid this?
+@skip_if_external_scheduler
+@pytest.mark.skipif(
+    sys.version_info < (3, 9),
+    reason="Some newer sklearn classes are only available with Python version >= 3.9",
```
sarahyurick, line 1055: We use sklearn = 1.0.0 instead of sklearn >= 1.0.0 for our Python 3.8 integration tests:
https://github.com/dask-contrib/dask-sql/blob/main/continuous_integration/environment-3.8-dev.yaml#L30

tests/integration/test_model.py
```diff
@@ -1032,3 +1054,130 @@ def test_predict_with_nullable_types(c):
...
+
+    for key in classes_dict:
+        if not ("XGB" in key and xgboost is None) and not (
+            "LGBM" in key and lightgbm is None
+        ):
+            import_class(classes_dict[key])
+
+
+# TODO - many ML tests fail on clusters without sklearn - can we avoid this?
+@skip_if_external_scheduler
```
sarahyurick, line 1092:
```suggestion
@xfail_if_external_scheduler
```

tests/integration/test_model.py
```diff
@@ -1032,3 +1054,130 @@ def test_predict_with_nullable_types(c):
         result,
         check_dtype=False,
     )
+
+
+# TODO - many ML tests fail on clusters without sklearn - can we avoid this?
+@skip_if_external_scheduler
```
sarahyurick, line 1060:
```suggestion
@xfail_if_external_scheduler
```

tests/integration/test_model.py
```diff
@@ -1032,3 +1054,130 @@ def test_predict_with_nullable_types(c):
         result,
         check_dtype=False,
     )
+
+
+# TODO - many ML tests fail on clusters without sklearn - can we avoid this?
+@xfail_if_external_scheduler
```
sarahyurick, line 1060: Hi @charlesbluca I'm not sure why this test is failing on the Dask cluster job when `@xfail_if_external_scheduler` is here (it worked before with `@skip_if_external_scheduler`)?

ayushdg: looks like the test was xpassing. Can we try removing the xfail if and seeing if things changed?

sarahyurick: Ohh I see, thanks!

ayushdg: Looks like these tests are passing now. @charlesbluca Idk if we started including sklearn in the container or if some of the changes to re-use the fixture at a file level changed this. But things seem to work now

charlesbluca: This test is `test_ml_class_mappings`, right? If so I would expect things to pass here on an independent cluster since it seems like all the test is doing is importing the given classes on the client-side, which should have all the relevant ML libraries installed.

Was there a run when this test was failing in the cluster testing but passing elsewhere?

sarahyurick: I think originally I had all of the tests in a single method, and then when I broke them out into `test_ml_class_mappings`, `test_agnostic_cpu_xgb_models`, and `test_agnostic_gpu_xgb_models` I just kept the skip/xfail decorator on all of them. It took switching to xfail to catch the xpass 🙂

dask_sql/physical/rel/custom/ml_classes.py
charlesbluca: My preference would be to generate this mapping dynamically for a few reasons:

- as scikit-learn updates, it's very likely that many of these classes will get shuffled around, which will end up breaking CI until a PR is made to manually update things
- though this mapping currently works on the scikit-learn version getting pulled in our python>=3.9 testing, we don't actually know how many versions this mapping holds up for between 1.0.0 and 1.2.1

It looks like for scikit-learn, we should be able to do something like

```python
from sklearn.utils import all_estimators

estimators = {k: v.__module__ + "." + v.__qualname__ for k,v in all_estimators()}
```

To grab its estimators dynamically; I'm not sure if the other packages have something similar to this, but would be interested in if it's possible to generalize the code used for [`all_estimators`](https://github.com/scikit-learn/scikit-learn/blob/e2e705021eb6c9f23f0972f119b56e37cd7567ef/sklearn/utils/discovery.py#L17-L127).

tests/integration/test_model.py
```diff
@@ -77,66 +79,68 @@ def test_training_and_prediction(c, training_df):
...
 
 
 @pytest.mark.gpu
 def test_cuml_training_and_prediction(c, gpu_training_df):
-    model_query = """
+    c.sql(
+        """
         CREATE OR REPLACE MODEL my_model WITH (
-            model_class = 'cuml.linear_model.LogisticRegression',
+            model_class = 'LogisticRegression',
```
charlesbluca, line 90: Now that this test has been generalized, is it possible to remove the `pytest.mark.gpu` and run on both CPU and GPU? This question also applies to the other tests marked for GPU here

tests/integration/test_model.py
```diff
@@ -77,66 +79,68 @@ def test_training_and_prediction(c, training_df):
...
         )
-        """
-    c.sql(model_query)
+    """
+    )
     check_trained_model(c)
 
 
 @xfail_if_external_scheduler
 @pytest.mark.gpu
```
charlesbluca, line 121: Don't think this should have too much of an impact, but just noting that `xfail_if_external_scheduler` shouldn't have any impact on a `pytest.mark.gpu` test since they're automatically skipped on our CPU testing

dask_sql/physical/rel/custom/ml_classes.py
```diff
@@ -0,0 +1,382 @@
+def get_cpu_classes():
```
charlesbluca, line 1: For organizational purposes, wonder if this file would make more sense under `dask_sql/physical/utils/`?

tests/integration/test_model.py
```diff
@@ -1036,3 +1054,128 @@ def test_predict_with_nullable_types(c):
...
         check_dtype=False,
     )
+
+
+@pytest.mark.skipif(
+    sys.version_info < (3, 9),
+    reason="Some newer sklearn classes are only available with Python version >= 3.9",
+)
+@pytest.mark.parametrize("gpu", [False, pytest.param(True, marks=pytest.mark.gpu)])
+def test_ml_class_mappings(gpu):
```
charlesbluca, line 1064: Since this is testing the mappings directly, maybe this would make sense in its own file in the unit tests? Or we could merge this into `test_ml_wrapppers.py` and just have that be something like `test_ml_utils.py`

tests/integration/test_model.py
sarahyurick: I ended up not doing this because of the `training_df` and `gpu_training_df` fixtures. There's a series of commits on this PR where I tried a couple of different ways but was having trouble getting it to work, and I didn't necessarily want to modify the fixtures themselves because that would require also modifying other unrelated tests.

I guess another solution is to just not use the fixtures and have a `gpu` param instead to create the tables within the test. Does that sound ok?

charlesbluca: Would it work if we merged `training_df` and `gpu_training_df` into a single fixture that registers both tables (skipping the GPU case if dask-cudf isn't available) and then used the `gpu` param to select which table is used to create the model?

dask_sql/physical/rel/custom/ml_classes.py
sarahyurick: I didn't know about this, cool! LightGBM and XGBoost don't seem to have a similar function, but they have a much smaller number of estimators anyway where I wouldn't expect them to be prone to changing.

I couldn't find the function in cuML, but I opened https://github.com/rapidsai/cuml/issues/5162 as a possibility for the future. In the meantime, I kept the `gpu_classes` dict as is.

sarahyurick:
Thanks @charlesbluca ! I've applied all of your suggestions, so let me know what you think.

dask_sql/physical/rel/custom/create_model.py
```diff
@@ -130,6 +135,13 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
         wrap_fit = kwargs.pop("wrap_fit", None)
         fit_kwargs = kwargs.pop("fit_kwargs", {})
 
+        training_df = context.sql(select)
+
+        if type(training_df) == dd.core.DataFrame:
+            model_class = cpu_classes.get(model_class, model_class)
+        elif "cudf" in str(training_df._partition_type):
+            model_class = gpu_classes.get(model_class, model_class)
```
charlesbluca, line 143: Nitpick but we should be able to avoid the Dask import here by inverting the conditional:

```suggestion
        if is_cudf_type(training_df):
            model_class = gpu_classes.get(model_class, model_class)
        else:
            model_class = cpu_classes.get(model_class, model_class)
```

dask_sql/physical/rel/custom/create_model.py
```diff
@@ -1,11 +1,13 @@
 import logging
 from typing import TYPE_CHECKING
 
+import dask.dataframe as dd
```
sarahyurick, line 4:
```suggestion
```

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -145,6 +149,13 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
         y = training_df[target_column]
 
         if model_class and experiment_class:
+            if type(training_df) == dd.core.DataFrame:
+                model_class = cpu_classes.get(model_class, model_class)
+                experiment_class = cpu_classes.get(experiment_class, experiment_class)
+            elif "cudf" in str(training_df._partition_type):
+                model_class = gpu_classes.get(model_class, model_class)
+                experiment_class = gpu_classes.get(experiment_class, experiment_class)
```
sarahyurick, line 157:
```suggestion
            if is_cudf_type(training_df):
                model_class = gpu_classes.get(model_class, model_class)
                experiment_class = gpu_classes.get(experiment_class, experiment_class)
            else:
                model_class = cpu_classes.get(model_class, model_class)
                experiment_class = cpu_classes.get(experiment_class, experiment_class)
```

dask_sql/physical/rel/custom/create_experiment.py
```diff
@@ -6,6 +6,7 @@
 
 from dask_sql.datacontainer import ColumnContainer, DataContainer
 from dask_sql.physical.rel.base import BaseRelPlugin
+from dask_sql.physical.utils.ml_classes import get_cpu_classes, get_gpu_classes
 from dask_sql.utils import convert_sql_kwargs, import_class
```
sarahyurick, line 10:
```suggestion
from dask_sql.utils import convert_sql_kwargs, import_class, is_cudf_type
```

dask_sql/physical/rel/custom/create_model.py
```diff
@@ -1,11 +1,13 @@
...
 import numpy as np
 from dask import delayed
 
 from dask_sql.datacontainer import DataContainer
 from dask_sql.physical.rel.base import BaseRelPlugin
+from dask_sql.physical.utils.ml_classes import get_cpu_classes, get_gpu_classes
 from dask_sql.utils import convert_sql_kwargs, import_class
```
sarahyurick, line 11:
```suggestion
from dask_sql.utils import convert_sql_kwargs, import_class, is_cudf_type
```

tests/integration/test_model.py
```diff
@@ -47,140 +47,176 @@ def check_trained_model(c, model_name=None):
...
-
-
-@pytest.fixture()
-def gpu_training_df(c):
     if dask_cudf:
-        df = timeseries(freq="1d").reset_index(drop=True)
         df = dask_cudf.from_dask_dataframe(df)
         c.create_table("timeseries", input_table=df)
+    else:
+        c.create_table("timeseries", df, persist=True)
```
charlesbluca, line 54: Think we'll want the CPU/GPU tables to have separate names that we parametrize on in tests, otherwise we end up defaulting to GPU testing whenever dask-cudf is importable

charlesbluca (commented):
Looks like the gpuCI failures here are due to `is_cudf_type` not behaving as expected; I think the modifications you made to the utility in #984 should be sufficient to resolve here

tests/integration/test_model.py
```diff
@@ -44,153 +35,165 @@ def check_trained_model(c, model_name=None):
...
         )
     """
     )
+    check_trained_model(c, df_name=timeseries)
 
-    check_trained_model(c)
+    # TODO: In this query, we are using cuml.dask.linear_model.LinearRegression
+    # instead of cuml.linear_model.LinearRegression.
+    # Is there any way to assert that we are using the cuML Dask estimator
+    # (and not just the cuML estimator)?
```
charlesbluca, line 82: Is this a concern for just this specific model class, or throughout the test suite? In either case, as long as `wrap_predict=False` then something like

```python
assert "dask" in str(c.schema["root"].models["my_model"][0].__class__).lower()
```

Should work here

sarahyurick: Cool, thanks! The reason I wanted to have a check is because originally this code was part of `test_dask_cuml_training_and_prediction` which was specifically meant to test `cuml.dask.linear_model.LinearRegression` functionality.

With this PR, `LinearRegression` and [some others](https://github.com/dask-contrib/dask-sql/blob/9425286a326051613403fd03bc513ef302876a11/dask_sql/physical/utils/ml_classes.py#L32) map to the `cuml.dask` version automatically. I figured it was nicer to map to the Dask version of a model if it's available - what do you think?

dask_sql/physical/utils/ml_classes.py
```diff
@@ -0,0 +1,130 @@
...
+    cpu_classes["XGBRegressor"] = "xgboost.XGBRegressor"
+    cpu_classes["XGBClassifier"] = "xgboost.XGBClassifier"
+    cpu_classes["XGBRanker"] = "xgboost.XGBRanker"
+    cpu_classes["XGBRFRegressor"] = "xgboost.XGBRFRegressor"
+    cpu_classes["XGBRFClassifier"] = "xgboost.XGBRFClassifier"
+    cpu_classes["DaskXGBClassifier"] = "xgboost.dask.DaskXGBClassifier"
+    cpu_classes["DaskXGBRegressor"] = "xgboost.dask.DaskXGBRegressor"
+    cpu_classes["DaskXGBRanker"] = "xgboost.dask.DaskXGBRanker"
+    cpu_classes["DaskXGBRFRegressor"] = "xgboost.dask.DaskXGBRFRegressor"
+    cpu_classes["DaskXGBRFClassifier"] = "xgboost.dask.DaskXGBRFClassifier"
```
charlesbluca, line 25: Just noticed that these classes are shared between both the CPU and GPU classes; knowing this, would it make sense to make this a common dict that's added to both mappings with `dict.update` to cut down on repetition?

tests/integration/test_model.py
charlesbluca: > I figured it was nicer to map to the Dask version of a model if it's available

That makes sense to me, this shouldn't be too limiting since users can always just specify the non-Dask version explicitly if they need to use it for some reason.

IMO unless there's some implicit fallback behavior in the `cuml.dask` classes when certain conditions aren't met such that they default to their non-Dask variants, I would find the checks a little superfluous, as the mapping to Dask classes and their import is relatively straightforward and I wouldn't expect it to change unless we specifically modified the mapping; if there's a case where this could change unexpectedly though would be happy to know more about it 🙂 and would recommend adding this check conditionally to `check_trained_model` so that it can easily be extended to other cases where we might want to assert the Dask class is being used.

dask_sql/physical/utils/ml_classes.py
sarahyurick: Done, thanks!

dask_sql/physical/utils/ml_classes.py
```diff
@@ -0,0 +1,122 @@
...
+    gpu_classes = {
+        # cuml.dask
+        "DBSCAN": "cuml.dask.cluster.dbscan.DBSCAN",
+        "KMeans": "cuml.dask.cluster.kmeans.KMeans",
+        "PCA": "cuml.dask.decomposition.pca.PCA",
+        "TruncatedSVD": "cuml.dask.decomposition.tsvd.TruncatedSVD",
+        "RandomForestClassifier": "cuml.dask.ensemble.randomforestclassifier.RandomForestClassifier",
+        "RandomForestRegressor": "cuml.dask.ensemble.randomforestregressor.RandomForestRegressor",
+        # ImportError: dask-glm >= 0.2.1.dev was not found, please install it to use multi-GPU logistic regression.
+        # "LogisticRegression": "cuml.dask.extended.linear_model.logistic_regression.LogisticRegression",
```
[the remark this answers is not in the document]
sarahyurick: Sure, I opened https://github.com/dask-contrib/dask-sql/issues/1015 and linked to it.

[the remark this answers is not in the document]
sarahyurick: Not sure if this is something we should fix on the Dask-SQL side (see https://github.com/dask-contrib/dask-sql/issues/1015#issuecomment-1409076884). Would it make sense to have some sort of try/except logic around this?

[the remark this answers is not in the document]
charlesbluca: Thanks @sarahyurick 🙂 with that additional context, IMO I feel like the best option on our end for now would be to include `cuml.dask.extended.linear_model.logistic_regression.LogisticRegression` in the mapping and raise the `ImportError` directly,  with the idea being that users would just subsequently install `dask-glm` into their environment and try rerunning their query.

I suppose if we wanted to be as informative as possible, we could reraise the error with some addendum about using the non-Dask equivalent class if installing `dask-glm` isn't an option, though I think that message would probably make more sense to add upstream since it would be informative to cuML users at large.

tests/integration/test_model.py
```diff
@@ -44,153 +36,167 @@ def check_trained_model(c, model_name=None):
...
+            LIMIT 100
         )
         """
-    c.sql(model_query)
-    check_trained_model(c)
+        )
+        check_trained_model(c)
 
+    # For GPU tests, set tree_method = 'gpu_hist'
+    tree_method = "gpu_hist" if gpu else "hist"
```
ayushdg, line 138: As seen here if users are working with gpu df's and haven't specified tree_method `gpu_hist` we end up training the model on the cpu's. 
Do you have any opinions on whether we should warn the users in this case or if it's up to the user to specify these arguments as expected.

tests/integration/test_model.py
```diff
@@ -663,20 +678,22 @@ def test_mlflow_export_xgboost(c, client, training_df, tmpdir):
...
+def test_mlflow_export_lightgbm(c, tmpdir):
     # Test only when mlflow & lightgbm was installed
     mlflow = pytest.importorskip("mlflow", reason="mlflow not installed")
     lightgbm = pytest.importorskip("lightgbm", reason="lightgbm not installed")
+
     c.sql(
         """
         CREATE MODEL IF NOT EXISTS my_model_lightgbm WITH (
-            model_class = 'lightgbm.LGBMClassifier',
+            model_class = 'LGBMClassifier',
```
ayushdg, line 696: Do we know what happens if our input df's are gpu tables and we try using LightGBM. I don't recall if we error or not when passing a dask-cudf input to lgbm

ayushdg (commented):
Changes generally lgtm. Just a few clarifying questions

tests/integration/test_model.py
sarahyurick: I remember trying it with `LGBMClassifier` and it failed. I'm not sure about all of the others like `LGBMRegressor`, etc. At least with XGBoost, the Regressors worked but the Classifiers failed, so it might be similar behavior.

I have https://github.com/dask-contrib/dask-sql/issues/943 open to investigate some cases where the GPU version fails with `experiment_class`, as many of them fail with the same `TypeError: Implicit conversion to a NumPy array is not allowed`. I could also open an issue to further investigate GPU LightGBM and XGBoost with `model_class`.

tests/integration/test_model.py
sarahyurick: Personally, I think that type of warning would make more sense coming from the XGBoost side. I could open an issue with them of that's of interest?

If you think we should just raise the warning on the Dask-SQL side instead, though, I'd be happy to add it.

tests/integration/test_model.py
sarahyurick: Opened https://github.com/dask-contrib/dask-sql/issues/1020

tests/integration/test_model.py
ayushdg: No I think it's a good idea to suggest xgboost to warn users when such a situation arises. Okay to not have anything on the dask-sql side

ayushdg (approved):
Thanks a lot @sarahyurick. As a followup to this we should update some of the docs to incorporate some of the features added in this pr. https://dask-sql.readthedocs.io/en/latest/machine_learning.html

charlesbluca (approved):
Thanks again @sarahyurick 😄

PR #963, opened 2022-12-08
Add basic ROLLUP support for groupbys

charlesbluca:
Adds in a crude implementation of SQL groupbys with rollup; this is achieved by iterated through the various combinations of column that need to be grouped with `_do_aggregations`, and then concating all the resulting frames together.


charlesbluca:
Looks like the failures here are because in the bump from datafusion 14 -> 22, we are no longer checking for `rollup`/`cube` as reserved function names?

https://github.com/apache/arrow-datafusion/blob/6d00bd990ce5644181ad1549a6c70c8406219070/datafusion/sql/src/planner.rs#L2148-L2155

Though it seems like there are still working tests using rollup so this might just be an indication that we need to make some modifications to our parsing logic around aggregate functions

PR #977, opened 2022-12-21
Remove `np.float` in PyTest and skip new NumPy failures

sarahyurick:
[NumPy 1.24.0](https://numpy.org/doc/stable/release/1.24.0-notes.html#expired-deprecations) removed the `np.float` alias, so we have to replace `np.float` with `np.float64` in PyTest `test_custom_function_row_two_args`.

Another option is to replace `np.float` with Python's `float`, although the NumPy version may be useful for consistency with NumPy arrays ([source](https://numpy.org/doc/stable/release/1.20.0-notes.html#deprecations)).

### DIFF
--- a/tests/integration/test_model.py
+++ b/tests/integration/test_model.py
@@ -899,6 +899,7 @@ def test_ml_experiment(c, client, training_df):
 
 # TODO - many ML tests fail on clusters without sklearn - can we avoid this?
 @skip_if_external_scheduler
+@pytest.mark.skip(reason="Waiting on https://github.com/EpistasisLab/tpot/pull/1280")
 def test_experiment_automl_classifier(c, client, training_df):
     tpot = pytest.importorskip("tpot", reason="tpot not installed")
     # currently tested with tpot==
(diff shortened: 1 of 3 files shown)


tests/integration/test_model.py
```diff
@@ -924,6 +925,7 @@ def test_experiment_automl_classifier(c, client, training_df):
 
 # TODO - many ML tests fail on clusters without sklearn - can we avoid this?
 @skip_if_external_scheduler
+@pytest.mark.skip(reason="Waiting on https://github.com/EpistasisLab/tpot/pull/1280")
```
sarahyurick, line 928: TPOT fails with the newest NumPy version as well. I've opened https://github.com/EpistasisLab/tpot/pull/1280 to resolve this.

tests/integration/test_rex.py
```diff
@@ -140,6 +141,8 @@ def test_literal_null(c):
     assert_eq(df, expected_df)
 
 
+# TODO - https://github.com/dask-contrib/dask-sql/issues/978
+@skip_if_external_scheduler
 def test_random(c):
```
sarahyurick, line 146: Opened https://github.com/dask-contrib/dask-sql/issues/978

ayushdg (approved):

PR #984, opened 2023-01-05
Support `DATE` Extraction

### LINKED ISSUE #901: [ENH] `EXTRACT(DATE from d)`
Now that https://github.com/sqlparser-rs/sqlparser-rs/pull/691 has been merged, we should be able to support `SELECT EXTRACT(DATE FROM d)` once DataFusion updates their SQLParser version.

sarahyurick:
Closes https://github.com/dask-contrib/dask-sql/issues/901


dask_sql/physical/rex/core/call.py
```diff
@@ -629,10 +629,21 @@ def extract(self, what, df: SeriesOrScalar):
...
             raise NotImplementedError(f"Extraction of {what} is not (yet) implemented.")
 
 
+class ExtractDateOperation(Operation):
+    def __init__(self):
+        super().__init__(self.extract_date)
+
+    def extract_date(self, df: SeriesOrScalar):
+        df = convert_to_datetime(df)
+        return df.date
```
ayushdg, line 644: since `date` is a special case in that returns a numpy array of python datetime.date objects instead of a standard numpy type do we know what would happen if we try using the results from an extract elsewhere within a query?

ayushdg (changes requested):

dask_planner/src/dialect.rs
```diff
@@ -194,6 +194,31 @@ impl Dialect for DaskDialect {
...
+                    // convert to function args
+                    let args = vec![FunctionArg::Unnamed(FunctionArgExpr::Expr(expr))];
+
+                    Ok(Some(Expr::Function(Function {
+                        name: ObjectName(vec![Ident::new("extract_date")]),
+                        args,
+                        over: None,
+                        distinct: false,
+                        special: false,
+                    })))
```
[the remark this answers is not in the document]
jdye64: Same question for me

jdye64:
Syntactically PR looks good. I'm going to leave the actual logical review to @ayushdg however as he knows far more about python and date types/behavior than I do.

dask_planner/src/dialect.rs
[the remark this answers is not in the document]
sarahyurick: Yes, the new code in `dialect.rs` and `sql.rs` are needed because normally `extract` returns an integer, but we need to be able to return a `datetime` (without this extra code, it fails because it tries to cast the date as an integer).

dask_sql/physical/rex/core/call.py
```diff
@@ -629,10 +629,21 @@ def extract(self, what, df: SeriesOrScalar):
...
+
+
```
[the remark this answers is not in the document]
sarahyurick: Yes, thanks! I'll update that.

dask_sql/physical/rex/core/call.py
sarahyurick: It looks like it depends on the query. For example, the query
```
SELECT EXTRACT(YEAR FROM EXTRACT(DATE FROM d)) as "date" FROM df
```
fails, but the query
```
SELECT TIMESTAMPADD(YEAR, 1, EXTRACT(DATE FROM d)) FROM df
```
works fine and produces the expected output.

sarahyurick: Let me know if there's any other specific cases you'd like me to try out.

sarahyurick:
Just pushed a commit that uses `df.strftime("%Y-%m-%d")` instead of `df.date`, since the latter seems to be shaky when combined with other use cases. Here are some examples that work with these changes (`timestampadd`, date/string comparison, and the `between` operator): https://gist.github.com/sarahyurick/d5741d8a8e7f05a5ea8fcb54039bb6c5

There are still some examples that don't work though, like `SELECT EXTRACT(DATE FROM d) + INTERVAL '2 days' FROM df`.

dask_sql/physical/rex/core/call.py
```diff
@@ -960,6 +962,11 @@ def date_part(self, what, df: SeriesOrScalar):
             return df.week
         elif what in {"YEAR", "YEARS"}:
             return df.year
+        elif what == "DATE":
+            if "cudf" in str(df._meta):
+                return df.strftime("%Y-%m-%d")
+            else:
+                return dd.to_datetime(df.strftime("%Y-%m-%d"))
```
sarahyurick, line 969: This chunk of code allows something like `SELECT EXTRACT(DATE FROM d) + INTERVAL '2 days' FROM df` to work on the CPU, since the DataFrame needs to be converted to a datetime dtype in order to use the `+` operator. However, if it's a GPU DataFrame, then `dd.to_datetime()` causes errors. I couldn't get `cudf.to_datetime()` to work in this case, and I don't think there's a `dask_cudf.to_datetime()` function at all.

TLDR: This code allows flexibility with different operations on the CPU, but doesn't work for every GPU example. See: https://gist.github.com/sarahyurick/d5741d8a8e7f05a5ea8fcb54039bb6c5 where `df` is a CPU table and `table1` is a GPU table.

dask_sql/physical/rex/core/call.py
```diff
@@ -629,6 +629,8 @@ def extract(self, what, df: SeriesOrScalar):
             return df.week
         elif what in {"YEAR", "YEARS"}:
             return df.year
+        elif what == "DATE":
+            return df.date()
```
charlesbluca, line 633: Random question - is `ExtractOperation` still accessible through queries? Running through the test suite, it seems like the only time that this function is entered is through `test_dates` when it's called explicitly, but it seems like in all other cases we would use `DatePartOperation`

dask_sql/physical/rex/core/call.py
```diff
@@ -960,6 +962,11 @@ def date_part(self, what, df: SeriesOrScalar):
             return df.week
         elif what in {"YEAR", "YEARS"}:
             return df.year
+        elif what == "DATE":
+            if "cudf" in str(df._meta):
```
charlesbluca, line 966: Once `main` is merged in we should be able to use `is_cudf_type` for this 🙂

charlesbluca (commented):
Thanks @sarahyurick! Still want to take some time to look into the GPU case you outlined [here](https://github.com/dask-contrib/dask-sql/pull/984/files#r1067581424) but these are some initial thoughts

dask_sql/physical/rex/core/call.py
sarahyurick: Another note: I already have an issue https://github.com/rapidsai/cudf/issues/12062 about `dd.to_datetime` which would probably solve this error.

dask_sql/physical/rex/core/call.py
sarahyurick: It doesn't look like it - none of the combinations of `EXTRACT(DATE FROM d)`, `EXTRACT(YEAR FROM d)`, etc. queries I tested out used it.

When I remove the `ExtractOperation` class entirely, it doesn't appear to cause any of the tests in `test_rex` to fail.

Should we still keep the class? I could also open another PR for removing it.

dask_planner/src/dialect.rs
```diff
@@ -194,6 +194,36 @@ impl Dialect for DaskDialect {
...
+                    // EXTRACT(DATE FROM d)
+                    parser.next_token(); // skip extract
+                    parser.expect_token(&Token::LParen)?;
+                    if !parser.parse_keyword(Keyword::DATE) {
+                        // Parse EXTRACT(x FROM d) as normal
+                        parser.prev_token();
+                        parser.prev_token();
+                        return Ok(None);
+                    }
+                    let _from = parser.parse_keyword(Keyword::FROM);
```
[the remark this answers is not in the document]
sarahyurick: That works, thanks! Updating this in a different commit since there's an extra parenthesis in the suggestion.

dask_sql/physical/rex/core/call.py
charlesbluca: Yeah removing it in a follow up makes sense to me, thanks @sarahyurick!

sarahyurick: Sounds good, opened https://github.com/dask-contrib/dask-sql/pull/1011

sarahyurick:
Ready for re-review

Should be okay as the CI errors (https://github.com/dask-contrib/dask-sql/actions/runs/5214904844/jobs/9411776312?pr=984) seem to be related to conda bugginess. We can try to rerun the tests later.

charlesbluca:
Now that #1170 is in and we're using 23.08 nightlies looks like we're getting some new Python test failures

charlesbluca (approved):

dask_sql/physical/rex/core/call.py
```diff
@@ -929,6 +929,11 @@ def date_part(self, what, df: SeriesOrScalar):
             return df.week
         elif what in {"YEAR", "YEARS"}:
             return df.year
+        elif what == "DATE":
+            if isinstance(df, pd.Timestamp):
+                return df.date()
+            else:
+                return dd.to_datetime(df.strftime("%Y-%m-%d"))
```
ayushdg, line 936: It might also make sense to check for dask-cudf todatetime support and raise here if we have an older version, mentioning the lack of support for versions older than 2023.5.1 if df is a dask-cudf dataframe

ayushdg (changes requested):
Really small suggestion. Everything else looks good to me. Thanks @sarahyurick !

dask_sql/physical/rex/core/call.py
sarahyurick: Thanks! Updated.

tests/integration/test_rex.py
```diff
@@ -1054,3 +1057,66 @@ def test_totimestamp(c, gpu):
...
+    "gpu",
+    [
+        False,
+        pytest.param(
+            True,
+            marks=(
+                pytest.mark.gpu,
+                pytest.mark.skipif(
+                    not DASK_CUDF_TODATETIME_SUPPORT,
+                    reason="Requires https://github.com/dask/dask/pull/9881",
```
ayushdg, line 1072: In line with the previous change, we can also expect a runtime error here instead of skipping the test for this case. 
I think `pytest.mark.xfail` with the `raises` param should work https://docs.pytest.org/en/7.1.x/reference/reference.html#pytest-mark-xfail

ayushdg (approved):
lgtm! Thanks

PR #990, opened 2023-01-09
Resolve test failures from `fugue=0.8.0`

charlesbluca:
It looks like the API of `FugueSQLWorkflow` has changed such that we no longer pass through an execution engine at creation; this PR changes things accordingly to unblock CI.

### DIFF
--- a/continuous_integration/environment-3.10-dev.yaml
+++ b/continuous_integration/environment-3.10-dev.yaml
@@ -5,7 +5,7 @@ channels:
 dependencies:
 - dask>=2022.3.0
 - fastapi>=0.69.0
-- fugue>=0.7.0
+- fugue>=0.7.3
 - intake>=0.6.0
 - jsonschema
 - lightgbm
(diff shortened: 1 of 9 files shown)


ayushdg (approved):
Changes generally lgtm! 

Do you think it makes sense for bumping our min version req to 0.8.0 or leave it as is for now?

charlesbluca:
We should be good leaving the min version as is, since IIUC this style of using `FugueSQLWorkflow` was also supported for fugue 0.7.0.

That being said, I'm okay with bumping to the most recent patch release of fugue 0.7 (0.7.3) to see if that resolves the occasional [failures](https://github.com/dask-contrib/dask-sql/actions/runs/3876796586/jobs/6611064639) we see on macOS.

ayushdg:
> We should be good leaving the min version as is, since IIUC this style of using `FugueSQLWorkflow` was also supported for fugue 0.7.0.
> 
> That being said, I'm okay with bumping to the most recent patch release of fugue 0.7 (0.7.3) to see if that resolves the occasional [failures](https://github.com/dask-contrib/dask-sql/actions/runs/3876796586/jobs/6611064639) we see on macOS.

Yeah let's try 0.7.3 to see if that reduces the occasional failures on the 3.8 tests on Mac and sometimes windows

PR #998, opened 2023-01-18
Upgrade to DataFusion 17.0.0

andygrove:
Changes in DataFusion that are causing tests to fail:

- Window frame offsets can have different data types depending on the context but Dask SQL currently expects UInt64
- Join conditions are no longer limited to `Column` but can be any `Expr`


andygrove:
current failures:

```
FAILED tests/integration/test_compatibility.py::test_single_agg_count_no_group_by - ValueError: conflicting aggregation functions: [('count', 'a'), ('count', 'a')]
FAILED tests/integration/test_compatibility.py::test_multi_agg_count_no_group_by - ValueError: conflicting aggregation functions: [('count', 'a'), ('count', 'a')]
FAILED tests/integration/test_compatibility.py::test_agg_count_distinct_group_by - AssertionError: DataFrame.iloc[:, 1] (column name="cd_b") are different
FAILED tests/integration/test_compatibility.py::test_agg_count_distinct_no_group_by - AssertionError: DataFrame.iloc[:, 0] (column name="cd_a") are different
FAILED tests/integration/test_compatibility.py::test_window_row_number - RuntimeError: Internal("Unexpected window frame bound: Int64(NULL)")
FAILED tests/integration/test_compatibility.py::test_window_row_number_partition_by - RuntimeError: Internal("Unexpected window frame bound: Int64(NULL)")
FAILED tests/integration/test_compatibility.py::test_window_sum_avg - AssertionError
FAILED tests/integration/test_compatibility.py::test_window_sum_avg_partition_by - AssertionError
FAILED tests/integration/test_compatibility.py::test_window_min_max - AssertionError
FAILED tests/integration/test_compatibility.py::test_window_min_max_partition_by - AssertionError
FAILED tests/integration/test_compatibility.py::test_window_count - AssertionError
FAILED tests/integration/test_compatibility.py::test_window_count_partition_by - AssertionError
FAILED tests/integration/test_compatibility.py::test_nested_query - RuntimeError: Internal("Unexpected window frame bound: Float64(NULL)")
FAILED tests/integration/test_filter.py::test_predicate_pushdown[SELECT * FROM parquet_ddf WHERE (b > 5 AND b < 10) OR a = 1-<lambda>-filters2] - AssertionError: assert frozenset({fr..., '==', 1)})}) == frozenset({fr..., '<', 10)})})
FAILED tests/integration/test_filter.py::test_predicate_pushdown[SELECT a FROM parquet_ddf WHERE (b > 5 AND b < 10) OR a = 1-<lambda>-filters5] - AssertionError: assert frozenset({fr..., '==', 1)})}) == frozenset({fr..., '<', 10)})})
FAILED tests/integration/test_join.py::test_join_literal - pyo3_runtime.PanicException: assertion failed: !filters.is_empty()
FAILED tests/integration/test_join.py::test_join_on_unary_cond_only - pyo3_runtime.PanicException: assertion failed: !filters.is_empty()
FAILED tests/integration/test_join.py::test_intersect - TypeError: "unsupported join condition"
FAILED tests/integration/test_over.py::test_over_with_sorting - RuntimeError: Internal("Unexpected window frame bound: Int64(NULL)")
FAILED tests/integration/test_over.py::test_over_with_partitioning - AssertionError
FAILED tests/integration/test_over.py::test_over_with_grouping_and_sort - RuntimeError: Internal("Unexpected window frame bound: Int64(NULL)")
FAILED tests/integration/test_over.py::test_over_with_different - RuntimeError: Internal("Unexpected window frame bound: Int64(NULL)")
FAILED tests/integration/test_over.py::test_over_calls - RuntimeError: Internal("Unexpected window frame bound: Int64(NULL)")
FAILED tests/integration/test_over.py::test_over_with_windows - AssertionError
FAILED tests/integration/test_rex.py::test_literals - AssertionError: numpy array are different
```

dask_planner/src/sql/logical/window.rs
```diff
@@ -156,26 +165,26 @@ impl PyWindowFrameBound {
...
+                x if x.is_null() => Ok(None),
+                ScalarValue::UInt64(v) => Ok(*v),
+                ScalarValue::Int64(v) => Ok(v.map(|n| n as u64)),
+                ref x => Err(DaskPlannerError::Internal(format!(
+                    "Unexpected window frame bound: {:?}",
+                    x
+                ))
+                .into()),
+            },
+            // The below is only safe because window bounds cannot be negative
```
ayushdg, line 179: There's probably a better way of writing the same logic

dask_sql/physical/rex/core/alias.py
```diff
@@ -0,0 +1,41 @@
...
+        assert len(operands) == 1
+
+        sub_rex = operands[0]
+
+        value = RexConverter.convert(rel, sub_rex, dc, context=context)
+
+        if isinstance(value, DataContainer):
+            return value.df
+
+        return value
```
charlesbluca, line 41: Added this to resolve the query failures coming from our lack of handling for rex aliases; for now I've kept this separate from the relatively similar `RexScalarSubqueryPlugin`, not sure if there's a case to have these merged into one plugin that then conditionally grabs and converts the sub-rel/rex

.github/workflows/conda.yml
```diff
@@ -27,6 +27,9 @@ defaults:
   run:
     shell: bash -l {0}
 
+env:
```
jdye64, line 30: I vote we just leave this. Doesn't hurt to always have this kind of information in the CI builds

continuous_integration/recipe/meta.yaml
```diff
@@ -18,16 +18,18 @@ build:
     - dask-sql-server = dask_sql.server.app:main
     - dask-sql = dask_sql.cmd:main
   string: py{{ python | replace(".", "") }}_{{ GIT_DESCRIBE_HASH }}_{{ GIT_DESCRIBE_NUMBER }}
-  script: {{ PYTHON }} -m pip install . --no-deps -vv
+  script: RUST_BACKTRACE=full {{ PYTHON }} -m pip install . --no-deps -vv
```
jdye64, line 21: Mine as well just leave this for more details in CI builds. No need to remove it.

dask_planner/src/sql/optimizer.rs
```diff
@@ -63,29 +57,27 @@ impl DaskSqlOptimizer {
             Arc::new(FilterNullJoinKeys::default()),
             Arc::new(EliminateOuterJoin::new()),
             Arc::new(PushDownFilter::new()),
-            Arc::new(LimitPushDown::new()),
+            Arc::new(PushDownLimit::new()),
             // Dask-SQL specific optimizations
-            Arc::new(FilterColumnsPostJoin::new()),
+            // Arc::new(FilterColumnsPostJoin::new()),
```
jdye64, line 62: This is temporary. There are some changes coming with optimizers but this one is currently causing issues with certain queries.

jdye64 (approved):
LGTM. This was a long haul of a PR

.github/workflows/conda.yml
charlesbluca: I feel like we could remove this backtrace flag as IIUC it doesn't actually get propagated down to the conda build? Okay with leaving it in the recipe itself though

.github/workflows/conda.yml
```diff
@@ -35,6 +38,12 @@ jobs:
       - uses: actions/checkout@v3
         with:
           fetch-depth: 0
+      - name: Setup Rust Toolchain
+        uses: actions-rs/toolchain@v1
+        id: rust-toolchain
+        with:
+          toolchain: stable
+          override: true
```
charlesbluca, line 46: Do we still need this? From what I gather looks like builds started passing once we removed the Substrait components, is that correct? I would think this wouldn't have an impact on whether or not the conda build succeeds based on our previous attempts

continuous_integration/recipe/meta.yaml
```diff
@@ -18,16 +18,18 @@ build:
...
 
 requirements:
   build:
     - {{ compiler('rust') }}
-    - setuptools-rust >=1.4.1
+    - setuptools-rust >=1.5.2
+    - protobuf
```
charlesbluca, line 27: Think we should also add `protobuf` to our CI environments to explicitly make sure it's getting pulled, and maybe establish a min version like we do for `setuptools-rust`; can look into this a bit more

tests/integration/test_filter.py
```diff
@@ -157,12 +157,7 @@ def test_filter_year(c):
         (
             "SELECT * FROM parquet_ddf WHERE (b > 5 AND b < 10) OR a = 1",
             lambda x: x[((x["b"] > 5) & (x["b"] < 10)) | (x["a"] == 1)],
-            [
-                [("a", "==", 1), ("b", "<", 10)],
-                [("a", "==", 1), ("b", ">", 5)],
-                [("b", ">", 5), ("b", "<", 10)],
-                [("a", "==", 1)],
-            ],
+            [[("b", ">", 5), ("b", "<", 10)], [("a", "==", 1)]],
```
charlesbluca, line 160: Just checking; are these param changes related to disabling `FilterColumnsPostJoin`?

charlesbluca (commented):
Thanks @andygrove and @jdye64 for taking the time to unblock CI 🎉 couple of comments / questions but things mostly look good:

continuous_integration/recipe/meta.yaml
charlesbluca: So it looks like the package we want to pull in the `protoc` bin is `libprotobuf`, and based on the environment solve we probably want to pull in `libprotobuf=3` to make sure conflicts aren't raised with the `protobuf` package, can try this out locally and push up

charlesbluca: Actually looks like `protoc` was only needed for the Substrait components, so should be good to remove for now until we try to add in those components later on

tests/integration/test_filter.py
jdye64: I don't know why those changed? I don't think it was me that changed them unless I just don't remember. It should not be related to `FilterColumnsPostJoin` however.

jdye64: I will say the new change seems more correct than the original way anyway.

ayushdg: I can comment a bit more on this. This isn't related to `FilterColumnsPostJoin`, but how datafusion optimizes predicates for table_scan operations. Starting df14, they started using cnf (conjunctive normal form) for the expressions being passed as filters. (See https://github.com/dask-contrib/dask-sql/pull/903#issuecomment-1302767610). 
Since this particular test already had filters in dnf, the conversion of dnf->cnf (by df) -> dnf (by dask-sql since that's what arrow expects) resulted in an overly complex filter.

df16/17 seems to simplify some of these expressions and doesn't convert everything to cnf.

dask_planner/Cargo.toml
```diff
@@ -9,12 +9,12 @@ edition = "2021"
 rust-version = "1.62"
 
 [dependencies]
-arrow = { version = "28.0.0", features = ["prettyprint"] }
-async-trait = "0.1.64"
-datafusion-common = "15.0.0"
-datafusion-expr = "15.0.0"
-datafusion-optimizer = "15.0.0"
-datafusion-sql = "15.0.0"
+async-trait = "0.1.63"
```
ayushdg, line 12: Do we specifically need an older version of `async-trait` or is this just a result of a bad merge with main?

dask_planner/src/sql/optimizer.rs
```diff
@@ -7,40 +7,35 @@ use datafusion_optimizer::{
...
 pub struct DaskSqlOptimizer {
-    skip_failing_rules: bool,
     optimizer: Optimizer,
 }
 
 impl DaskSqlOptimizer {
     /// Creates a new instance of the DaskSqlOptimizer with all the DataFusion desired
     /// optimizers as well as any custom `OptimizerRule` trait impls that might be desired.
-    pub fn new(skip_failing_rules: bool) -> Self {
+    pub fn new() -> Self {
```
ayushdg, line 38: Curious why we removed this

tests/integration/test_join.py
```diff
@@ -381,6 +381,9 @@ def test_join_alias_w_projection(c, parquet_ddf):
     assert_eq(result_df, expected_df, check_index=False)
 
 
+@pytest.mark.skip(
```
ayushdg, line 384: Should we mark this as `xfail` instead to be in line with some of the efforts made in #867

ayushdg (changes requested):
Generally lgtm! Just minor comments/questions that should be a quick fix

dask_planner/src/sql/optimizer.rs
jdye64: That configuration value is no longer exposed via datafusion therefore this was "dead_code" and the compiler issuing warnings and `cargo clippy` was failing

jdye64:
@ayushdg how do the changes suit you?

dask_planner/src/sql/optimizer.rs
ayushdg: Ah they might have renamed that one to `skip_failed_rules`. https://docs.rs/datafusion-common/17.0.0/datafusion_common/config/struct.OptimizerOptions.html

We can update in a followup if needed.

ayushdg (approved):

charlesbluca (approved):

PR #1002, opened 2023-01-20
Switch from skipping to xfailing some Python tests

charlesbluca:
Noticed some remaining `pytest.mark.skips` that weren't caught in #867; interested to see if any of these are passing now


tests/integration/test_model.py
```diff
@@ -903,7 +903,6 @@ def test_ml_experiment(c, client, training_df):
 
 # TODO - many ML tests fail on clusters without sklearn - can we avoid this?
 @xfail_if_external_scheduler
-@pytest.mark.skip(reason="Waiting on https://github.com/EpistasisLab/tpot/pull/1280")
```
charlesbluca, line 906: cc @sarahyurick since this is your PR; are you aware of any other upstream changes that could've resolved these failures?

sarahyurick: AFAIK, the only other solution is to restrict the NumPy version to be < 1.24.0

ayushdg (approved):
lgtm! And should be good to go assuming conflicts are fixed and the tests are passing

PR #993, opened 2023-01-17
Use Mambaforge for bare requirements testing

charlesbluca:
It looks like our bare requirements import testing is failing on main:

https://github.com/dask-contrib/dask-sql/actions/runs/3915253264/jobs/6693236881

Looking through the failures, I was surprised to see the `defaults` channel included in the solve, since we specify `nodefaults` in the list of channels; this PR switches this check to using Mambaforge, which should successfully exclude `defaults` as intended - interested in seeing how this impacts the failures.

cc @ayushdg

PR #1005, opened 2023-01-23
Use Mambaforge for scheduled import testing

charlesbluca:
Forgot to make this change in #993; this should unblock the failures in our scheduled upstream testing.


ayushdg:
Triggered a manual upstream run here: https://github.com/dask-contrib/dask-sql/actions/runs/3988766002

PR #1006, opened 2023-01-23
Remove `EliminateAggDistinct` optimizer rule, implement distinct aggs in Python

charlesbluca:
This PR attempts to remove the `EliminateAggDistinct` optimizer rule and instead handle distinct aggregations directly from the Python aggregate plugin using `drop_duplicates`.

This ideally should unblock the groupby related failures cropping up in #998 and resolve the existing groupby agg failures that we had seen in test_compatibility.py.


dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -513,11 +519,21 @@ def _perform_aggregation(
 
             aggregations_dict[input_col][output_col] = aggregation_f
 
+        group_columns = [
```
ayushdg, line 522: Can't comment below but this is duplicated below and can probably be removed from line 542

dask_sql/physical/rel/logical/aggregate.py
```diff
@@ -513,11 +519,21 @@ def _perform_aggregation(
...
+            for group_name in group_columns
+        ]
+
         # filter dataframe if specified
         if filter_column:
             filter_expression = tmp_df[filter_column]
             tmp_df = tmp_df[filter_expression]
             logger.debug(f"Filtered by {filter_column} before aggregation.")
+        if distinct_column:
+            tmp_df = tmp_df.drop_duplicates(subset=(group_columns + [distinct_column]))
```
ayushdg, line 533: Should also propagate the `sql.aggregate` config options here.

tests/unit/test_queries.py
```diff
@@ -42,7 +42,6 @@
     89,
     92,
     94,
-    95,
```
ayushdg, line 45: Nice!

ayushdg (changes requested):

ayushdg (commented):
Changes lgtm!

ayushdg (approved):

PR #1011, opened 2023-01-26
Rename `DatePartOperation` class

sarahyurick:
From discussion here: https://github.com/dask-contrib/dask-sql/pull/984#discussion_r1087071765


tests/unit/test_call.py
```diff
@@ -188,24 +188,7 @@ def test_string_operations():
 
 
 def test_dates():
-    op = call.ExtractOperation()
```
sarahyurick, line 191: Actually, maybe we should keep this test but use `DatePartOperation` instead. The name isn't as intuitive though, would it make sense to rename it as something else?

charlesbluca: Would it make sense to use `ExtractOperation` instead here? Feel like that name made sense for `EXTRACT` operations to begin with, I suppose we just introduced `DatePartOperation` as part of the DataFusion migration?

In favor of retaining these tests for whatever we end up naming the `DatePartOperation`

ayushdg (approved):

charlesbluca (approved):

Issue #982, opened 2023-01-04
Timestamp improvements


sarahyurick:
Following discussions in #973, we should look into the following:

- [x] Create common methods to be used by `timestampadd` and `timestampdiff`
- [x] Check expected logic for scalar inputs
- [x] Add more scalar timestamp tests
- [x] Add GPU timestamp tests

For the second point, look more closely into the expected behaviors of:
```
int_input1 = 1203073300
c.sql(f"SELECT TIMESTAMPADD(YEAR, 2, to_timestamp({int_input1})) as date").compute()
```
```
int_input2 = 1203073300000000000
c.sql(f"SELECT TIMESTAMPADD(YEAR, 2, {int_input2}) as date").compute()
```
```
int_input3 = 1203073300
c.sql(f"SELECT TIMESTAMPADD(YEAR, 2, CAST('{int_input3}' AS TIMESTAMP)) as date").compute()
```

Since `int_input3` is interpreted as `2026-09-05 04:11:38.109558784` (before timestamp operation), and `int_input2` and `int_input1` are both interpreted as `2008-02-15 11:01:40` (even though `int_input2 == int_input1 * 10**9`). The `timestampadd` operation can handle each input, but the value of the input is interpreted differently depending on how we read it (e.g., as a scalar, with `to_timestamp`, or explicitly casting as a timestamp).

sarahyurick:
From Slack discussion:

1. https://gist.github.com/sarahyurick/8eb89826b2c7a8be27d26d7dc425c1f5
2. [as_timelike](https://github.com/dask-contrib/dask-sql/blob/2a8e9582de7b9a2519244796574054f1866222c7/dask_sql/physical/rex/core/call.py#L44) function
3. https://github.com/rapidsai/cudf/issues/12062

PR #1018, opened 2023-01-30
[DO NOT MERGE] Clean up timestamp functions

sarahyurick:
Investigating https://github.com/dask-contrib/dask-sql/issues/982


sarahyurick:
Can we merge `convert_to_datetime` and `as_timelike` into a single utility function?

sarahyurick:
Opened https://github.com/dask/dask/issues/9899

PR #1021, opened 2023-01-31
Skip `test_dask_fsql` on Python 3.8 pytests

sarahyurick:
For the past couple of months, `test_dask_fsql` (formerly `test_fsql`) would fail randomly on our Python 3.8 tests. I was wondering if we should just mark these cases as skips, or if there are any plans to debug this?

From what I've seen, these failures happen on Windows and MacOS, and I don't recall seeing it happen on Ubuntu. If desired, I could add an even more specific condition to skip Windows+Python3.8 and MacOS+Python3.8 and still run Ubuntu+Python3.8.

### DIFF
--- a/tests/integration/test_fugue.py
+++ b/tests/integration/test_fugue.py
@@ -1,3 +1,5 @@
+import sys
+
 import dask.dataframe as dd
 import pandas as pd
 import pytest
@@ -44,6 +46,10 @@ def test_fugue_fsql(client):
 # TODO: Revisit fixing this on an independant cluster (without dask-sql) based on the
 # discussion in https://github.com/dask-contrib/dask-sql/issues/407
 @xfail_if_external_scheduler
+@pytest.mark.skipif(
+    sys.version_info < (3, 9),
+    reason="Intermittent AssertionError on Python 3.8 tests"
+)
 def test_dask_fsql(client):
     def assert_fsql(df: pd.DataFrame) -> None:
         assert_eq(df, pd.DataFrame({"a": [1]}))
--- a/tests/integration/test_fugue.py
+++ b/tests/integration/test_fugue.py
@@ -47,8 +47,7 @@ def test_fugue_fsql(client):
 # discussion in https://github.com/dask-contrib/dask-sql/issues/407
 @xfail_if_external_scheduler
 @pytest.mark.skipif(
-    sys.version_info < (3, 9),
-    reason="Intermittent AssertionError on Python 3.8 tests"
+    sys.version_info < (3, 9), reason="Intermittent AssertionError on Python 3.8 tests"
 )
 def test_dask_fsql(client):
     def assert_fsql(df: pd.DataFrame) -> None:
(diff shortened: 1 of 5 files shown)


tests/integration/test_fugue.py
```diff
@@ -44,6 +46,9 @@ def test_fugue_fsql(client):
 # TODO: Revisit fixing this on an independant cluster (without dask-sql) based on the
 # discussion in https://github.com/dask-contrib/dask-sql/issues/407
 @xfail_if_external_scheduler
+@pytest.mark.skipif(
+    sys.version_info < (3, 9), reason="Intermittent AssertionError on Python 3.8 tests"
+)
```
charlesbluca, line 51: Wonder if this flaky test is worth adding `pytest-rerunfailures` to our CI environments and marking as `pytest.mark.flaky`? cc @ayushdg for opinions

Also might want to update this condition to target to be based on the Fugue version instead of Python since we'll presumably still have this issue in the near future when 3.8 support is dropped and our mindeps testing is done with 3.9.

ayushdg: > Wonder if this flaky test is worth adding pytest-rerunfailures to our CI environments and marking as pytest.mark.flaky? cc @ayushdg for opinions

I think it's worth exploring that options and checking if rerunning just that test in a failed state multiple times results in a success. If so then it's a better alternative to skip. But if a single failure guarantees that subsequent re-runs of that test will fail as well then it's probably not as helpful.
For now might be worth adding flaky w reruns and if that still results in failures, we can skip on win/Mac on a fugue version condition.

> Also might want to update this condition to target to be based on the Fugue version instead of Python since we'll presumably still have this issue in the near future when 3.8 support is dropped and our mindeps testing is done with 3.9.

That makes sense, but just wanted to point out that the fugue version in our current mindeps testing used to work prior to the latest fugue release when it was a part of our 3.9/3.10 tests, so it's probably some combo of the dask version with fugue that leads to these flaky failures.

sarahyurick: Thanks, makes sense to me! I've added flaky with `@pytest.mark.flaky(reruns=2)`. Wasn't sure whether there should be more or less reruns.

It's also possible to include conditions, so would it make sense to do something like `@pytest.mark.flaky(reruns=2, condition="sys.version_info < (3, 9)")` (or it sounds like `fugue.__version__ < (7, 4)` would be preferred)?

Not sure how that's how the version would be formatted: https://github.com/dask-contrib/dask-sql/blob/main/continuous_integration/environment-3.8-dev.yaml#L8

charlesbluca: Based on @ayushdg's comments, I think it makes sense to go for `sys.version_info < (3, 9)` now since these failures seem to be a larger issue with the 3.8 testing environment versus Fugue-specific.

I think starting with a small number of reruns makes sense, we can always tweak this to be higher if we still find that things are failing (or remove if we find that things are failing consistently even with reruns)

charlesbluca (approved):
Thanks @sarahyurick!

PR #819, opened 2022-09-28
Enable crate sorting with `rustfmt`

charlesbluca:
This PR builds off of #817 to enable crate sorting through `rustfmt`, a feature currently only contained in nightly packages.

PR #1023, opened 2023-01-31
Fix Rust `format!` clippy errors

sarahyurick:
See: https://rust-lang.github.io/rust-clippy/master/index.html#uninlined_format_args


charlesbluca (approved):
Thanks @sarahyurick! Given the fact that this isn't the first time where CI has broken due to changes in the nightly Rust toolchain, wonder if we should also opt to pin to a specific version and bump periodically (would be preferable to have this automated somehow).

Think that this should be controlled with this step:

https://github.com/dask-contrib/dask-sql/blob/910fe199c03316827f00c98b6f540b83d0fddba0/.github/workflows/style.yml#L18-L21

sarahyurick:
> https://github.com/dask-contrib/dask-sql/blob/910fe199c03316827f00c98b6f540b83d0fddba0/.github/workflows/style.yml#L18-L21

Is there a reason we aren't using the stable version?

charlesbluca:
Yeah this was because we wanted to introduce sorting for crate statements as part of our linting process, which at the time was only supported with the nightly version of `rustfmt` (xref https://github.com/dask-contrib/dask-sql/pull/819); it's unclear to me if this is still the case now, though theoretically at some point in the future we shouldn't need that step once this feature becomes part of the stable toolchain.

PR #1025, opened 2023-02-01
Improve scalar logic for timestamps

### LINKED ISSUE #982: Timestamp improvements

sarahyurick:
Closes https://github.com/dask-contrib/dask-sql/issues/982

After some investigation, I found inconsistencies in our timestamp logic regarding whether scalars are interpreted as being in seconds versus in nanoseconds. Since Spark and PostgreSQL both interpret scalars as seconds, the changes in this PR reflect that.


tests/integration/test_rex.py
```diff
@@ -1015,3 +1015,114 @@ def test_totimestamp(c, gpu):
...
+    df1 = c.sql(f"SELECT TIMESTAMPADD(DAY, 5, to_timestamp({scalar1})) AS dt")
+    df2 = c.sql(f"SELECT TIMESTAMPADD(DAY, 5, {scalar1}) AS dt")
+    df3 = c.sql(f"SELECT TIMESTAMPADD(DAY, 5, CAST({scalar1} AS TIMESTAMP)) AS dt")
+    assert_eq(df1, expected_df)
+    assert_eq(df2, expected_df)
+    assert_eq(df3, expected_df)
+
+    expected_df = pd.DataFrame({"day": [15]})
+    df1 = c.sql(f"SELECT EXTRACT(DAY FROM to_timestamp({scalar1})) AS day")
+    # df2 = c.sql(f"SELECT EXTRACT(DAY FROM CAST({scalar1} AS TIMESTAMP)) AS day")
```
sarahyurick, line 1102: Not really sure what's going on in this case. When we're using a column of scalars, everything behaves as expected. But when we're just using a scalar like this query, the value doesn't appear to go through [`CastOperation`](https://github.com/dask-contrib/dask-sql/blob/2a8e9582de7b9a2519244796574054f1866222c7/dask_sql/physical/rex/core/call.py#L237) or [`ExtractOperation`](https://github.com/dask-contrib/dask-sql/blob/2a8e9582de7b9a2519244796574054f1866222c7/dask_sql/physical/rex/core/call.py#L596), so I'm not really sure where I can catch the integer and convert it to seconds.

tests/integration/test_rex.py
```diff
@@ -1015,3 +1015,128 @@ def test_totimestamp(c, gpu):
...
+    assert_eq(df2, expected_df)
+    df3 = c.sql(f"SELECT TIMESTAMPADD(DAY, 5, CAST({scalar1} AS TIMESTAMP)) AS dt")
+    assert_eq(df3, expected_df)
+
+    expected_df = pd.DataFrame({"day": [15]})
+    df1 = c.sql(f"SELECT EXTRACT(DAY FROM to_timestamp({scalar1})) AS day")
+    assert_eq(df1, expected_df)
+    # TODO: Fix seconds/nanoseconds conversion
+    # df2 = c.sql(f"SELECT EXTRACT(DAY FROM CAST({scalar1} AS TIMESTAMP)) AS day")
+    # assert_eq(df2, expected_df)
```
sarahyurick, line 1118: Not really sure what's going on in this case. When we're using a column of scalars, everything behaves as expected. But when we're just using a scalar like this query, the value doesn't appear to go through [`CastOperation`](https://github.com/dask-contrib/dask-sql/blob/2a8e9582de7b9a2519244796574054f1866222c7/dask_sql/physical/rex/core/call.py#L237) or [`ExtractOperation`](https://github.com/dask-contrib/dask-sql/blob/2a8e9582de7b9a2519244796574054f1866222c7/dask_sql/physical/rex/core/call.py#L596), so I'm not really sure where I can catch the integer and convert it to seconds.

sarahyurick:
Doesn't look like the changes from https://github.com/dask/dask/pull/9881 are being pulled yet. Once we upgrade to the newest Dask version this PR should be good to go.

cc @charlesbluca

sarahyurick:
Ready for re-review

dask_sql/physical/rex/core/call.py
```diff
@@ -51,6 +51,10 @@ def as_timelike(op):
         raise ValueError(f"Don't know how to make {type(op)} timelike")
 
 
+def is_timestamp_nano(obj):
+    return "int" in str(type(obj)) or "int" in str(getattr(obj, "dtype", ""))
```
charlesbluca, line 55: Would it make sense to use something like `pd.api.types.is_integer_dtype` for this check, or is there a specific case I'm missing?

dask_sql/physical/rex/core/call.py
```diff
@@ -634,7 +635,10 @@ def to_timestamp(self, df, format):
         else:
             if format != default_format:
                 raise RuntimeError("Integer input does not accept a format argument")
-            return dd.to_datetime(df, unit="s")
+            if is_cudf_type(df):
+                return df
```
charlesbluca, line 639: Looking into the GPU failures, it seems like in some cases by skipping over the `to_datetime` call here means we end up passing an integer series to `convert_to_datetime` and erroring.

Is there a particular case where we wouldn't want to call `to_datetime` here?

dask_sql/physical/rex/core/call.py
```diff
@@ -699,11 +706,23 @@ def __init__(self):
...
 
     def datetime_sub(self, unit, df1, df2):
+        if is_timestamp_nano(df1):
+            df1 = df1 * 10**9
+        if is_timestamp_nano(df2):
+            df2 = df2 * 10**9
+        if "datetime64[s]" == str(getattr(df1, "dtype", "")):
+            df1 = df1.astype("datetime64[ns]")
+        if "datetime64[s]" == str(getattr(df2, "dtype", "")):
+            df2 = df2.astype("datetime64[ns]")
```
charlesbluca, line 716: Getting a little tied up in the logic here, but think we might also want to scale the `result` based on the values of `is_timestamp_nano(df1|df2)`?

dask_sql/physical/rex/core/call.py
sarahyurick: Yes, the case that fails is
```
df = pd.DataFrame({"d1": [1203073300], "d2": [1503073700]})
c.create_table("df", df, gpu=gpu)
expected_df = pd.DataFrame({"dt": [3472]})
df1 = c.sql(
    "SELECT TIMESTAMPDIFF(DAY, to_timestamp(d1), to_timestamp(d2)) AS dt FROM df"
)
df1.compute()
```
which returns 0 instead of 3472 in the GPU case when `to_datetime` is used here. I'm not really sure why this happens, and there wasn't an obvious type check to avoid this. But you're right, it makes more sense to call `to_datetime` either way, so for now we can just skip that case.

dask_sql/physical/rex/core/call.py
sarahyurick: I'm not really sure what you mean here? Whenever we do a subtraction operation (like with `timestampdiff`), this extra logic is to ensure uniformity so that we're working in nanoseconds.

sarahyurick:
Thanks @charlesbluca ! Lmk what you think.

dask_sql/physical/rex/core/call.py
```diff
@@ -52,6 +52,10 @@ def as_timelike(op):
         raise ValueError(f"Don't know how to make {type(op)} timelike")
 
 
+def is_timestamp_nano(obj):
+    return pd.api.types.is_integer_dtype(obj)
```
charlesbluca, line 56: I think my preference here would be to use the `is_integer_dtype` check directly in place of `is_timestamp_nano` since we aren't doing any additional checks on `obj` here

dask_sql/physical/rex/core/call.py
```diff
@@ -656,7 +657,10 @@ def timestampadd(self, unit, interval, df: SeriesOrScalar):
         interval = int(interval)
         if interval < 0:
             raise RuntimeError(f"Negative time interval {interval} is not supported.")
-        df = df.astype("datetime64[ns]")
+        if is_timestamp_nano(df):
+            df = df.astype("datetime64[s]")
+        else:
+            df = df.astype("datetime64[ns]")
```
charlesbluca, line 663: Could condense this to something like

```suggestion
        df = df.astype("datetime64[s]" if is_timestamp_nano(df) else "datetime64[ns]")
```

tests/integration/test_rex.py
```diff
@@ -1121,3 +1107,134 @@ def test_extract_date(c, gpu):
...
+    df2 = c.sql("SELECT FLOOR(CAST(d AS TIMESTAMP) TO DAY) AS floor_to_day FROM df")
+    assert_eq(df2, expected_df)
+
+    df = pd.DataFrame({"d1": [1203073300], "d2": [1503073700]})
+    c.create_table("df", df, gpu=gpu)
+    expected_df = pd.DataFrame({"dt": [3472]})
+    df1 = c.sql(
+        "SELECT TIMESTAMPDIFF(DAY, to_timestamp(d1), to_timestamp(d2)) AS dt FROM df"
+    )
+    # TODO: The GPU case returns an incorrect value here
```
charlesbluca, line 1183: Could you file an issue to track these failures on GPU (and the seconds/nanoseconds conversion issue below)?

charlesbluca (approved):
Thanks @sarahyurick! This LGTM, just one minor comment around opening issues to track the remaining failures

tests/integration/test_rex.py
sarahyurick: Yes, just opened https://github.com/dask-contrib/dask-sql/issues/1187. Thanks!

dask_sql/physical/rex/core/call.py
```diff
@@ -138,6 +138,9 @@ def reduce(self, *operands, **kwargs):
                 )
             ):
                 operands = tuple(map(as_timelike, operands))
+            if is_cudf_type(operands[0]) and isinstance(operands[1], np.timedelta64):
+                operands = (dd.to_datetime(operands[0], unit="s"), operands[1])
```
ayushdg, line 142: Curious if this is still needed. I stepped through the tests and didn't need this change as the operands were already in unit `'s` from `to_timestamp`

ayushdg (approved):
Minor question but lgtm!

dask_sql/physical/rex/core/call.py
sarahyurick: Good point, thanks! I think now the gpuCI failures are just ML failures unrelated to this PR, as the same failures are appearing in https://github.com/dask-contrib/dask-sql/pull/1197 right now too.

ayushdg:
Merging in since the failures are unrelated.

PR #1027, opened 2023-02-02
Implement join reordering of fact-dimension joins

sarahyurick:
Reopening #950 here.

This PR implements a new logical plan optimization rule based on the paper [Improving Join Reordering for Large Scale Distributed Computing](https://ieeexplore.ieee.org/document/9378281).


dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,385 @@
...
+    /// filtered relation
+    filter_selectivity: f64,
+}
+
+impl Default for JoinReorder {
+    fn default() -> Self {
+        Self {
+            max_fact_tables: 2,
+            // FIXME: fact_dimension_ratio should be 0.3
+            fact_dimension_ratio: 0.7,
```
sarahyurick, line 29: Ideally, this should be 0.3, but queries 17, 25, 29, and 85 currently fail without a stricter ratio. This suggests that more work needs to be done with reordering fact-to-fact joins.

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,385 @@
...
+            for key in join_keys {
+                conds.remove(&key);
+            }
+
+            /* FIXME: Build join with join_keys when needed
+            self.join(
+                right: LogicalPlan,
+                join_type: JoinType,
+                join_keys: (Vec<impl Into<Column>>, Vec<impl Into<Column>>),
+                filter: Option<Expr>,
```
sarahyurick, line 362: Probably the biggest improvement we need is to support join filters correctly. This should allow us to run query 72, which is expected to have the largest performance gain with join reordering. Other queries affected include queries 75 and 93.

sarahyurick (commented):
While this PR does not work for all cases, it implements basic join reordering logic. I've opened https://github.com/dask-contrib/dask-sql/issues/1069 to address additional features in the future.

For now, though, I'll be focusing on dynamic partition pruning, so I wanted to open this PR up for review.

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,385 @@
...
+    /// Plan containing the table scan for the fact or dimension table
+    /// May also contain Filter and SubqueryAlias
+    plan: LogicalPlan,
+    /// Estimated size of the underlying table before any filtering is applied
+    size: usize,
+}
+
+impl Relation {
+    fn new(plan: LogicalPlan) -> Self {
+        let size = get_table_size(&plan).unwrap_or(100);
```
jdye64, line 194: I feel like a warning should be issued here if we default to 100. I could see situations where undesired optimizing could occur if this defaulted to 100. This isn't something we can "fix" but we should alert the user that it is happening so they can be aware.

jdye64 (changes requested):
Great PR. This is a complicated bit of logic but was well written and easy to follow/understand. I would like to see a user facing warning when the statistics defaults to 100 for the row count just so users don't get blindsided by unexpected optimizations. Otherwise this is great.

dask_planner/src/sql/optimizer/join_reorder.rs
sarahyurick: Thanks for reviewing! Updated.

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,392 @@
+//! Join reordering based on the paper "Improving Join Reordering for Large Scale Distributed Computing"
+//! https://ieeexplore.ieee.org/document/9378281
```
charlesbluca, line 2: Do we have a link to the modified join reorder algorithm implemented in spark-rapids that we could also include here?

charlesbluca (commented):
Thanks @sarahyurick! Just a few initial comments before I dig into the bulk of the algorithm itself:

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,392 @@
...
+    /// Maximum number of fact tables to allow in a join
+    max_fact_tables: usize,
+    /// Ratio of the size of the dimension tables to fact tables
+    fact_dimension_ratio: f64,
+    /// Whether to preserve user-defined order of unfiltered dimensions
+    preserve_user_order: bool,
+    /// Constant to use when determining the number of rows produced by a
+    /// filtered relation
+    filter_selectivity: f64,
+}
```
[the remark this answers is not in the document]
sarahyurick: That's not something I had thought of, but I don't see why not. Could be a nice feature.

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,392 @@
...
+
+    if !is_supported_join(join) {
+        return Ok(Some(plan.clone()));
+    }
+
+    // Extract the relations and join conditions
+    let (rels, conds) = extract_inner_joins(plan);
+
+    // Split rels into facts and dims
+    let rels: Vec<Relation> = rels.into_iter().map(Relation::new).collect();
```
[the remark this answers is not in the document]
sarahyurick: Nope, could totally be done in the `extract_inner_joins` function. Do you have a preference here?

dask_planner/src/sql/optimizer/join_reorder.rs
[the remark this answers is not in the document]
charlesbluca: Mostly thinking about if this is something we would expect end-users to want to tune manually, or if it's something we expect to have a sensible default for that tends to just work - definitely okay with handling something like this in a follow-up if it's worthwhile

dask_planner/src/sql/optimizer/join_reorder.rs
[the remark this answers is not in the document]
charlesbluca: My preference would be to do this in `extract_inner_joins` (if doing so isn't particularly nasty), since we seemingly don't have any use for the returned `Vec<LogicalPlan>` past this point

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,392 @@
...
+    let largest_rel = rels.iter().map(|rel| rel.size).max().unwrap() as f64;
```
ayushdg, line 83: Would it be possible to unwrap or raise an error since we expect `rels` to have atleast 1 element.

ayushdg (changes requested):
Could we add a couple of test cases here with different table sizes to assert things are re-ordering as expected? No strong preference on whether it should be python or a rust test.

dask_planner/src/sql/optimizer/join_reorder.rs
sarahyurick: I can't really imagine a case where we'd run into this - we've already checked that there are joins in the query, and in the case where we don't know the table sizes it defaults to 100 and issues a warning. So we can always expect at least 2 tables with (in the default case scenario) the largest size being 100.

dask_planner/src/sql/optimizer/join_reorder.rs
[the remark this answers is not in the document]
sarahyurick: Added to https://github.com/dask-contrib/dask-sql/issues/1069

sarahyurick:
Thanks @charlesbluca and @ayushdg ! Updated from your suggestions, lmk what you think.

randerzander (approved):
I compared Dask-SQL main vs this PR on a set of internal benchmark queries. This PR is about 3% faster overall. Any query where perf was worse looks within typical noise range.

tests/integration/test_join.py
```diff
@@ -402,6 +403,50 @@ def test_filter_columns_post_join(c):
     assert_eq(result_df, expected_df)
 
 
+def test_join_reorder(c):
```
charlesbluca, line 406: Could we add some tests around other potential join cases we would run into? In particular maybe some with:

- unsupported join types / conditions
- joins that would result in more than 2 dominant fact tables
-

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,392 @@
...
+                result.push(unfiltered_dimensions.remove(0));
+            }
+        } else if !filtered_dimensions.is_empty() {
+            result.push(filtered_dimensions.remove(0));
+        } else {
+            result.push(unfiltered_dimensions.remove(0));
+        }
+    }
+    assert!(filtered_dimensions.is_empty());
+    assert!(unfiltered_dimensions.is_empty());
```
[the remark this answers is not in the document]
sarahyurick: Yeah, they can be removed now.

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,392 @@
...
+    for cond in &conds {
+        match cond {
+            (Expr::Column(l), Expr::Column(r)) => {
+                join_conds.insert((l.clone(), r.clone()));
+            }
+            _ => {
+                return Ok(Some(plan.clone()));
+            }
+        }
+    }
```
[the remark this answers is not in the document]
sarahyurick: Sure, that makes sense to me!

tests/integration/test_join.py
sarahyurick: The logic for joins involving different combinations of numbers of fact tables versus numbers of dimension tables is a little shaky (especially depending on the number of fact tables and fact-fact joins), so I'd prefer to hold off there until I have a better solution for multiple fact tables. I have this listed in https://github.com/dask-contrib/dask-sql/issues/1069

I can work on examples for unsupported join types and conditions, though.

dask_planner/src/sql/optimizer/join_reorder.rs
```diff
@@ -0,0 +1,392 @@
...
+        // If the ratio is larger than the fact_dimension_ratio, it is a fact table
+        // Else, it is a dimension table
+        if rel.size as f64 / largest_rel > rule.fact_dimension_ratio {
+            facts.push(rel.clone());
+        } else {
+            dims.push(rel.clone());
+        }
+    }
+
+    if facts.is_empty() || dims.is_empty() {
```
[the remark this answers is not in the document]
sarahyurick: I tried removing `facts.is_empty()` and ended up with a couple of PyTest failures. You're right that it's serving basically the same purpose as having `rels.is_empty()` would.

sarahyurick:
Thanks @charlesbluca ! I've updated from your suggestions. I also added an additional test: by default, join reordering should only be done with filtered tables, so we demonstrate that the plan remains unchanged with unfiltered tables. Let me know if there's any additional conditions we should check.

charlesbluca (approved):
Thanks for addressing all my comments @sarahyurick! Overall this LGTM, though would also like to get a thumbs up from @jdye64 and/or @ayushdg here who have a little more experience with optimizers

ayushdg (approved):

PR #1030, opened 2023-02-06
Break off source distribution publication to separate release job

charlesbluca:
Breaks off the building and publishing of source distribution into its own job as part of the release workflow. This should make it so we can move forward with conda-forge release while the wheels are being built, which should cut down on release time significantly.

cc @ayushdg

### DIFF
--- a/.github/workflows/release.yml
+++ b/.github/workflows/release.yml
@@ -9,8 +9,8 @@ defaults:
     shell: bash -l {0}
 
 jobs:
-  build:
-    name: build py3.${{ matrix.python }} on ${{ matrix.os }}
+  wheels:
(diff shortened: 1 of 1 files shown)


.github/workflows/release.yml
```diff
@@ -78,3 +78,31 @@ jobs:
           TWINE_USERNAME: ${{ secrets.PYPI_USERNAME }}
           TWINE_PASSWORD: ${{ secrets.PYPI_PASSWORD }}
         run: twine upload dist/*
```
ayushdg, line 80: Can get rid of the build source dist step and the checks downstream

ayushdg (changes requested):

ayushdg (approved):

Issue #994, opened 2023-01-17
[ENH] Collect and track table statistics when applicable


ayushdg:
**Is your feature request related to a problem? Please describe.**
When creating tables that were read from parquet files, dask-sql can also collect and keep partition wise statistics that can be used downstream for future optimizations (like join reordering, or a heuristics to decide on physical plan paths like join algorithms)

**Describe the solution you'd like**
- [x] Add a utility to dask-sql that has the ability to return partition wise statistics (wip #999)
- [ ] Update DataContainer to hold these stats (and populate them if applicable) after table creation (or table_scan)
- [ ] Update provide these stats to `DaskTable` and make it accessible via rust (for optimizer rules) 

**Additional context**
Required for the implementation of #950 

Initially the scope of this can be for parquet files only and can be expanded in the future to other file types.

sarahyurick:
Will be exploring this as I work on https://github.com/dask-contrib/dask-sql/pull/1027

sarahyurick:
Related to https://github.com/dask-contrib/dask-sql/issues/851

Issue #851, opened 2022-10-11
[ENH] Support table statistics based query optimization rules


randerzander:
As Dask [adds support for gathering dataset statistical metadata](https://github.com/dask/dask/pull/9473), Dask-SQL should look into passing these stats into DataFusion's query planner/optimizer, which should result in more efficient query plans.

PR #1037, opened 2023-02-09
Table Statistics Support

sarahyurick:
Closes https://github.com/dask-contrib/dask-sql/issues/994
Closes https://github.com/dask-contrib/dask-sql/issues/851

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -63,6 +63,7 @@ use crate::{
             show_tables::ShowTablesPlanNode,
             PyLogicalPlan,
         },
+        table::DaskStatistics,
     },
 };
 
@@ -479,18 +480,27 @@ impl DaskSQLContext {
             .map_err(py_parsing_exp)
     }
 
-    pub fn get_table_statistics(&self) -> HashMap<String, f64> {
-        let mut table_statistics: HashMap<String, f64> = HashMap::new();
+    pub fn get_table_statistics(&self) -> Option<HashMap<String, DaskStatistics>> {
+        let mut table_statistics: HashMap<String, DaskStatistics> = HashMap::new();
+        // We use all_row_counts to ensure that we have statistics
+        // for at least one of the tables in the DaskSQLContext
+        let mut all_row_counts = 0.0;
+
         for schema in &self.schemas {
             for table in &schema.1.tables {
                 let dask_table = &table.1;
-                // TODO: Use qualified name
                 let table_name = &dask_table.table_name;
-                let statistics = dask_table.statistics.get_row_count().unwrap();
-                table_statistics.insert(table_name.to_string(), statistics);
+                let statistics = &dask_table.statistics;
+                table_statistics.insert(table_name.to_string(), statistics.clone());
+                all_row_counts += statistics.get_row_count().unwrap();
             }
         }
-        table_statistics
+
+        if all_row_counts > 0.0 {
+            Some(table_statistics)
+        } else {
+            None
+        }
     }
 
     /// Accepts an existing relational plan, `LogicalPlan`, and optimizes it
@@ -507,7 +517,7 @@ impl DaskSQLContext {
         match existing_plan.original_plan.accept(&mut visitor) {
             Ok(valid) => {
                 if valid {
-                    optimizer::DaskSqlOptimizer::new(true, Some(statistics))
+                    optimizer::DaskSqlOptimizer::new(true, statistics)
                         .optimize(existing_plan.original_plan)
                         .map(|k| PyLogicalPlan {
                             original_plan: k,
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -130,11 +130,21 @@ impl ContextProvider for DaskSQLContext {
                 // If the Table is not found return None. DataFusion will handle the error propagation
                 match resp {
                     Some(e) => {
-                        let statistics = &self.schemas.get(reference.schema).unwrap().tables.get(reference.table).unwrap().statistics;
+                        let statistics = &self
+                            .schemas
+                            .get(reference.schema)
+                            .unwrap()
+                            .tables
+                            .get(reference.table)
+                            .unwrap()
+                            .statistics;
                         if statistics.get_row_count() == 0.0 {
                             Ok(Arc::new(table::DaskTableSource::new(Arc::new(e), None)))
                         } else {
-                            Ok(Arc::new(table::DaskTableSource::new(Arc::new(e), Some(statistics.clone()))))
+                            Ok(Arc::new(table::DaskTableSource::new(
+                                Arc::new(e),
+                                Some(statistics.clone()),
+                            )))
                         }
                     }
                     None => Err(DataFusionError::Plan(format!(
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -63,7 +63,6 @@ use crate::{
             show_tables::ShowTablesPlanNode,
             PyLogicalPlan,
         },
-        table::DaskStatistics,
     },
 };
 
@@ -130,7 +129,14 @@ impl ContextProvider for DaskSQLContext {
 
                 // If the Table is not found return None. DataFusion will handle the error propagation
                 match resp {
-                    Some(e) => Ok(Arc::new(table::DaskTableSource::new(Arc::new(e)))),
+                    Some(e) => {
+                        let statistics = &self.schemas.get(reference.schema).unwrap().tables.get(reference.table).unwrap().statistics;
+                        if statistics.get_row_count() == 0.0 {
+                            Ok(Arc::new(table::DaskTableSource::new(Arc::new(e), None)))
+                        } else {
+                            Ok(Arc::new(table::DaskTableSource::new(Arc::new(e), Some(statistics.clone()))))
+                        }
+                    }
                     None => Err(DataFusionError::Plan(format!(
                         "Table '{}.{}.{}' not found",
                         reference.catalog, reference.schema, reference.table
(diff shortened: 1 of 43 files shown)


dask_planner/src/sql/optimizer.rs
```diff
@@ -34,13 +34,15 @@ use filter_columns_post_join::FilterColumnsPostJoin;
...
     optimizer: Optimizer,
 }
 
 impl DaskSqlOptimizer {
     /// Creates a new instance of the DaskSqlOptimizer with all the DataFusion desired
     /// optimizers as well as any custom `OptimizerRule` trait impls that might be desired.
-    pub fn new(skip_failing_rules: bool) -> Self {
+    pub fn new(skip_failing_rules: bool, statistics: HashMap<String, f64>) -> Self {
         let rules: Vec<Arc<dyn OptimizerRule + Sync + Send>> = vec![
             Arc::new(InlineTableScan::new()),
```
sarahyurick, line 47: With these changes, we can pass table statistics into optimizer rules. For example, with https://github.com/dask-contrib/dask-sql/pull/1027 we could do something like `Arc::new(JoinReorder::new(2, 0.3, true, 1.0, statistics.clone()))` where `statistics` is just another one of the fields that we're passing into a new instance of the `JoinReorder` struct.

dask_sql/physical/utils/statistics.py
```diff
@@ -87,21 +87,13 @@ def parquet_statistics(
     if not isinstance(layer, DataFrameIOLayer) or not isinstance(
         layer.io_func, ParquetFunctionWrapper
     ):
-        logger.warning(
-            f"Could not extract Parquet statistics from {ddf}."
-            f"\nAttempted IO layer: {layer}"
```
sarahyurick, line 92: I removed the warnings because since we're always trying to capture statistics now, this warning would come up every time the user works with a non-Parquet DataFrame. For example:
```
df = pd.DataFrame(
    {
        "a": [1.0] * 100 + [2.0] * 200 + [3.0] * 400,
        "b": 10 * np.random.rand(700),
    }
)
c.create_table("df", df)

result_df = c.sql("SELECT * FROM df")
```

Would throw a warning, which wouldn't make sense to the user.

dask_sql/context.py
```diff
@@ -248,9 +253,28 @@ def create_table(
...
+            try:
+                stats = parquet_statistics(input_table)
+                row_count = 0
+                if stats:
+                    # List of dicts
+                    for d in stats:
+                        row_count += d["num-rows"]
+                    dc_statistics = Statistics(row_count)
+            except ValueError:
+                dc_statistics = None
```
sarahyurick, line 267: So one thing about this PR is that it only works on Dask DataFrames that were read in with `read_parquet` and then added to the DaskSQLContext with `c.create_table("ddf", ddf)`. However, in our TPC-DS tests, we are skipping `read_parquet()` and just doing `c.create_table(table_name, my_filepath, format="parquet", gpu=True)`, so `parquet_statistics()` doesn't work on it. If we wanted to, we could do something like:

```suggestion
                if format.lower() == "parquet":
                    if dask_cudf:
                        ddf = dask_cudf.read_parquet(input_table)
                    else:
                        ddf = dd.read_parquet(input_table)
                    stats = parquet_statistics(ddf)
                    row_count = 0
                    if stats:
                        # List of dicts
                        for d in stats:
                            row_count += d["num-rows"]
                        dc_statistics = Statistics(row_count)
```

Does this seem reasonable? If I'm understanding it correctly, since we're not calling `ddf.compute()` it should be OK performance-wise?

sarahyurick:
I ended up just keeping the statistics on the Dask-SQL side instead of using any DataFusion imports. It seems to me that DataFusion statistics are too limited at the moment and so I'm worried about blockages there, since issues like https://github.com/apache/arrow-datafusion/issues/3988 and https://github.com/apache/arrow-datafusion/issues/4625 are currently open.

dask_planner/src/sql.rs
```diff
@@ -479,6 +480,29 @@ impl DaskSQLContext {
...
+        // We use all_row_counts to ensure that we have statistics
+        // for at least one of the tables in the DaskSQLContext
+        let mut all_row_counts = 0.0;
+
+        for schema in &self.schemas {
+            for table in &schema.1.tables {
+                let dask_table = &table.1;
+                let table_name = &dask_table.table_name;
+                let statistics = &dask_table.statistics;
+                table_statistics.insert(table_name.to_string(), statistics.clone());
```
ayushdg, line 494: We might end up with incorrect stats if we have tables with the same name in different schemas. A fully qualified name if applicable might be a better key.

dask_planner/src/sql.rs
```diff
@@ -488,11 +512,12 @@ impl DaskSQLContext {
     ) -> PyResult<logical::PyLogicalPlan> {
         // Certain queries cannot be optimized. Ex: `EXPLAIN SELECT * FROM test` simply return those plans as is
         let mut visitor = OptimizablePlanVisitor {};
+        let statistics = self.get_table_statistics();
 
         match existing_plan.original_plan.accept(&mut visitor) {
             Ok(valid) => {
                 if valid {
-                    optimizer::DaskSqlOptimizer::new(true)
+                    optimizer::DaskSqlOptimizer::new(true, statistics)
```
ayushdg, line 520: I'm a bit conflicted b/w pre-loading every single table stat in a dictionary and passing the whole thing over to the optimizer config vs getting it from the specific rule that actually needs it.

ayushdg (changes requested):
A few general comments. 

More generally I was thinking if instead of front loading the whole stats and passing it off as a config it might be viable to add a `statistics` method to `DaskTableSource` that could be queried from within the optimizer rule itself, for the specific tables we're interesting in for that query. 
There might be pro's and cons to either approach. Curious if you have any thoughts.

dask_planner/src/sql.rs
ayushdg: Elaborating a bit more, being able to do something like this: https://github.com/dask-contrib/dask-sql/pull/1027/files#diff-9b258b399cd311b18a5b265d5bbdd0a73eb211a7f0975e967a4b17eae9a804a6R400 by just making modifications to `DaskTableSource`. Though I'm not familiar enough to comment on whether that approach will work purely in dask-sql without depending on any additional improvements to datafusion.

sarahyurick: Okay, thanks! I've updated the code with modifications to `DaskTableSource`. In PR #1027, it looks like it would work by having a function:
```
fn get_table_size(plan: &LogicalPlan) -> Option<usize> {
    match plan {
        LogicalPlan::TableScan(scan) => {
            let source = scan
                .source
                .as_any()
                .downcast_ref::<DaskTableSource>()
                .expect("should be a DaskTableSource");
            if let Some(stats) = source.statistics() {
                Some(stats.get_row_count() as usize)
            } else {
                None
            }
        }
        _ => get_table_size(plan.inputs()[0]),
    }
}
```
in the `join_reorder.rs` file. The function appears to work as desired, at least within the Join Reordering PR 🙂

Let me know what you think.

dask_planner/src/sql/table.rs
```diff
@@ -25,12 +25,20 @@ use crate::{
 /// DaskTable wrapper that is compatible with DataFusion logical query plans
 pub struct DaskTableSource {
     schema: SchemaRef,
+    #[allow(dead_code)]
```
ayushdg, line 28: Curious why this is needed

dask_sql/context.py
```diff
@@ -20,10 +20,16 @@
     LogicalPlan,
 )
 
+try:
+    from dask_sql.physical.utils.statistics import parquet_statistics
+except ModuleNotFoundError:
+    parquet_statistics = None
+
```
ayushdg, line 27: Are there cases where we wouldn't be able to import?

dask_sql/context.py
```diff
@@ -248,9 +254,37 @@ def create_table(
...
+                    row_count = 0
+                    for d in stats:
+                        row_count += d["num-rows"]
+                    dc_statistics = Statistics(row_count)
+            except ValueError:
+                if str(format).lower() == "parquet":
+                    if dask_cudf:
+                        ddf = dask_cudf.read_parquet(input_table)
+                    else:
+                        ddf = dd.read_parquet(input_table)
```
ayushdg, line 271: In what case do we need to go down this branch?

dask_sql/physical/utils/statistics.py
ayushdg: For simplicity we could try setting the level to `logger.info/logger.debug` which shouldn't be seen to the user unless they've set that specific logging level

ayushdg (changes requested):
A few followup questions. Thanks for the updates so far!

dask_planner/src/sql/table.rs
sarahyurick: Rust currently throws up warnings that the `statistics` field isn't being used anywhere else (which shows up in our CI), but with the Join Reordering PR the warning will go away.

dask_sql/context.py
sarahyurick: I added this because our bare requirements test errored with:
```
import pyarrow.parquet as pq
ModuleNotFoundError: No module named 'pyarrow'
```

dask_planner/src/sql.rs
ayushdg: looks good to me!

dask_sql/context.py
sarahyurick: This handles if we were doing `c.create_table(table_name, my_filepath, format="parquet", gpu=True)` (instead of having a `read_parquet` then `c.create_table("ddf", ddf)`.

dask_sql/context.py
sarahyurick: This was with my first commit to this PR https://github.com/dask-contrib/dask-sql/actions/runs/4129811432/jobs/7135835860

ayushdg: Ah I see. I think this is okay for now then. cc: @charlesbluca if you have thoughts

sarahyurick:
Updated. Thanks @ayushdg !

dask_sql/context.py
```diff
@@ -246,12 +251,25 @@ def create_table(
...
+            statistics = parquet_statistics(dc.df)
+            if statistics:
+                row_count = 0
+                for d in statistics:
+                    row_count += d["num-rows"]
+                statistics = Statistics(row_count)
+                self.schema[schema_name].statistics[table_name.lower()] = statistics
+
+        # If no statistics are obtainable, we will just assume 100 rows
+        if not statistics:
```
ayushdg, line 267:
```suggestion
        else:
```

sarahyurick: I originally had `else`, but I don't think that would cover all cases. If we get into the `elif` block (meaning that `parquet_statistics()` is importable) but nothing is returned by `parquet_statistics()`, then a value for `statistics` would never be assigned and fail a couple lines later with `statistics.row_count`.

We could do 2 `else` blocks (one associated with the inner `if statistics` and the other with the outer `if statistics`/`elif parquet_statistics`) but this way is less lines of code.

ayushdg: Thanks for the explanation that makes sense to me!

ayushdg (approved):
Changes lgtm! 

In a followup pr, we could also move this assignment below https://github.com/dask-contrib/dask-sql/pull/1037/files#diff-fe785e602208037be3dc2ce47281e665a73077cce4305d951759f5e4e2da81faR253
and create a dc with the statistics set on the dc as well. 
This would allow adding some tests on the python side that can test if we're keeping these results as expected

Issue #1044, opened 2023-02-13
[ENH] Bump Datafusion -> 18.0.0


jdye64:
**Is your feature request related to a problem? Please describe.**
With the release of DataFusion 18.0.0 we should bump our dependency version to `18.0.0`

**Describe the solution you'd like**
Bump dependency version and resolve any conflicts or build issues.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #1046, opened 2023-02-13
[REVIEW] Bump to datafusion-18

jdye64:
This closes: #1044


dask_planner/src/sql/logical.rs
```diff
@@ -368,6 +368,7 @@ impl PyLogicalPlan {
                     "Extension"
                 }
             }
+            LogicalPlan::Unnest(_unnest) => "UnNest",
```
sarahyurick, line 371: I don't think it makes a difference, but for consistency with DF the capitalization is `Unnest`:

```suggestion
            LogicalPlan::Unnest(_unnest) => "Unnest",
```

dask_planner/src/sql/table.rs
```diff
@@ -188,7 +188,7 @@ pub(crate) fn table_from_logical_plan(
 
             let table_ref: TableReference = table_scan.table_name.as_str().into();
             let (schema, tbl) = match table_ref {
-                TableReference::Bare { table } => ("", table),
+                TableReference::Bare { table } => (Cow::from(""), table),
```
sarahyurick, line 191: Cool, I guess the other choice is to do:
```
TableReference::Bare { table } => ("".to_string(), table),
                TableReference::Partial { schema, table } => (schema.to_string(), table),
                TableReference::Full {
                    catalog: _,
                    schema,
                    table,
                } => (schema.to_string(), table),
```
to avoid the extra import, but I don't really have a strong opinion one way or another.

sarahyurick (commented):
I left some minor comments, but otherwise LGTM!

sarahyurick (approved):
Looks good to me!

ayushdg (approved):

PR #1048, opened 2023-02-15
Remove call to `nest_asyncio.apply` in server startup

charlesbluca:
Attempting to unblock some issues starting up the server with `run_server(blocking=True)` reported by @randerzander; seems like these crop up when we attempt to patch the event loop, interested in how CI will respond to us removing this patch.

### DIFF
--- a/dask_sql/server/app.py
+++ b/dask_sql/server/app.py
@@ -6,7 +6,6 @@ from uuid import uuid4
 import dask.distributed
 import uvicorn
 from fastapi import FastAPI, HTTPException, Request
-from nest_asyncio import apply
 from uvicorn import Config, Server
 
 from dask_sql.context import Context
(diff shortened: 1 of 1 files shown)


ayushdg (approved):
Does this unblock a few tests?

charlesbluca:
Don't think this has an impact on the server tests since those are accessing the server through the `fastapi.testclient.TestClient` and so we don't actually start up a uvicorn server; not sure if there's any larger reasoning behind the decision to use the `TestClient` there but could be worthwhile to explore actually spinning up and shutting down a server for the tests to make sure things work as expected there.

Should also note that some of the failures are pretty trivial, and just come down to differences in the expected column / dtype names, if we begin pushing on updating server support those shouldn't be too difficult to get passing again.

PR #1049, opened 2023-02-15
Resolve tests in `test_server.py`

charlesbluca:
This PR resolves the failures in `test_server.py` by:

- modifying the method of extracting the string SQL type for columns in `get_column_description `
- commenting out `errorLocation` functionality/checks for now
- modifying `test_sql_query` to account for changes to scalar integer behavior with the DataFusion parser

### DIFF
--- a/dask_planner/src/sql/types.rs
+++ b/dask_planner/src/sql/types.rs
@@ -120,6 +120,10 @@ impl DaskTypeMap {
         })
     }
 
+    fn __str__(&self) -> String {
+        format!("{:?}", self.sql_type)
+    }
+
     #[pyo3(name = "getSqlType")]
     pub fn sql_type(&self) -> SqlTypeName {
         self.sql_type.clone()
--- a/dask_sql/server/responses.py
+++ b/dask_sql/server/responses.py
@@ -64,9 +64,7 @@ class QueryResults:
 class DataResults(QueryResults):
     @staticmethod
     def get_column_description(df):
-        sql_types = [
-            str(python_to_sql_type(t).getSqlType()).partition(".")[2] for t in df.dtypes
-        ]
+        sql_types = [str(python_to_sql_type(t)) for t in df.dtypes]
         column_names = df.columns
         return [
             {
--- a/dask_sql/server/responses.py
+++ b/dask_sql/server/responses.py
@@ -64,7 +64,9 @@ class QueryResults:
 class DataResults(QueryResults):
     @staticmethod
     def get_column_description(df):
-        sql_types = [str(python_to_sql_type(t)) for t in df.dtypes]
+        sql_types = [
+            str(python_to_sql_type(t).getSqlType()).partition(".")[2] for t in df.dtypes
+        ]
         column_names = df.columns
         return [
             {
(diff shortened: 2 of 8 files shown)


dask_sql/server/responses.py
```diff
@@ -64,7 +64,9 @@ def __init__(self, request: Request, next_url: str = None, cancel_url: str = Non
 class DataResults(QueryResults):
     @staticmethod
     def get_column_description(df):
-        sql_types = [str(python_to_sql_type(t)) for t in df.dtypes]
+        sql_types = [
+            str(python_to_sql_type(t).getSqlType()).partition(".")[2] for t in df.dtypes
```
charlesbluca, line 68: Ideally we should add a `toString` method to either `DaskTypeMap` or `SqlTypeName`, but I'm interested in if we can define a dunder method through pyo3 so that the original code would just work here?

ayushdg: It looks like this might be doable? https://pyo3.rs/main/class/object.html?highlight=__str__#string-representations

dask_planner/src/sql/types.rs
```diff
@@ -120,6 +120,10 @@ impl DaskTypeMap {
         })
     }
 
+    fn __str__(&self) -> String {
+        format!("{:?}", self.sql_type)
```
jdye64, line 124: This doesn't need to be a "Debug" print statement but no harm in leaving it this way.

dask_sql/server/responses.py
```diff
@@ -131,10 +131,11 @@ def __init__(self, error: Exception):
...
         self.errorType = "USER_ERROR"
 
-        try:
-            self.errorLocation = {
-                "lineNumber": error.from_line + 1,
-                "columnNumber": error.from_col + 1,
-            }
-        except AttributeError:  # pragma: no cover
-            pass
+        # FIXME: ParserErrors currently don't contain information on where the syntax error occurred
```
jdye64, line 134: Yeah this is something that will require changes far up the third party stream in SqlParser-RS and then later in DataFusion. We can address this later however.

jdye64 (approved):
LGTM thanks @charlesbluca

dask_sql/server/responses.py
randerzander: cc @sarahyurick in case you have a minute to look into that, given your other work on the parser side of things

dask_planner/src/sql/types.rs
charlesbluca: Yeah mostly did this to get around the need to implement `Display` for `SqlTypeName`, without which we end up getting a compiler error

tests/unit/test_context.py
```diff
@@ -184,24 +182,18 @@ def test_function_adding():
     assert c.schema[c.schema_name].function_lists[0].name == "F"
     assert c.schema[c.schema_name].function_lists[0].parameters[0][0] == "x"
     assert (
-        c.schema[c.schema_name].function_lists[0].parameters[0][1].getSqlType()
-        == int_sql_type
-    )
-    assert (
-        c.schema[c.schema_name].function_lists[0].return_type.getSqlType()
-        == float_sql_type
+        str(c.schema[c.schema_name].function_lists[0].parameters[0][1])
```
ayushdg, line 185: Not needed for this PR, but I wonder if it would be better to start being more explicit in our function registration examples to use `int32` or `int64` to avoid the platform specific ambiguity.

ayushdg (approved):
lgtm! Thanks

PR #1052, opened 2023-02-16
Split gpuCI environment into 3.8/3.10 environments

charlesbluca:
With #1042, we now run GPU tests on python 3.8/3.10; this PR splits our existing 3.9 environment file to recreate the GPU CI environment into 2 files for the 3.8 and 3.10 environments.

### DIFF
--- a/continuous_integration/gpuci/environment-3.8.yaml
+++ b/continuous_integration/gpuci/environment-3.8.yaml
@@ -7,37 +7,36 @@ channels:
 - nodefaults
 dependencies:
 - dask>=2022.3.0
-# FIXME: handling is needed for httpx-based fastapi>=0.87.0
-- fastapi>=0.69.0,<0.87.0
-- fugue>=0.7.3
-- intake>=0.6.0
+- fastapi=0.69.0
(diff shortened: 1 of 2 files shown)


continuous_integration/gpuci/environment-3.8.yaml
```diff
@@ -0,0 +1,49 @@
+name: dask-sql
+channels:
+- rapidsai
+- rapidsai-nightly
+- nvidia
+- conda-forge
+- nodefaults
+dependencies:
+- dask>=2022.3.0
```
ayushdg, line 9: Is this file checking for min dependencies except dask/pyarrow/numpy since that conflicts with the latest rapids version?

charlesbluca: Yup, this is essentially the 3.8 environment with the following modifications:

https://github.com/rapidsai/dask-build-environment/blob/c4aac1a7718413965d382b4e9f92f9d1028561af/dask_sql/Dockerfile#L35-L38

Merged with some RAPIDS requirements

Issue #1058, opened 2023-02-21
[BUG] CASE statements with columns starting with Number fail


mlahir1:
CASE statements with columns starting with Number fail,

Repro:
```
df = cudf.DataFrame()
df['a'] = range(10)
df['1b'] = range(10)
df = dask_cudf.from_cudf(df, npartitions=2)
c = Context()
c.create_table('table_pq', df)
x  = c.sql('SELECT (case when 1b=1 then 0 end) as x from table_pq')

File /conda/envs/dask-sql/lib/python3.10/site-packages/dask_sql/context.py:823, in Context._get_ral(self, sql)
    821     sqlTree = self.context.parse_sql(sql)
    822 except DFParsingException as pe:
--> 823     raise ParsingException(sql, str(pe))
    824 logger.debug(f"_get_ral -> sqlTree: {sqlTree}")
    826 rel = sqlTree

ParsingException: ParserError("Expected THEN, found: b")

```

PR #1060, opened 2023-02-22
Support column names starting with a number

sarahyurick:
Closes https://github.com/dask-contrib/dask-sql/issues/1058

While we cannot handle something like `SELECT 1b FROM df`, it should now work with double quotes `SELECT "1b" FROM df`. Example: https://stackoverflow.com/questions/6114193/column-name-beginning-with-a-number

Handling the former case would require changes in SQLParser and DataFusion.

### DIFF
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -188,16 +188,14 @@ impl PyExpr {
                     Err(_) => {
                         // Handles cases when from_qualified_name doesn't format the Column correctly.
                         // Here, we split the name string and grab the relation/table names
-                        let split_name: Vec<&str> = name.split(".").collect();
-                        let relation = &split_name.get(0);
+                        let split_name: Vec<&str> = name.split('.').collect();
+                        let relation = &split_name.first();
                         let table = &split_name.get(1);
                         let col = Column {
                             relation: Some(relation.unwrap().to_string()),
                             name: table.unwrap().to_string(),
                         };
-                        schema
-                            .index_of_column(&col)
-                            .map_err(py_runtime_err)
+                        schema.index_of_column(&col).map_err(py_runtime_err)
                     }
                 }
             }
--- a/dask_planner/src/expression.rs
+++ b/dask_planner/src/expression.rs
@@ -180,12 +180,9 @@ impl PyExpr {
                     schema.merge(plan.schema().as_ref());
                 }
                 let name = get_expr_name(&self.expr).map_err(py_runtime_err)?;
-                let index = schema
+                schema
                     .index_of_column(&Column::from_qualified_name(name.clone()))
-                    .map_err(py_runtime_err);
-                match index {
-                    Ok(i) => Ok(i),
-                    Err(_) => {
+                    .or_else(|_| {
                         // Handles cases when from_qualified_name doesn't format the Column correctly.
                         // Here, we split the name string and grab the relation/table names
                         let split_name: Vec<&str> = name.split('.').collect();
@@ -196,8 +193,7 @@ impl PyExpr {
                             name: table.unwrap().to_string(),
                         };
                         schema.index_of_column(&col).map_err(py_runtime_err)
-                    }
-                }
+                    })
             }
             _ => Err(py_runtime_err(
                 "We need a valid LogicalPlan instance to get the Expr's index in the schema",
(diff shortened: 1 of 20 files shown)


dask_sql/physical/rel/logical/project.py
```diff
@@ -46,8 +46,11 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
             # shortcut: if we have a column already, there is no need to re-assign it again
             # this is only the case if the expr is a RexInputRef
             if expr.getRexType() == RexType.Reference:
-                index = expr.getIndex()
-                backend_column_name = cc.get_backend_by_frontend_index(index)
+                try:
```
ayushdg, line 49: I haven't looked into this deeper but do we know why our `expr.getIndex()` call fails for cases like these? It seems like the schema does have `1b` but somehow we don't pick it up as expected.

sarahyurick: You're right, thanks! Looking further into our [getIndex()](https://github.com/dask-contrib/dask-sql/blob/f895346cb878e384410aa1ce2fd3bb3161d560e1/dask_planner/src/expression.rs#L173) function, it works as expected until we get to https://github.com/dask-contrib/dask-sql/blob/f895346cb878e384410aa1ce2fd3bb3161d560e1/dask_planner/src/expression.rs#L184

At first, I thought the issue was DataFusion's [index_of_column()](https://github.com/apache/arrow-datafusion/blob/f6e49ac7a027abb95d8b7fa755502dfa7d53c21c/datafusion/common/src/dfschema.rs#L238) function not behaving as expected. Actually, it's the [from_qualified_name()](https://github.com/apache/arrow-datafusion/blob/ec24724ea1130909402c64d5651a5815dcaf8285/datafusion/common/src/column.rs#L54) function, which we can create a workaround for on the Dask-SQL side. Just pushed new changes, let me know what you think.

ayushdg: Thanks for looking into this. I dug into this a bit deeper and it turns out that `Column::from_qualified_name(name)` calls sqlparser's tokenizer to tokenize the qualified name.
For cases like these `table_pq.a` it gets tokenized to 
```
[
  Word(
    Word { value: "table_pq", quote_style: None, keyword: NoKeyword }), 
    Period, 
    Word(Word { value: "a", quote_style: None, keyword: NoKeyword }
  )
]
```
while for cases like these `table_pq.1b` it's tokenized to 
```
[
  Word(
    Word { value: "table_pq", quote_style: None, keyword: NoKeyword }), 
    Number(".1", false), 
    Word(Word { value: "b", quote_style: None, keyword: NoKeyword }
  )
]
```

I'm not sure if it's a bug on sqlparser's end or if datafusion should separate the qualified name differently.

dask_planner/src/expression.rs
```diff
@@ -180,9 +180,24 @@ impl PyExpr {
                     schema.merge(plan.schema().as_ref());
                 }
                 let name = get_expr_name(&self.expr).map_err(py_runtime_err)?;
-                schema
-                    .index_of_column(&Column::from_qualified_name(name))
-                    .map_err(py_runtime_err)
+                let index = schema
+                    .index_of_column(&Column::from_qualified_name(name.clone()))
+                    .map_err(py_runtime_err);
```
ayushdg, line 185: Minor style preference. It might be possible for us to use `or_else` and instead map the error path to a function that implements the logic below.
Also we could split the logic below into a utility method in `parserUtils` that that returns the relation and table name, and use a closure here instead to map that result to a column and get the index

dask_sql/physical/rel/logical/project.py
sarahyurick: For now, I opened an issue https://github.com/apache/arrow-datafusion/issues/5426 about this on the DataFusion side.

dask_planner/src/expression.rs
sarahyurick: Thanks! Lmk if the most recent changes are what you had in mind.

ayushdg: Yeah this lgtm!

ayushdg (approved):
Changes generally lgtm! Will wait for Ci to be unblocked before re-running tests here.

PR #1062, opened 2023-02-23
Add server support for null values

charlesbluca:
Currently, `convert_cell` only has support for `np.nan`, which corresponds to `NaN` in query responses. This PR adds support for `pd.NA` (which crops up in columns with nullable or pyarrow dtypes), which should correspond to `NULL` in responses.

### DIFF
--- a/dask_sql/server/responses.py
+++ b/dask_sql/server/responses.py
@@ -2,6 +2,7 @@ import uuid
 
 import dask.dataframe as dd
 import numpy as np
+import pandas as pd
 from fastapi import Request
 
 from dask_sql.mappings import python_to_sql_type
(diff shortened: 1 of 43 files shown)


dask_sql/server/responses.py
```diff
@@ -87,7 +88,9 @@ def get_data_description(df):
     @staticmethod
     def convert_cell(cell):
         try:
-            if np.isnan(cell):  # pragma: no cover
+            if pd.isna(cell):
```
ayushdg, line 91: I'm not completely familiar with the server side of dask-sql but is `cell` always expected to be a `pandas` / host based object?

charlesbluca: Yeah, at this point in the response generation we would've called `to_pandas` on any cuDF frames

ayushdg (approved):

PR #1063, opened 2023-02-24
Add support for optional catalog name to `SHOW SCHEMAS | TABLES`

charlesbluca:
WIP to unblock some of the functionality of Metabase using its deprecated Presto driver with our server implementation.

TODO:

- [x] Add support for `SHOW TABLES FROM <catalog-name>.<schema-name>`

### DIFF
--- a/continuous_integration/gpuci/build.sh
+++ b/continuous_integration/gpuci/build.sh
@@ -52,9 +52,6 @@ python -m pip install git+https://github.com/dask/dask
 gpuci_logger "Install distributed"
 python -m pip install git+https://github.com/dask/distributed
 
-gpuci_logger "Install latest dask-cuda"
-gpuci_mamba_retry update -y -c rapidsai-nightly dask-cuda
-
 gpuci_logger "Install dask-sql"
 pip install -e ".[dev]"
 
--- a/dask_planner/src/parser.rs
+++ b/dask_planner/src/parser.rs
@@ -259,7 +259,9 @@ pub struct DescribeModel {
 /// Dask-SQL extension DDL for `SHOW SCHEMAS`
 #[derive(Debug, Clone, PartialEq, Eq)]
 pub struct ShowSchemas {
-    /// like
+    /// optional FROM identifier
+    pub from: Option<String>,
+    /// optional LIKE identifier
     pub like: Option<String>,
 }
 
@@ -1205,8 +1207,23 @@ impl<'a> DaskParser<'a> {
 
     /// Parse Dask-SQL SHOW SCHEMAS statement
     fn parse_show_schemas(&mut self) -> Result<DaskStatement, ParserError> {
-        // Check for existence of `LIKE` clause
-        let like_val = match self.parser.peek_token().token {
+        // parse optional `FROM` clause
+        let from = match self.parser.peek_token().token {
+            Token::Word(w) => {
+                match w.keyword {
+                    Keyword::FROM => {
+                        // move one token forward
+                        self.parser.next_token();
+                        // use custom parsing
+                        Some(self.parser.parse_identifier()?.value)
+                    }
+                    _ => None,
+                }
+            }
+            _ => None,
+        };
+        // parse optional `LIKE` clause
+        let like = match self.parser.peek_token().token {
             Token::Word(w) => {
                 match w.keyword {
                     Keyword::LIKE => {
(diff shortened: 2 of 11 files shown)


dask_sql/physical/rel/custom/show_schemas.py
```diff
@@ -16,23 +16,29 @@ class ShowSchemasPlugin(BaseRelPlugin):
...
         schemas.append("information_schema")
         df = pd.DataFrame({"Schema": schemas})
 
-        # We currently do not use the passed additional parameter FROM.
+        # currently catalogs other than the default `dask_sql` are not supported
+        catalog_name = show_schemas.getFrom() or context.catalog_name
+        if catalog_name != context.catalog_name:
+            raise RuntimeError(
+                f"A catalog with the name {catalog_name} is not present."
+            )
```
charlesbluca, line 39: For now, we are able to parse statements like `SHOW SCHEMAS FROM "dask_sql"`, but fail in any case where the catalog name isn't `dask_sql`; I figured it would make more sense to actually bother refactoring the Python handling of catalogs once we add functionality to add/remove/switch catalogs through queries.

dask_sql/server/app.py
```diff
@@ -74,6 +74,7 @@ async def query(request: Request):
     """
     try:
         sql = (await request.body()).decode().strip()
+        breakpoint()
```
charlesbluca, line 77:
```suggestion
```

jdye64:
@charlesbluca just wanted to share this here since we talked about this today. Seems like after this merges things should be much more straightforward here.

https://github.com/apache/arrow-datafusion/pull/5343#issuecomment-1466572908

dask_planner/src/parser.rs
```diff
@@ -1222,17 +1240,23 @@ impl<'a> DaskParser<'a> {
...
+            catalog_name,
+            like,
         })))
     }
 
     /// Parse Dask-SQL SHOW TABLES [FROM] statement
     fn parse_show_tables(&mut self) -> Result<DaskStatement, ParserError> {
+        let mut catalog_name = None;
         let mut schema_name = None;
         if !self.parser.consume_token(&Token::EOF) {
```
jdye64, line 1252: I know this was an existing code block that you only modified the inner behavior of but while we are at working on it I think it would benefit us to change this to an `if let` statement which would remove the requirement for the mutable `catalog_name` and `schema_name` variables.

dask_planner/src/sql/logical/show_schemas.rs
```diff
@@ -66,9 +68,14 @@ pub struct PyShowSchema {
...
 impl PyShowSchema {
+    #[pyo3(name = "getCatalogName")]
+    fn get_from(&self) -> PyResult<Option<String>> {
+        Ok(self.show_schema.catalog_name.clone())
+    }
+
     #[pyo3(name = "getLike")]
-    fn get_like(&self) -> PyResult<String> {
-        Ok(self.show_schema.like.as_ref().cloned().unwrap_or_default())
+    fn get_like(&self) -> PyResult<Option<String>> {
```
jdye64, line 77: I like this much better than the way it was.

jdye64 (changes requested):
@charlesbluca looks good. Had one small request

tests/integration/test_show.py
```diff
@@ -59,14 +59,23 @@ def test_wrong_input(c):
         c.sql(f'SHOW COLUMNS FROM "{c.schema_name}"."table"')
     with pytest.raises(AttributeError):
         c.sql('SHOW TABLES FROM "wrong"')
+    with pytest.raises(RuntimeError):
+        c.sql(f'SHOW TABLES FROM "wrong"."{c.schema_name}"')
```
ayushdg, line 63: Is it also possible to add a test where we're explicitly selecting the default catalog name and it works as expected?

ayushdg (approved):
Minor suggestion but not a blocker

continuous_integration/gpuci/build.sh
```diff
@@ -52,9 +52,6 @@ python -m pip install git+https://github.com/dask/dask
 gpuci_logger "Install distributed"
 python -m pip install git+https://github.com/dask/distributed
 
-gpuci_logger "Install latest dask-cuda"
-gpuci_mamba_retry update -y -c rapidsai-nightly dask-cuda
-
```
charlesbluca, line 57: Don't think this should be necessary anymore? Removing because the recent pinning of dask/distributed in dask-cuda nightlies is causing GPU failures due to the order of install here

Issue #1054, opened 2023-02-17
[ENH] Add support for Decimal types


randerzander:
A simple test of querying data which has decimal columns fails:
```
>>> import cudf
>>> df  = cudf.DataFrame({"id": [0, 1, 2]}
>>> ddf = dd.from_cudf(df, npartitions=1)
>>> ddf["id"] = ddf["id"].astype(Decimal64Dtype(10, 2))

>>> from dask_sql import Context
>>> c = Context()
>>> c.create_table("ddf", ddf)
>>> c.sql("select sum(id) from ddf").compute()
```
Trace:
```
Traceback (most recent call last):
  File "<input>", line 1, in <module>
    c.sql("select sum(id) from ddf").compute()
  File "/raid/rgelhausen/conda/envs/pynds3/lib/python3.10/site-packages/dask_sql/context.py", line 515, in sql
    rel, _ = self._get_ral(sql)
  File "/raid/rgelhausen/conda/envs/pynds3/lib/python3.10/site-packages/dask_sql/context.py", line 817, in _get_ral
    schemas = self._prepare_schemas()
  File "/raid/rgelhausen/conda/envs/pynds3/lib/python3.10/site-packages/dask_sql/context.py", line 777, in _prepare_schemas
    sql_data_type = python_to_sql_type(data_type)
  File "/raid/rgelhausen/conda/envs/pynds3/lib/python3.10/site-packages/dask_sql/mappings.py", line 113, in python_to_sql_type
    raise NotImplementedError(
NotImplementedError: The python type decimal64 is not implemented (yet)
```

PR #1073, opened 2023-03-07
Implement initial decimal support

ChrisJar:
Closes: #1054


dask_planner/src/sql/types.rs
```diff
@@ -111,6 +111,29 @@ impl DaskTypeMap {
...
+                        let scale: i8 = match dict.get_item("scale") {
+                            Some(e) => {
+                                let res: PyResult<i8> = e.extract();
+                                res.unwrap()
+                            }
+                            None => 0,
+                        };
+                        (precision, scale)
+                    }
+                    None => (38, 0),
```
jdye64, line 133: Assuming this is the default value of Decimal arrow data type? I'm really not sure

jdye64 (approved):
LGTM. Thanks @ChrisJar

dask_planner/src/sql/types.rs
ChrisJar: Yeah I was a little unsure about what the default there should be. 38 is just the maximum precision for the arrow Decimal128 type

dask_sql/mappings.py
```diff
@@ -107,6 +107,18 @@ def python_to_sql_type(python_type) -> "DaskTypeMap":
             tz=str(python_type.tz),
         )
 
+    try:
+        from cudf.api.types import is_decimal_dtype
+
+        if is_decimal_dtype(python_type):
```
charlesbluca, line 113: Wonder if there's a way we could implement this function internally in a similar fashion to `is_cudf_type` to avoid the cuDF import here?

dask_sql/mappings.py
```diff
@@ -280,6 +292,14 @@ def cast_column_to_type(col: dd.Series, expected_type: str):
     """Cast the given column to the expected type"""
     current_type = col.dtype
 
+    try:
+        from cudf.api.types import is_decimal_dtype
+
+        if is_decimal_dtype(current_type):
+            return col
```
charlesbluca, line 299: Do we run into errors when trying to cast a decimal column to a different type? If so and we need to avoid conversion, then we should probably change this to

```suggestion
            return None
```

To be consistent with the behavior below when conversion is skipped

dask_planner/src/sql/types.rs
```diff
@@ -111,6 +111,29 @@ impl DaskTypeMap {
                 };
                 DataType::Timestamp(unit, tz)
             }
+            SqlTypeName::DECIMAL => {
+                let (precision, scale) = match py_kwargs {
+                    Some(dict) => {
```
charlesbluca, line 116: More of a nit, is there any way to enforce the kwargs we receive here in the case that additional kwargs beyond `precision` and `scale` are provided?

charlesbluca (commented):
Thanks for the work here @ChrisJar 😄 a couple questions/comments:

charlesbluca:
After toying around a bit with this branch, I notice that the binary operation in `test_filter_decimal` triggers a cast on the datafusion end:

```
Projection: df.c
  TableScan: df projection=[a, b, c], full_filters=[df.a < CAST(df.b AS Decimal128(12, 3))]
```

This makes sense to me (the decimal columns in this case have different precision/scale), but I wonder if there's a way we can fully map cuDF's different sized types here so that the casting would be to `Decimal64`?

After making the changes I suggested above and having decimal columns casted to `expected_type`, I saw some failures because we haven't updated the Python type corresponding to `SqlTypeName.DECIMAL` here:

https://github.com/dask-contrib/dask-sql/blob/11fda749d4dc4c729350a1a1e87c0ea631f11bce/dask_sql/mappings.py#L71

Not sure if `DECIMAL` is used at all when we're only on CPU, but it might make sense to do an import check somewhere around here and replace `np.float64` with a cuDF decimal type so that decimal columns aren't accidentally cast to float (unfortunately think an import might be unavoidable in this case)

tests/integration/test_filter.py
```diff
@@ -253,3 +253,35 @@ def test_filtered_csv(tmpdir, c):
...
+    c.create_table("df", df)
+
+    result_df = c.sql(
+        """
+        SELECT
+            c
+        FROM
+            df
+        WHERE
+            a < b
```
charlesbluca, line 280: This is filtering by columns, but it might be worthwhile to either add a test or modify this one to filter by a scalar as well, to ensure that we have handling for creating a decimal scalar.

ChrisJar: Getting a cudf error when filtering by a scalar that I'm investigating

dask_sql/mappings.py
```diff
@@ -11,6 +11,20 @@
...
 
+
+def decimal_or_float():
+    """
+    Return a decimal type if available otherwise fallback to float
+    """
+    try:
+        from cudf import Decimal64Dtype
+
+        return Decimal64Dtype(1)
```
charlesbluca, line 22: cc @ayushdg @jdye64 do we have any preference for what the default decimal type should be here? This is essentially what would get used if we were to do something like

```python
from dask_sql import Context
from dask.datasets import timeseries

c = Context()
c.create_table("df", timeseries(), gpu=True)

c.sql("select cast(x as decimal) from df")
```

My first thought is that we should use `Decimal128Dtype` so that we can have the same maximum precision as the corresponding datafusion type; from there I'm not really sure what the precision/scale should be set to, though I think it should be changed from what it currently is since this results in floats getting rounded to 0.

Also interested in what our expectation is for casting here? Like if we attempt to cast a CPU column to decimal and have cuDF installed, should we create a cuDF dataframe to do this casting or just fail immediately? Currently we get an error deep in the pandas stack that isn't immediately clear

ChrisJar:
Also currently getting an error when filtering by a scalar on Decimal64Dtype or Decimal32Dtype column:
```python
from decimal import Decimal
import cudf
from dask_sql import Context


df = cudf.DataFrame({"d": [Decimal("1"), Decimal("2"), Decimal("3")]})
df["d"] = df["d"].astype(cudf.Decimal32Dtype(3,0))

c = Context()
c.create_table("df", df, gpu=True)

c.sql("select d from df where d < decimal '2'")
```
Throws:
```
ValueError('CUDF failure at:/opt/conda/conda-bld/work/cpp/src/binaryop/binaryop.cpp:206:
 Unsupported operator for these types')
```

ChrisJar:
^^^ This seems to be due to cudf not supporting operations between decimal columns of different sizes:
```
df.d < cudf.Scalar(2, dtype=cudf.Decimal32Dtype(3,0))
```
works but
```
df.d < cudf.Scalar(2, dtype=cudf.Decimal64Dtype(3,0))
```
throws the failure we see above

dask_sql/mappings.py
```diff
@@ -239,15 +252,20 @@ def similar_type(lhs: type, rhs: type) -> bool:
...
         is_td_ns,
         is_bool,
+        is_decimal,
     ]
 
     for check in checks:
         if check(lhs) and check(rhs):
+            # check that decimal columns have equal precision/scale
+            if check is is_decimal:
+                return lhs.precision == rhs.precision and lhs.scale == rhs.scale
```
charlesbluca, line 268: Depending on what type of upcast/downcasting is possible or desired, we may be able to loosen this check rather than doing an exact check on precision/scale

dask_sql/physical/rel/base.py
```diff
@@ -104,7 +104,13 @@ def fix_dtype_to_row_type(dc: DataContainer, row_type: "RelDataType"):
...
 
         for field_name, field_type in field_types.items():
-            expected_type = sql_to_python_type(field_type.getSqlType())
+            sql_type = field_type.getSqlType()
+            sql_type_args = tuple()
+
+            if str(sql_type) == "SqlTypeName.DECIMAL":
+                sql_type_args = field_type.getDataType().getPrecisionScale()
+
+            expected_type = sql_to_python_type(sql_type, *sql_type_args)
```
charlesbluca, line 113: Does this pattern seem okay for optionally deciding if we need to push creation args to the Python type we're creating? Would like to make sure there's not a better way to do this before applying it in the other places where we call `sql_to_python_type`

charlesbluca (commented):
The last commit I pushed should resolve the failing queries - main thing remaining are:

- applying whatever pattern we agree upon for creating decimal python types to wherever else we call `sql_to_python_type`
- adding a configuration variable to control whether we use CPU/GPU decimal handling

dask_sql/mappings.py
```diff
@@ -68,7 +75,8 @@
 _SQL_TO_PYTHON_FRAMES = {
     "SqlTypeName.DOUBLE": np.float64,
     "SqlTypeName.FLOAT": np.float32,
-    "SqlTypeName.DECIMAL": np.float64,  # We use np.float64 always, even though we might be able to use a smaller type
+    # a column of Decimals in panda is `object`, but cuDF has a dedicated dtype
+    "SqlTypeName.DECIMAL": object if not cudf else cudf.Decimal128Dtype(38, 10),
```
ayushdg, line 79: What happens in cases where `cudf` is importable but users want to still run on the CPU?

dask_sql/physical/rel/base.py
ayushdg: Yeah, this seems reasonable especially since Decimal right now is the only type we have where there is additional information attached to the type.

ayushdg (commented):
Minor questions/suggestions. But generally should be good to go

dask_sql/mappings.py
charlesbluca: Handling still needs to be added there, as in that case we would end up failing on CPU; think the best choice for now is adding a Dask config variable but once #1117 is further along it would probably make sense to base this on whether or not we're working with a GPU/CPU context

ayushdg: In the interest of this feature making it in, I'm okay with merging this in as is and following up with a config option or #1117 before the next release.

ayushdg (approved):

PR #1074, opened 2023-03-07
Add `filepath` fields in Python and Rust

sarahyurick:
As part of the dynamic partition pruning optimizer rule, we'll be reading in Parquet files on the Rust side. To do this, we need to obtain information about the file paths, when available, to pass into Rust. Here, I use similar logic as the statistics logic from https://github.com/dask-contrib/dask-sql/pull/1061 and https://github.com/dask-contrib/dask-sql/pull/1027.

### DIFF
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -133,22 +133,15 @@ impl ContextProvider for DaskSQLContext {
                 // If the Table is not found return None. DataFusion will handle the error propagation
                 match resp {
                     Some(e) => {
-                        let statistics = &self
+                        let table_ref = &self
                             .schemas
                             .get(reference.schema.as_ref())
                             .unwrap()
                             .tables
                             .get(reference.table.as_ref())
-                            .unwrap()
-                            .statistics;
-                        let filepath = &self
-                            .schemas
-                            .get(reference.schema.as_ref())
-                            .unwrap()
-                            .tables
-                            .get(reference.table.as_ref())
-                            .unwrap()
-                            .filepath;
+                            .unwrap();
+                        let statistics = &table_ref.statistics;
+                        let filepath = &table_ref.filepath;
                         if statistics.get_row_count() == 0.0 {
                             Ok(Arc::new(table::DaskTableSource::new(
                                 Arc::new(e),
--- a/dask_planner/src/sql.rs
+++ b/dask_planner/src/sql.rs
@@ -150,7 +150,10 @@ impl ContextProvider for DaskSQLContext {
                             .unwrap()
                             .filepath;
                         if statistics.get_row_count() == 0.0 {
-                            Ok(Arc::new(table::DaskTableSource::new(Arc::new(e), None, filepath.clone())))
+                            Ok(Arc::new(table::DaskTableSource::new(
+                                Arc::new(e),
+                                None, filepath.clone(),
+                            )))
                         } else {
                             Ok(Arc::new(table::DaskTableSource::new(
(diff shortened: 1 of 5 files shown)


dask_sql/context.py
```diff
@@ -250,6 +250,13 @@ def create_table(
             **kwargs,
         )
 
+        if type(input_table) == str:
+            filepath = input_table
+            dc.filepath = input_table
+            self.schema[schema_name].filepaths[table_name.lower()] = input_table
+        else:
+            filepath = ""
```
sarahyurick, line 258: Using empty string here as `None` throws an error.

dask_sql/context.py
```diff
@@ -250,6 +250,13 @@ def create_table(
             **kwargs,
         )
 
+        if type(input_table) == str:
+            filepath = input_table
+            dc.filepath = input_table
```
jdye64, line 255: I like this. I feel like this should have always been an attribute in the DataContainer all along

dask_sql/context.py
jdye64: I'm fine with `""` but wonder why `None` was causing issues? hmm

jdye64: What the error on the Rust side or Python side?

jdye64 (approved):
Changes look good. Little concerned about using `""` instead of `None` for the "empty" filepath but only from the Rust side as I worry someone might be manually checking for `Result::None` in Rust but `Result::Some("")` would be there but the intention would be meant for the filepath to in fact by "not present" Thanks @sarahyurick

dask_sql/context.py
jdye64: Ahh, yep. That is worth doing I think. Thanks for pointing that out Charles.

sarahyurick: Thanks @charlesbluca for looking into this!

sarahyurick:
Thanks @jdye64 !

charlesbluca (approved):
Thanks @sarahyurick!

dask_planner/src/sql.rs
```diff
@@ -141,12 +141,25 @@ impl ContextProvider for DaskSQLContext {
...
                             .unwrap()
                             .statistics;
+                        let filepath = &self
+                            .schemas
+                            .get(reference.schema.as_ref())
+                            .unwrap()
+                            .tables
+                            .get(reference.table.as_ref())
+                            .unwrap()
+                            .filepath;
```
ayushdg, line 151: Minor nit: Can re-use the daskTable reference we get after `unwrap` and re-use it for both statistics and filepath.

tests/unit/test_context.py
```diff
@@ -329,3 +330,29 @@ def test_alter_table(c, df_simple):
...
+            ]
+            * 5,
+            "index": range(15),
+        },
+    )
+    dd.from_pandas(df, npartitions=3).to_parquet(parquet_path)
+
+    c.create_table("df", parquet_path, format="parquet")
+
+    assert c.schema["root"].tables["df"].filepath == parquet_path
```
ayushdg, line 357: Minor suggestion: Could we also add a test for a case where the filepath isn't available and the values here would be `None`

ayushdg (changes requested):
Minor suggestions. Overall lgtm, thanks!

sarahyurick:
Thanks @ayushdg @charlesbluca ! Added the suggested changes, lmk what you think.

ayushdg (approved):

PR #1080, opened 2023-03-13
Pin `sqlalchemy<2` in CI environments

charlesbluca:
Looks like our test code is incompatible with API changes made in SQLAlchemy v2; this PR pins CI environments to `sqlalchemy<2` to unblock CI. Since there aren't any immediate reasons to bump to using SQLAlchemy v2, I'm okay with maintaining this pinning versus updating the test code.

Closes https://github.com/dask-contrib/dask-sql/issues/1078


ayushdg (commented):
Question: Since the sqlalchemy depends comes in from fugue and other libraries, does it makes sense to add the dependency in any file that has fugue as a dependency (including setup.py etc)?

Issue #1084, opened 2023-03-14
[ENH] Add Arrow DataFusion Python as Cargo dependency


jdye64:
**Is your feature request related to a problem? Please describe.**
Add Arrow DataFusion Python as a Cargo dependency in Dask-SQL. This means removing the existing DataFusion dependencies so there will be a decent amount of code refactoring although it will all be trivial use renaming for the most part.

**Describe the solution you'd like**
Remove all existing DataFusion dependencies from the Cargo file and add Arrow DataFusion Python and make the required changes to get everything working. Try to change the least amount of files as possible although the very nature of the refactor will call for quite a few files to be changed.

**Describe alternatives you've considered**
None

**Additional context**
None

PR #1085, opened 2023-03-14
Add arrow datafusion python dependency

jdye64:
Closed with #1084


jdye64:
Waiting on https://github.com/oxidecomputer/typify/pull/221 to merge so that conda builds can work with new setup.

.github/actions/setup-builder/action.yaml
```diff
@@ -0,0 +1,38 @@
...
+    required: true
+    default: 'stable'
+runs:
+  using: "composite"
+  steps:
+    - name: Install Build Dependencies
+      shell: bash
+      run: |
+        apt-get update
+        apt-get install -y protobuf-compiler
```
charlesbluca, line 15: Looks like we need to `sudo` these installs:

```suggestion
        sudo apt-get update
        sudo apt-get install -y protobuf-compiler
```

https://stackoverflow.com/questions/57982945/how-to-apt-get-install-in-a-github-actions-workflow

dask_sql/mappings.py
```diff
@@ -177,6 +178,12 @@ def sql_to_python_value(sql_type: "SqlTypeName", literal_value: Any) -> Any:
         # Calcite will always convert INTERVAL types except YEAR, QUATER, MONTH to milliseconds
         # Issue: if sql_type is INTERVAL MICROSECOND, and value <= 1000, literal_value will be rounded to 0
         return np.timedelta64(literal_value, "ms")
+    elif sql_type == SqlTypeName.INTERVAL_MONTH_DAY_NANOSECOND:
+        # DataFusion assumes 30 days per month. Therefore we multiply number of months by 30 and add to days
+        return pd.Timedelta(
+            np.timedelta64((literal_value[0] * 30) + literal_value[1], "D")
+            + np.timedelta64(literal_value[2], "ns")
+        )
```
ayushdg, line 186:
```suggestion
        return np.timedelta64(
            (literal_value[0] * 30) + literal_value[1], "D"
        ) + np.timedelta64(literal_value[2], "ns")
```

.github/workflows/conda.yml
```diff
@@ -49,6 +49,11 @@ jobs:
           which python
           pip list
           mamba list
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
+        with:
+          version: '3.x'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
charlesbluca, line 56: We should probably remove this and instead add `libprotobuf=3` to the nightly recipe's build requirements (in a similar fashion to [what's being done in ADP](https://github.com/apache/arrow-datafusion-python/blob/af1a8ebade3d537ce24f6f01a037a56177a917fa/conda/recipes/meta.yaml#L40)) so that compilers are handled entirely through conda-forge once this gets pushed into a stable release

```suggestion
```

.github/workflows/rust.yml
```diff
@@ -30,6 +30,11 @@ jobs:
         id: detect-trigger
         with:
           keyword: "[test-df-upstream]"
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
+        with:
+          version: '3.x'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
charlesbluca, line 37: Don't think this should be necessary since this job isn't doing any compilation

```suggestion
```

.github/workflows/test.yml
```diff
@@ -60,6 +60,11 @@ jobs:
         with:
           workspaces: dask_planner
           shared-key: test
+      - name: Install Protoc
+        uses: arduino/setup-protoc@v1
+        with:
+          version: '3.x'
+          repo-token: ${{ secrets.GITHUB_TOKEN }}
```
charlesbluca, line 67: Would remove this and instead add `libprotobuf=3` to the conda dev environments so that users don't need to install it manually

```suggestion
```

.github/workflows/test.yml
```diff
@@ -106,6 +111,11 @@ jobs:
...
```
charlesbluca, line 118: Same as above

```suggestion
```

dask_sql/mappings.py
```diff
@@ -177,6 +178,11 @@ def sql_to_python_value(sql_type: "SqlTypeName", literal_value: Any) -> Any:
         # Calcite will always convert INTERVAL types except YEAR, QUATER, MONTH to milliseconds
         # Issue: if sql_type is INTERVAL MICROSECOND, and value <= 1000, literal_value will be rounded to 0
         return np.timedelta64(literal_value, "ms")
+    elif sql_type == SqlTypeName.INTERVAL_MONTH_DAY_NANOSECOND:
+        # DataFusion assumes 30 days per month. Therefore we multiply number of months by 30 and add to days
+        return np.timedelta64(
+            (literal_value[0] * 30) + literal_value[1], "D"
+        ) + np.timedelta64(literal_value[2], "ns")
```
charlesbluca, line 185: This fix makes sense to me - two things I'm interested in:

- is there any way to hit the cases for `INTERVAL` and `INTERVAL_DAY` directly above this? Wasn't able to reach them with our standard interval tests, interested in what CI would look like if we remove them (and maybe also associated plumbing) altogether
- with this change, we might want to consider modifying @sarahyurick's handling for months/quarters in `DatetimeSubOperation` to make the same assumption around a uniform 30 day month for sake of consistency

dask_sql/physical/rex/core/literal.py
```diff
@@ -161,6 +161,9 @@ def convert(
         elif literal_type == "IntervalDayTime":
             literal_type = SqlTypeName.INTERVAL_DAY
             literal_value = rex.getIntervalDayTimeValue()
+        elif literal_type == "IntervalMonthDayNano":
+            literal_type = SqlTypeName.INTERVAL_MONTH_DAY_NANOSECOND
+            literal_value = rex.getIntervalMonthDayNanoValue()
```
charlesbluca, line 166: Same question around the `IntervalDayTime` case being reached as above

tests/integration/test_groupby.py
```diff
@@ -217,6 +217,7 @@ def test_aggregations(c):
     GROUP BY user_id
     """
     )
+
```
charlesbluca, line 220: Just removing seemingly unnecessary diff

```suggestion
```

tests/integration/test_select.py
```diff
@@ -95,6 +95,8 @@ def test_select_of_select_with_casing(c, df):
         """
     )
 
+    print(f"Result DF:\n{result_df.head()}")
+
```
charlesbluca, line 99:
```suggestion
```

charlesbluca (commented):
Thanks @jdye64 🎉 a few comments around how we include `protoc` in CI/packaging and some follow up around the timedelta changes

dask_planner/src/sql/optimizer.rs
```diff
@@ -45,8 +38,9 @@ impl DaskSqlOptimizer {
     /// optimizers as well as any custom `OptimizerRule` trait impls that might be desired.
     pub fn new() -> Self {
         debug!("Creating new instance of DaskSqlOptimizer");
+
         let rules: Vec<Arc<dyn OptimizerRule + Sync + Send>> = vec![
-            Arc::new(TypeCoercion::new()),
+            // Arc::new(TypeCoercion::new()),
```
ayushdg, line 43: Was there a specific issue with this rule enabled?

ayushdg (changes requested):

dask_planner/src/sql/logical.rs
```diff
@@ -325,7 +327,8 @@ impl PyLogicalPlan {
             LogicalPlan::CreateCatalogSchema(_create) => "CreateCatalogSchema",
             LogicalPlan::CreateCatalog(_create_catalog) => "CreateCatalog",
             LogicalPlan::CreateView(_create_view) => "CreateView",
-            LogicalPlan::SetVariable(_) => "SetVariable",
+            LogicalPlan::Statement(_) => "Statement",
+            // LogicalPlan::SetVariable(_) => "SetVariable",
```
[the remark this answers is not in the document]
jdye64: `LogicalPlan::SetVariable` was removed in the latest version of DataFusion 22. I will just remove it instead of commenting it out.

dask_planner/src/sql/optimizer.rs
jdye64: TypeCorercion is no longer a OptimizerRule. Instead it was moved to be an Analyzer. Will just remove the commented out line entirely

dask_planner/.cargo/config.toml
```diff
@@ -0,0 +1,11 @@
+[target.x86_64-apple-darwin]
```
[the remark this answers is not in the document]
jdye64: This is a fix for macOS builds. Specifically around substrait which ADP includes so therefore why it was added to this PR.

.github/workflows/rust.yml
jdye64: Actually this one needs to stay. Cargo Check runs a build command under the covers against ADP which in turn requires protoc to build out substrait proto files.

dask_sql/mappings.py
jdye64: It is possible for `INTERVAL` or `INTERVAL_DAY` cases to be hit. They are still variants of DataFusion. I think we should leave those as are.

Agree on @sarahyurick logic though. I will make the change to make it assume 30 days for consistency sake.

dask_sql/physical/rex/core/literal.py
jdye64: It is possible so I think we leave it. Our tests just don't hit it.

tests/integration/test_rex.py
```diff
@@ -926,8 +926,8 @@ def test_timestampdiff(c):
             "hours": [46, 6648, 591744],
             "days": [1, 277, 24656],
             "weeks": [0, 39, 3522],
-            "months": [0, 9, 810],
-            "quarters": [0, 3, 270],
+            "months": [0, 9, 821],
```
ayushdg, line 929: Since this behavior is different from the results we get from mysql and a few implementations that support timestampdiff, I'm not sure whether we want to keep the existing logic or modify it. I didn't find an easy way to get a timestampdiff with datafusion/postgresql

ayushdg (approved):
Thanks a lot for these changes @jdye64!

pyproject.toml
```diff
@@ -0,0 +1,13 @@
...
+[tool.isort]
+profile = "black"
+
+[tool.maturin]
+include = [
+    { path = "Cargo.lock", format = "sdist" }
+]
+exclude = [".github/**", "ci/**", ".asf.yaml"]
+# Require Cargo.lock is up to date
+locked = true
```
charlesbluca, line 13: Noting that I've been thinking about consolidating as much of our project metadata into pyproject.toml as possible (similar to other RAPIDS/Dask projects) which would make this file less stubby

charlesbluca (approved):

pyproject.toml
ayushdg: +1 to this. I've also been thinking of simplifying the setup.py scripts and moving as much as possible into the pyproject.toml file

PR #1089, opened 2023-03-16
Add columns data type creation arguments to server response for `CHAR`/`VARCHAR` columns

charlesbluca:
Some SQL clients (such as DBeaver and Looker) expect servers to send along the creation arguments for a column's data type; for example, `VARCHAR(10)` would map to:

```json
"typeSignature": {
    "rawType": "VARCHAR",
    "arguments": [{"type": "LONG", "value": 10}],
},
```

This PR adds these arguments to the response for `CHAR` and `VARCHAR` columns, which are 2 of the 3 types where this would be an issue (the third being `DECIMAL`, which is in progress in https://github.com/dask-contrib/dask-sql/pull/1073).

### DIFF
--- a/dask_sql/context.py
+++ b/dask_sql/context.py
@@ -9,7 +9,6 @@ import dask.dataframe as dd
 import pandas as pd
 from dask import config as dask_config
 from dask.base import optimize
-from dask.distributed import Client
 
 from dask_planner.rust import (
     DaskSchema,
@@ -690,14 +689,7 @@ class Context:
             self, auto_include=auto_include, disable_highlighting=disable_highlighting
         )
 
-    def run_server(
-        self,
-        client: Client = None,
-        host: str = "0.0.0.0",
-        port: int = 8080,
-        log_level=None,
-        blocking: bool = True,
-    ):  # pragma: no cover
+    def run_server(self, **kwargs):  # pragma: no cover
         """
         Run a HTTP server for answering SQL queries using ``dask-sql``.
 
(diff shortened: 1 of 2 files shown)


dask_sql/server/responses.py
```diff
@@ -65,13 +65,18 @@ def __init__(self, request: Request, next_url: str = None, cancel_url: str = Non
...
             {
                 "name": column_name,
-                "type": sql_type.lower(),
-                "typeSignature": {"rawType": sql_type.lower(), "arguments": []},
+                "type": sql_type,
+                "typeSignature": {
+                    "rawType": sql_type,
+                    "arguments": []
+                    if sql_type not in ("char", "varchar")
+                    else [{"type": "LONG", "value": 10}],
```
charlesbluca, line 78: Hardcoded this for now as at least on DBeaver it doesn't seem to check the column data against this value; in reality we would ideally want to expose this information through the SQL type and use that instead

dask_sql/context.py
```diff
@@ -690,14 +689,7 @@ def ipython_magic(
...
 
-    def run_server(
-        self,
-        client: Client = None,
-        host: str = "0.0.0.0",
-        port: int = 8080,
-        log_level=None,
-        blocking: bool = True,
-    ):  # pragma: no cover
+    def run_server(self, **kwargs):  # pragma: no cover
```
charlesbluca, line 692: Made this change because currently it isn't possible to do something like

```python
c = Context()
c.create_table("df", timeseries())
c.run_server(jdbc_metadata=True)
```

Think it might also make sense to use the same docstring for `Context.run_server` and `app.run_server` so that they're kept in sync (currently `Context.run_server` is outdated)

jdye64 (approved):
Good find @charlesbluca LGTM

ayushdg (commented):
Changes lgtm! 
Quick question: If treating `char/varchar` sql_types as `long(10)` is something specific to a few  a sql clients, going forward we might want to think of customizing the responses based on some kwargs/info about the connecting client

ayushdg (approved):

PR #1095, opened 2023-03-22
Run upstream testing with Dask pickle serialization branches

charlesbluca:
cc @rjzamora


charlesbluca:
Looks like the relevant PRs are merged in and reflected in the Dask nightlies, so this should now be getting captured in upstream testing

rjzamora:
Thanks for this PR @charlesbluca ! It was very useful to know that the "pickle" changes were compatible with dask-sql

PR #1094, opened 2023-03-20
Bump datafusion -> 21.0.0 and add dyn and dyn_hash functions to custo…

jdye64:
…m parsers

PR #1102, opened 2023-03-28
Dynamic partition pruning

sarahyurick:
WIP

Update: Should be ready to go, in combination with https://github.com/dask-contrib/dask-sql/pull/1160 !


dask_planner/src/sql/optimizer/dynamic_partition_pruning.rs
```diff
@@ -0,0 +1,935 @@
...
+    TableScan,
+};
+use datafusion_optimizer::{OptimizerConfig, OptimizerRule};
+use log::warn;
+
+use crate::sql::table::DaskTableSource;
+
+// Optimizer rule for dynamic partition pruning
+// General TODOs:
+// - Depends on DataFusion 21.0.0
```
sarahyurick, line 38: Currently running with https://github.com/dask-contrib/dask-sql/pull/1094

sarahyurick:
Waiting on https://github.com/rapidsai/ops/issues/2644 to resolve some datatype issues for the gpuCI tests.

dask_planner/src/sql/optimizer/dynamic_partition_pruning.rs
```diff
@@ -0,0 +1,968 @@
...
+// Optimizer rule for dynamic partition pruning
+pub struct DynamicPartitionPruning {}
+
+impl DynamicPartitionPruning {
+    #[allow(missing_docs)]
+    #[allow(dead_code)]
+    pub fn new() -> Self {
+        Self {}
+    }
+}
```
jdye64, line 47: Since this isn't used and the `DynamicPartitionPruning` struct has no members can we just remove this?

dask_planner/src/sql/optimizer/dynamic_partition_pruning.rs
```diff
@@ -0,0 +1,968 @@
...
+    ) -> Result<Option<LogicalPlan>> {
+        // Parse the LogicalPlan and store tables and columns being (inner) joined upon. We do this
+        // by creating a HashSet of all InnerJoins' join.on and join.filters
+        let join_conds = gather_joins(plan);
+        let tables = gather_tables(plan);
+        let aliases = gather_aliases(plan);
+
+        if join_conds.is_empty() || tables.is_empty() {
+            // No InnerJoins to optimize with
+            Ok(None)
```
jdye64, line 67: I'm probably looking at this wrong but wouldn't this cause the `optimized_plan` variable in `optimizer.rs` to end up as `None`? Maybe the syntax defaults back to the output from `.optimize(existing_plan.original_plan)`?? I honestly don't know? More of a question than a statement.

dask_planner/src/sql/optimizer/dynamic_partition_pruning.rs
```diff
@@ -0,0 +1,968 @@
...
+
+            // Iterate through all inner joins in the query
+            for join_cond in &join_conds {
+                let join_on = &join_cond.on;
+                for on_i in join_on {
+                    // Obtain tables and columns (fields) involved in join
+                    let left_on = &on_i.0;
+                    let right_on = &on_i.1;
+                    let mut left_table: Option<String> = None;
+                    let mut left_field: Option<String> = None;
```
jdye64, line 91: Since this is used as a "pair" might make sense to create a tuple like `(Option<String>, Option<String>)` to keep the variables stored "together"

jdye64 (changes requested):

dask_planner/src/sql/optimizer/dynamic_partition_pruning.rs
sarahyurick: From my understanding, returning `Ok(None)` is the same as returning `Ok(plan.clone())`. At least that's what [this comment](https://github.com/apache/arrow-datafusion/blob/beee1d91303d5eff220fadd08b2c28404c2b3e5a/datafusion/optimizer/src/optimizer.rs#L58) indicates, I haven't actually dug too deep into the optimizer code on the DataFusion side.

dask_planner/src/sql/optimizer/dynamic_partition_pruning.rs
sarahyurick: Yeah, I like this idea, thanks!

jdye64 (approved):
Thanks for the changes @sarahyurick looks good to me

dask_sql/context.py
```diff
@@ -102,6 +102,9 @@ def __init__(self, logging_level=logging.INFO):
         self.context = DaskSQLContext(self.catalog_name, self.schema_name)
         self.context.register_schema(self.schema_name, DaskSchema(self.schema_name))
 
+        if dask_config.get("sql.dynamic_partition_pruning"):
+            self.context.apply_dynamic_partition_pruning()
+
```
ayushdg, line 107: Sometime configs can be passed through the `c.sql` method as well https://github.com/dask-contrib/dask-sql/blob/main/dask_sql/context.py#L468. 

Unless you need this during context initialization, it might make sense to set this variable somewhere in `_get_ral` before the query is parsed.

ayushdg (changes requested):
Changes generally lgtm! Thanks a lot for this. 

Just some a couple of minor suggestions.

Would it be possible to followup this pr with some tests both for the config option as well as testing dpp.

dask_sql/context.py
sarahyurick: Thanks! Done with https://github.com/dask-contrib/dask-sql/pull/1102/commits/182fe8ed938ac71e6fefe99e8ebe5f59188a08e9 - with this, DPP can be configured either globally (default True) or on a per-query basis with the `config_options` parameter.

ayushdg (approved):

PR #1112, opened 2023-04-05
Pin dask/distributed for release

charlesbluca:
Sets a max version constraint for Dask/Distributed, and separates the `dask[dataframe,distributed]` constraint set in our setup.py so that we can allow for the Distributed patch release


setup.py
```diff
@@ -42,7 +42,8 @@
     python_requires=">=3.8",
     setup_requires=sphinx_requirements,
     install_requires=[
-        "dask[dataframe,distributed]>=2022.3.0",
+        "dask>=2022.3.0,<=2023.3.2",
+        "distributed>=2022.3.0,<2023.3.3.0a0",
```
charlesbluca, line 46: Roughly copying this from the conda pinnings in `dask`'s feedstock recipe

ayushdg (approved):

PR #1116, opened 2023-04-10
Avoid using timezone-aware datetime data in GPU tables, unpin `uvicorn` in 3.8 GPU environments

charlesbluca:
It looks like https://github.com/rapidsai/cudf/pull/13086 has made it so cuDF will now error when attempting to construct a dataframe containing timezone-aware datetime data.

This PR attempts to resolve the resulting gpuCI failures by casting any affected timezone-aware datetime columns to non-aware datetime - since cuDF was seemingly ignoring this information silently before now, don't think this should have an impact on relevant GPU tests

### DIFF
--- a/continuous_integration/gpuci/environment-3.8.yaml
+++ b/continuous_integration/gpuci/environment-3.8.yaml
@@ -37,7 +37,7 @@ dependencies:
 - sqlalchemy<2
 - tpot
 - tzlocal=2.1
-- uvicorn=0.13.4
+- uvicorn>=0.13.4
 # GPU-specific requirements
 - cudatoolkit=11.5
 - cudf=23.06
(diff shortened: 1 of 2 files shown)


continuous_integration/gpuci/environment-3.8.yaml
```diff
@@ -37,7 +37,7 @@ dependencies:
 - sqlalchemy<2
 - tpot
 - tzlocal=2.1
-- uvicorn=0.13.4
+- uvicorn>=0.13.4
```
charlesbluca, line 40: Mirrors a change made to the images in https://github.com/rapidsai/dask-build-environment/pull/64 that we need to make in order to pull in the latest `cuml` nightlies (the lack of which is causing the current GPU failures)

charlesbluca:
rerun tests

ayushdg (approved):

Issue #1111, opened 2023-04-05
[ENH] Making Fugue[dask] depend on Dask-sql


goodwanghan:
**Is your feature request related to a problem? Please describe.**
Currently dask-sql has integration with Fugue. The problem is some Fugue change will break this integration. Also Fugue-dask by default relies on the qpd project to support SQL, which I want to totally deprecate.

**Describe the solution you'd like**
Remove the dependency of fugue from this repo
Add the dependency of dask-sql on Fugue side
https://github.com/fugue-project/fugue/issues/454

**Can you please wait a little bit for the related code change (I will make a PR to remove Fugue) before your next release?**

ayushdg:
> Can you please wait a little bit for the related code change (I will make a PR to remove Fugue) before your next release?

We were planning on cutting a release EOD today/early tomm, to bump our dask version dependency to be compatible with the [upcoming](https://docs.rapids.ai/maintainers) Rapids release.
That being said we're happy to do another patch release in the coming week or two and we can add the relevant code changes in then.

PR #1120, opened 2023-04-17
Refactor independent cluster configs, pin `pandas<2` in cluster environments

charlesbluca:
Looks like the latest Dask release has made it so the images now pull in pandas 2.0, which ends up being inconsistent with the client environment for cluster testing (fugue pulls in a `pandas<2` constraint).

This applies that constraint to the cluster environments, which should resolve the resulting failures that began popping up; I've also made some modifications to the cluster configuring Docker Compose files so that:

- both stable and upstream cluster testing rely on the same shared environment file
- the cluster environment is kept consistent between scheduler and workers

Hoping that having this shared environment file should cut down on repetition and make it easier to document modifications we need to make to the base cluster environment to get things working


jdye64 (approved):
LGTM

ayushdg (approved):
Thanks for looking into this! 
Once fugue updates the pandas pinning or #1111 lands we should be able to remove the pinning and have pandas 2.0 in both places (hopefully).

Curious if in addition to the refactor here it might to useful to have some utility that checks that the environment/versions of packages on the client, scheduler and workers match before starting the tests. Maybe even upstream dask/distributed could benefit from something like this (if it's easy to do)

charlesbluca:
> Maybe even upstream dask/distributed could benefit from something like this

From what I recall of some previous runs, this should be something that currently exists in upstream Dask, as I recall there being warnings thrown if mismatched versions of some core packages were detected between the client/scheduler/workers - might be just seeing if there's appetite to configure this to raise an error instead of a warning

PR #1123, opened 2023-04-24
Remove independent cluster testing

charlesbluca:
After some internal conversation prompted by changes in Dask/Distributed that generally encourage having a consistent client/scheduler/worker environment (relevant [Dask blogpost](https://blog.dask.org/2023/04/14/scheduler-environment-requirements)), the general consensus is that it doesn't seem worthwhile to support/test the specific cluster setup required to run queries on a cluster without dask-sql.

This PR removes `make_pickable_without_dask_sql` from the codebase, as well as all the independent cluster testing infrastructure; it doesn't seem like there's much advertising of dask-sql as a client-only library in the documentation though this line from the docs might imply something:

> **Easy to install and maintain:** dask-sql is just a pip/conda install away (or a docker run if you prefer). No need for complicated cluster setups - dask-sql will run out of the box on your machine and can be easily connected to your computing cluster.

cc @randerzander @ayushdg


ayushdg:
rerun tests

.github/workflows/test.yml
```diff
@@ -33,16 +33,23 @@ jobs:
...
     strategy:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
         python: ["3.8", "3.9", "3.10"]
+        distributed: [false]
+        include:
+          # run tests on a distributed client
+          - os: "ubuntu-latest"
+            python: "3.8"
```
ayushdg, line 51: Do you have a preference on whether the distributed tests should run with the mindeps conda version vs new deps?

ayushdg (changes requested):
A few places in the readme/docs don't explicitly say it but seem to imply that only the client needs dask-sql like [here](https://github.com/dask-contrib/dask-sql/blob/main/README.md?plain=1#L23). It might make sense to update the wording there a bit.

Small question but on the whole looks good to me!

.github/workflows/test.yml
charlesbluca: I opted for mindeps here to make sure that our min versions work in a distributed context, but in reality feel like it would probably be good to test both old and new deps - do you have any opposition to adding another build here for 3.9/3.10?

ayushdg: nope that works for me

ayushdg (approved):
Thanks a lot for this. Should improve stability of our tests and simplify the workflow significantly.

PR #1128, opened 2023-04-27
Remove `python-xdist` workaround from CI environments

charlesbluca:
Think we should be good to remove `py` from CI environments now that https://github.com/pytest-dev/pytest-xdist/issues/832 is closed


ayushdg (approved):
Thanks for this. Completely forgot to followup on it.

Issue #1121, opened 2023-04-21
[ENH] Dynamic partition pruning improvements


sarahyurick:
https://github.com/dask-contrib/dask-sql/pull/1102 adds dynamic partition pruning functionality. While working on this, I noticed several features that could be used to enhance this optimization rule that are outside of the original intended scope of the project. I think DPP could benefit by expanding to include these cases in the future.

- [ ] Currently, we only check a join's `on` conditions, but we should also try to check and make use of join `filters`
- [ ] Right now, we only use DPP for joins between 2 columns. However, it would also be possible to run DPP for joins between binary expressions, e.g. `WHERE col1 + 10 = col2 + 20`
- [ ] In a similar vein, we should expand the `get_filtered_fields` function to be able to handle more complex binary expressions than it currently does
- [ ] Allow the `fact_dimension_ratio` and possibly other parameters to be specified by the user

PR #1130, opened 2023-05-02
TableScan filters to PyArrow DNF format

jdye64:
This PR adds the support to get the filters for a `LogicalPlan::TableScan` instance in the DNF format that is expected by PyArrow for performing predicate pushdown. Note this PR only adds support for `Expr::InList` for now.

An example of calling this functionality from Python goes like this

```python
# Rust table_scan instance handle obtained from LogicalPlan instance, "rel"
table_scan = rel.table_scan()
dnf_filters = table_scan.getDNFFilters()

# dnf_filters is of type `PyFilteredResult` from Rust. Contains the filters that are not supported by PyArrow DNF
# and also ones that can.
none_filterable_exprs = dnf_filters.io_unfilterable_exprs()
filterable_exprs = dnf_filters.filtered_exprs() # Tuple of (str, str, [str]) Ex: ('id', 'in', ['Int32(1)', 'Int32(2)'])

```


dask_planner/src/sql/logical/table_scan.rs
```diff
@@ -16,6 +17,73 @@ pub struct PyTableScan {
...
+        let mut filter_tuple: Vec<(String, String, Vec<String>)> = Vec::new();
+
+        match filter {
+            Expr::InList {
+                expr,
+                list,
+                negated: _,
+            } => {
+                let il: Vec<String> = list.iter().map(|f| f.canonical_name()).collect();
+                filter_tuple.push((expr.canonical_name(), "in".to_string(), il));
```
ayushdg, line 50: Does it make sense to check here if `expr` is a column or some scalar kind of type?
Can't think of a example but for cases where the expr instance maps to a binaryexpr or something that can't be applied directly, we'd want to use the unfiltered case.

dask_planner/src/sql/logical/table_scan.rs
```diff
@@ -16,6 +17,101 @@ pub struct PyTableScan {
...
+    // are stored here in a DNF format that is expected by pyarrow.
+    #[pyo3(get)]
+    pub filtered_exprs: Vec<(String, String, Vec<String>)>,
+}
+
+impl PyTableScan {
+    /// Ensures that a valid Expr variant type is present
+    fn _valid_expr_type(expr: &[Expr]) -> bool {
+        expr.iter()
+            .all(|f| matches!(f, Expr::Column(_) | Expr::Literal(_)))
```
sarahyurick, line 39: I like this for now. In https://github.com/dask-contrib/dask-sql/pull/1102 it also only deals with columns and literal values at the moment.

In the future we can add handling for more complex expressions. I have this listed in points 2 and 3 of https://github.com/dask-contrib/dask-sql/issues/1121 for DPP as well, so maybe down the road this PR and DPP can share some of the same logic to achieve this.

dask_planner/src/sql/logical/table_scan.rs
```diff
@@ -68,3 +170,45 @@ impl TryFrom<LogicalPlan> for PyTableScan {
...
+
+    #[test]
+    fn dnf_inlist() -> Result<()> {
+        let schema = Schema::new(vec![Field::new("id", DataType::Int32, false)]);
+
+        // Dummy logical plan
+        let plan = Arc::new(
+            table_scan(TableReference::none(), &schema, None)
+                .unwrap()
+                .filter(col("id").eq(Expr::Placeholder {
```
sarahyurick, line 192: Just wanted to ask why an `Expr::Placeholder` is used here instead of an `Expr::InList`?

sarahyurick (commented):
Thanks @jdye64 ! Left a few comments, changes generally look good to me.

dask_planner/src/sql/logical/table_scan.rs
jdye64: Short answer: Laziness lol

Long answer: `PyExpr` currently requires a `LogicalPlan` instance to perform schema based queries. However, that is removed in #1085 so once that merges this snippet of code will cause a compilation error and need removed anyway. So knowing that I got a little lazy and just copied a `LogicalPlan` node from another test in the codebase.

You are right though, I should, and will update it to use `Expr::InList`

sarahyurick (approved):
LGTM

ayushdg (commented):
Playing around with the example a bit, we end up getting `filterable_exprs` that looks something like: 
```
('df.a', 'in', ['Int64(1)', 'Int64(2)', 'Int64(3)', 'Int64(4)', 'Int64(5)'])
```

when passing this down to the reader we'd see the following issues:
`df.a` the reader doesn't understand this and is probably expecting `a`. I can just split by `.` on the python side but was curious if there's something better than `canonical` name that would give the column name.

`'Int64(1)'` Doesn't map to anything. Is it possible to return the literal values in this case? If not we would probably need to handle this on the python side.

jdye64:
I need some pieces from #1085 to complete this so waiting on that to merge first.

dask_planner/src/sql/logical/table_scan.rs
```diff
@@ -18,6 +19,129 @@ pub struct PyTableScan {
...
+                                ScalarValue::Int16(val) => val.unwrap().into_py(py),
+                                ScalarValue::Int32(val) => val.unwrap().into_py(py),
+                                ScalarValue::Int64(val) => val.unwrap().into_py(py),
+                                ScalarValue::UInt8(val) => val.unwrap().into_py(py),
+                                ScalarValue::UInt16(val) => val.unwrap().into_py(py),
+                                ScalarValue::UInt32(val) => val.unwrap().into_py(py),
+                                ScalarValue::UInt64(val) => val.unwrap().into_py(py),
+                                ScalarValue::Utf8(val) => val.clone().unwrap().into_py(py),
+                                ScalarValue::LargeUtf8(val) => val.clone().unwrap().into_py(py),
+                                _ => "".into_py(py),
```
ayushdg, line 91: For an unsupported scalar type we should raise an error and have it go via the unfiltered_exprs path

dask_planner/src/sql/logical/table_scan.rs
```diff
@@ -18,6 +19,129 @@ pub struct PyTableScan {
...
+                            },
+                            _ => f.canonical_name().into_py(py),
+                        })
+                        .collect();
+
+                    filter_tuple.push((ident.unwrap_or(expr.canonical_name()), op.to_string(), il));
```
ayushdg, line 97: Would prefer the same check for `expr` being a alias or column type.

ayushdg (changes requested):
Might also be worth adding a few tests on the rust side to ensure things are working as expected.

jdye64:
@ayushdg had Rust tests before. Would have loved to have kept it but now that we using the GIL in these methods it seems there is a bug in PyO3 that is preventing me for doing so. Once again this issue only occurs when using the Python GIL in Rust tests. Take a look at this FAQ page. I tried all of their recommendations but none of them worked. Still kept hitting linker errors so the project fails to build if I include them.  https://pyo3.rs/v0.18.3/faq

Long story short I don't see a way we can add those Rust tests until that issue is resolved on the PyO3 side.

dask_planner/src/sql/logical/table_scan.rs
```diff
@@ -18,6 +19,138 @@ pub struct PyTableScan {
...
+        filters: &[Expr],
+        py: Python,
+    ) -> PyFilteredResult {
+        let mut filtered_exprs: Vec<(String, String, Vec<PyObject>)> = Vec::new();
+        let mut unfiltered_exprs: Vec<PyExpr> = Vec::new();
+
+        filters
+            .iter()
+            .for_each(|f| match PyTableScan::_expand_dnf_filter(f, py) {
+                Ok(mut expanded_dnf_filter) => filtered_exprs.append(&mut expanded_dnf_filter),
```
ayushdg, line 141: Just leaving this in as a note for followup work:
The approach of appending to the filtered list of expr's wouldn't return a dnf for more complex cases and right now is only limited to expr's that are `and` with each other.
eg: `a in (1,2,3,4,5) or b=5` would not get expanded by this logic.

ayushdg (approved):

PR #1131, opened 2023-05-02
Add configuration variable for CPU/GPU decimal support

charlesbluca:
This should unblock CPU support for decimal scalar/columns when working in an environment with cuDF installed; note that this means that decimal support for cuDF must now be enabled manually with something like:

```python
dask.config.set({"sql.mappings.decimal_support": "cudf"})
```

Once #1117 is further along ideally this would be set implicitly based on the CPU/GPU flag of the context.


ayushdg (approved):

charlesbluca:
rerun tests

PR #1141, opened 2023-05-09
Add C compiler to conda environments / recipes

charlesbluca:
Adds `c-compiler` to our conda environment files and `compiler('c')` to our conda nightly recipe, so that the package can be successfully tested/built on systems that don't have `cc` pre-installed; haven't placed much thought into what version we should pin to (if any).

### DIFF
--- a/continuous_integration/environment-3.10-dev.yaml
+++ b/continuous_integration/environment-3.10-dev.yaml
@@ -3,6 +3,7 @@ channels:
 - conda-forge
 - nodefaults
 dependencies:
+- c-compiler
 - dask>=2022.3.0
 # FIXME: handling is needed for httpx-based fastapi>=0.87.0
 - fastapi>=0.69.0,<0.87.0
--- a/continuous_integration/recipe/meta.yaml
+++ b/continuous_integration/recipe/meta.yaml
@@ -22,8 +22,8 @@ build:
 
 requirements:
   build:
-    - {{ compiler('c') }} =11
-    - {{ compiler('rust') }} >=1.65.0
+    - {{ compiler('c') }}
+    - {{ compiler('rust') }}
     - setuptools-rust >=1.5.2
   host:
     - pip
--- a/continuous_integration/recipe/meta.yaml
+++ b/continuous_integration/recipe/meta.yaml
@@ -22,6 +22,7 @@ build:
 
 requirements:
   build:
+    - {{ compiler('c') }} =11
     - {{ compiler('rust') }} >=1.65.0
     - setuptools-rust >=1.5.2
   host:
--- a/continuous_integration/environment-3.8-dev.yaml
+++ b/continuous_integration/environment-3.8-dev.yaml
@@ -3,6 +3,7 @@ channels:
 - conda-forge
 - nodefaults
 dependencies:
+- c-compiler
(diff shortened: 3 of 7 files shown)


continuous_integration/recipe/meta.yaml
```diff
@@ -22,6 +22,7 @@ build:
 
 requirements:
   build:
+    - {{ compiler('c') }} =11
```
charlesbluca, line 25: I set this pin somewhat arbitrarily, but it seems like GCC 11 successfully built the shared object locally for me

jakirkham: Would do this by adding a `conda_build_config.yaml` file with `c_compiler_version` specified (like [so]( https://github.com/conda-forge/conda-forge-pinning-feedstock/blob/6debcd5a454e7163cd1adeae066906617feb6622/recipe/conda_build_config.yaml#L6-L7 ))

jakirkham (commented):
Thanks Charles! 🙏

Added a few suggestions below

continuous_integration/recipe/meta.yaml
```diff
@@ -22,6 +22,7 @@ build:
...
     - {{ compiler('rust') }} >=1.65.0
```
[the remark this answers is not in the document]
charlesbluca: Are we still able to specify a lower bound like thing with something like

```yaml
rust_compiler_version:
    - '>=1.65'
```

Tried this locally and things seemed to work fine, but not sure if this is just because the compiler version is invalid and getting ignored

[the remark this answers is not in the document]
jakirkham: Think that would work

Though it could also make sense to pick a version for reproducibility reasons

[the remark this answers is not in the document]
charlesbluca: That's fair, I'll go with the latest version being picked up for now (1.69) and we can revisit if there are any issues around that.

continuous_integration/environment-3.10-dev.yaml
```diff
@@ -3,6 +3,7 @@ channels:
 - conda-forge
 - nodefaults
 dependencies:
+- c-compiler
```
charlesbluca, line 6: Despite this being tightly pinned in conda builds, I don't think there's an easy way to pin gcc in the conda environments? `c-compiler` brings in different compiler packages based on platform, and without the use of conditional selectors I don't think there's a way to pin all of those compiler packages in a single conda environment without causing conflict on at least one platform.

cc @jakirkham if you know of a way to achieve this

jakirkham: Could have different environment files per platform

charlesbluca: Yeah that makes sense - I think that for the sake of minimizing environment files to manage, I'm going to leave things as is since CI seems to be passing with GCC left unpinned. If we do run into failures at some point due to leaving it this way, I'd imagine that platform-specific environment files is the best workaround.

jakirkham: Yeah that makes sense. Would probably do the same thing

Happy to revisit if it becomes more relevant

PR #1143, opened 2023-05-10
Drop python 3.8, add 3.11 & 3.12 to testing

charlesbluca:
With RAPIDS planning to drop python 3.8 support in 23.06 and discussion happening in https://github.com/dask/community/issues/315 around Dask doing the same relatively soon, now seems like a good time to begin exploring dropping 3.8 support/testing ourselves.


ayushdg:
A few more places we can update:
https://github.com/dask-contrib/dask-sql/blob/ab7340b1e8e374a462a00dd8c3ee5b2d837e0aeb/dask_planner/Cargo.toml#L16
https://github.com/dask-contrib/dask-sql/blob/ab7340b1e8e374a462a00dd8c3ee5b2d837e0aeb/setup.py#L42

Can be done in a followup but we could remove a bunch of instances in the code for type hints with `from typing import ...` as it should just work py3.9 and above

charlesbluca:
There are some intermittent mamba env solve failures I'm trying to work through (could be worth getting a minimal example together to open an issue in setup-miniconda), but seems like in general the main blocker here is lack of `mlflow` 3.11 packages in conda-forge; going to try rerunning https://github.com/conda-forge/mlflow-feedstock/pull/120 to see if there are any blockers there.

.github/workflows/release.yml
```diff
@@ -16,7 +16,8 @@ jobs:
       fail-fast: false
       matrix:
         os: [ubuntu-latest, windows-latest, macos-latest]
-        python: ["8", "9", "10"] # 3.x
+        # corresponds to python 3.9, 3.10, 3.11
+        python: ["9", "10", "11"]
```
charlesbluca, line 20: In an ideal world, we would be using the full python versions here, but since manipulation of string variables in GHA is difficult I think adding this comment to make things more easily grep-able is a fair compromise

ayushdg:
Thanks for looking into this. Assuming sail cannot be patched in time for 3.11 we might want to evaluate skipping hive tests for 3.11.
For CIBW happy to sync to get the builds around protobuf working.

sarahyurick:
I think gpuCI should pass now after merging main.

charlesbluca:
Going to guess that the issues in the conda build are caused by the recent release of `anyio=3.7`; going to check if I can trivially reproduce to file an issue and unblock things on our end for now

PR #1144, opened 2023-05-11
Start building `linux-aarch64` nightlies

charlesbluca:
From internal conversations with RAPIDS ops, it sounds like there would be some benefit to having `linux-aarch64` nightlies available for testing; this could also be of some benefit to us in that it gives some better insights into what issues we would run into during cross-compilation (for example, working on this PR uncovered that we needed to add `libprotobuf` to our build requirements).


continuous_integration/recipe/meta.yaml
```diff
@@ -42,15 +45,15 @@ requirements:
...
-    - pip
+test:                                        # [build_platform == target_platform]
+  imports:                                   # [build_platform == target_platform]
+    - dask_sql                               # [build_platform == target_platform]
+  commands:                                  # [build_platform == target_platform]
+    - pip check                              # [build_platform == target_platform]
+    - dask-sql-server --help                 # [build_platform == target_platform]
+    - dask-sql --help                        # [build_platform == target_platform]
+  requires:                                  # [build_platform == target_platform]
+    - pip                                    # [build_platform == target_platform]
```
charlesbluca, line 56: When we run tests on the `linux-aarch64` packages, we get the following error:

```
/usr/share/miniconda3/envs/test/conda-bld/dask-sql_1683837509285/test_tmp/conda_test_runner.sh: line 3: /usr/share/miniconda3/envs/test/conda-bld/dask-sql_1683837509285/_test_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehol/bin/python: cannot execute binary file: Exec format error
```

Which I'm assuming is because we're attempting to use ARM python on an x86 runner? Not sure if there's a way to unblock tests outside of outright skipping tests for the `linux-aarch64` packages (@jakirkham not sure if you have any thoughts here)

ayushdg:
Given the long build times for the conda nightly, it might make sense to have this run at a schedule rather than a per pr basis. Also for a quick test would it be possible to upload the artifacts from these runs to github? I wanted to give it a try on an aarch64 machine

continuous_integration/recipe/meta.yaml
jakirkham: Yeah we can't run tests on a different architecture from the binaries we have

Instead of using selectors here, would suggest making the following change to the GHA workflow

```yaml
      - name: Build conda packages
        run: |
          # suffix for nightly package versions
          export VERSION_SUFFIX=a`date +%y%m%d`
          conda mambabuild continuous_integration/recipe \
                           --no-anaconda-upload \
                           --no-test
                           --output-folder .

      - name: Test conda package
        run: |
          conda mambabuild --test linux-64/*.tar.bz2
```

This way one can still run the tests on the package locally (if they have the right architecture) and we skip them on CI when we don't have that architecture

continuous_integration/recipe/meta.yaml
```diff
@@ -22,9 +22,12 @@ build:
 
 requirements:
   build:
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
     - {{ compiler('c') }}
     - {{ compiler('rust') }}
     - setuptools-rust >=1.5.2
+    - libprotobuf =3
```
jakirkham, line 30: Curious what prompted adding `libprotobuf` to `build`?

jakirkham (commented):
Thanks Charles! 🙏

Had a couple comments below

continuous_integration/recipe/meta.yaml
charlesbluca: I would've thought we needed it here before this PR, as adding `datafusion-substrait` to our Rust dependencies once required `protoc` for compilation (cc @jdye64), but I suppose it was getting picked up transitively up until now?

I recall getting failures building the `linux-aarch64` packages due to missing the `protoc`, can double check that this is the case; if so, I suppose we could add a selector here to only pick this up for `linux-aarch64`, but thought it made sense to specify explicitly for all builds.

jakirkham: That should already be possible with the dependency in `host`. No need to have it in `build` also

Would think of `libprotobuf` more like `python` and less like `cmake` as it is something we want in the environment installed into as we may want to link to it

charlesbluca: Looks like builds are failing without `libprotobuf` specified as a build-time requirement:

```
  error: failed to run custom build command for `substrait v0.7.5`
  note: To improve backtraces for build dependencies, set the CARGO_PROFILE_RELEASE_BUILD_OVERRIDE_DEBUG=true environment variable to enable debug information generation.

  Caused by:
    process didn't exit successfully: `/usr/share/miniconda3/envs/test/conda-bld/dask-sql_1684428974210/work/dask_planner/target/release/build/substrait-79784529fc7e3b5c/build-script-build` (exit status: 1)
    --- stdout
    cargo:rerun-if-env-changed=FORCE_REBUILD
    cargo:rerun-if-changed=substrait
    cargo:rerun-if-changed=substrait/text/simple_extensions_schema.yaml
    cargo:rerun-if-changed=substrait/proto/substrait/algebra.proto
    cargo:rerun-if-changed=substrait/proto/substrait/type_expressions.proto
    cargo:rerun-if-changed=substrait/proto/substrait/extensions/extensions.proto
    cargo:rerun-if-changed=substrait/proto/substrait/plan.proto
    cargo:rerun-if-changed=substrait/proto/substrait/extended_expression.proto
    cargo:rerun-if-changed=substrait/proto/substrait/capabilities.proto
    cargo:rerun-if-changed=substrait/proto/substrait/type.proto
    cargo:rerun-if-changed=substrait/proto/substrait/function.proto
    cargo:rerun-if-changed=substrait/proto/substrait/parameterized_types.proto

    --- stderr
    Error: Custom { kind: Other, error: "protoc failed: /usr/share/miniconda3/envs/test/conda-bld/dask-sql_1684428974210/_h_env_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_placehold_pla/bin/protoc: 1: Syntax error: \"(\" unexpected\n" }
  warning: build failed, waiting for other jobs to finish...
```

cc @jdye64 not sure if this looks familiar to you?

continuous_integration/recipe/conda_build_config.yaml
```diff
@@ -6,3 +6,6 @@ c_compiler_version:
     - 11
```
jakirkham, line 6:
```suggestion
    - 11
libprotobuf:
    - 3
```

jakirkham (commented):
Ok sounds like we need something more like this then

Also would drop the ` =3` from `libprotobuf` in `host`

continuous_integration/recipe/meta.yaml
jakirkham: Added new review comments below related to this

continuous_integration/recipe/conda_build_config.yaml
```diff
@@ -4,5 +4,10 @@ python:
     - 3.10
 c_compiler_version:
     - 11
+libprotobuf:
+    - 3
 rust_compiler_version:
     - 1.69
```
jakirkham, line 10:
```suggestion
setuptools_rust:
    - 1.5.2
```

Or maybe some newer version if desired

continuous_integration/recipe/meta.yaml
```diff
@@ -22,14 +22,17 @@ build:
 
 requirements:
   build:
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
     - {{ compiler('c') }}
     - {{ compiler('rust') }}
-    - setuptools-rust >=1.5.2
+    - libprotobuf                            # [build_platform != target_platform]
+    - setuptools-rust >=1.5.2                # [build_platform != target_platform]
```
jakirkham, line 30:
```suggestion
    - setuptools-rust                        # [build_platform != target_platform]
```

continuous_integration/recipe/meta.yaml
```diff
@@ -22,14 +22,17 @@ build:
...
   host:
     - pip
     - python
     - setuptools-rust >=1.5.2
```
jakirkham, line 34:
```suggestion
    - setuptools-rust
```

jakirkham (commented):
Maybe we can move the `setuptools-rust` version into `conda_build_config.yaml` as well to reduce duplication

jakirkham (commented):
Minor nit: Maybe we can group the cross-compilation changes together to make things a bit more readable?

jakirkham (approved):
LGTM. Thanks Charles! 🙏

Let's see if anything else is needed based on CI

Think it would be good to update the conda-forge recipe with these changes. Could happen in the next release. We could just raise an issue to track for now

.github/workflows/conda.yml
```diff
@@ -49,23 +54,34 @@ jobs:
...
                            --output-folder .
-      - name: Upload conda package
+      - name: Test conda packages
+        if: matrix.arch == 'linux-64'  # can only test native platform packages
+        run: |
+          conda mambabuild --test ${{ matrix.arch }}/*.tar.bz2
+      - name: Upload conda packages as artifacts
+        uses: actions/upload-artifact@v3
+        with:
+          name: "conda nightlies (python: ${{ matrix.python }}, arch: ${{ matrix.arch }})"
```
jakirkham, line 75: Looks like `:` causes [parsing errors]( https://github.com/dask-contrib/dask-sql/actions/runs/5026700402/jobs/9015287457?pr=1144#step:7:14 )

Maybe this would work?

```suggestion
          name: "conda nightlies (python - ${{ matrix.python }}, arch - ${{ matrix.arch }})"
```

jakirkham (commented):
Thanks Charles! 🙏

This is looking good

GHA flagged a minor parsing error. Added a suggestion below

.github/workflows/conda.yml
```diff
@@ -49,23 +54,34 @@ jobs:
...
+          path: ${{ matrix.arch }}/*.tar.bz2
+      - name: Upload conda packages to Anaconda
         if: |
           github.event_name == 'push'
           && github.repository == 'dask-contrib/dask-sql'
         env:
           ANACONDA_API_TOKEN: ${{ secrets.DASK_CONDA_TOKEN }}
-          LABEL: ${{ github.ref == 'refs/heads/datafusion-sql-planner' && 'dev_datafusion' || 'dev' }}
```
jakirkham, line 66: Did these get consolidated?

charlesbluca: Yeah, should no longer need the logic for that branch - there's probably some other things that can be removed as well

jakirkham:
Looks like that passed (excluding gpuCI) 🎉

jakirkham:
Maybe there are some clues on [how to configure Rust cross-compilation in this doc]( https://github.com/japaric/rust-cross#readme )

continuous_integration/recipe/meta.yaml
```diff
@@ -24,12 +24,15 @@ requirements:
   build:
     - {{ compiler('c') }}
     - {{ compiler('rust') }}
-    - setuptools-rust >=1.5.2
+    - libprotobuf                            # [build_platform != target_platform]
+    - python                                 # [build_platform != target_platform]
+    - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+    - setuptools-rust                        # [build_platform != target_platform]
```
jakirkham, line 30: Think we may want to drop `setuptools-rust` from `build` after looking at [the `requirements` of another `setuptools-rust`-based package, `cryptography`]( https://github.com/conda-forge/cryptography-feedstock/blob/e70d84149adf02d7482113fea612ea2211a4b10c/recipe/meta.yaml#L61-L77 ) and only have it in `host`. Looks like [`maturin` does the same thing]( https://github.com/conda-forge/maturin-feedstock/blob/6528b97e83ca8d708f3300938cc1aa2eb0f86b02/recipe/meta.yaml#L18-L35 )

```suggestion
```

jakirkham (commented):
Perhaps the bug is due to an issue in our requirements?

continuous_integration/recipe/meta.yaml
jakirkham: Done below

jdye64 (approved):
LGTM. Thanks for struggling through that one @charlesbluca and @jakirkham that one was a bear

continuous_integration/recipe/meta.yaml
```diff
@@ -43,6 +43,8 @@ requirements:
     - prompt-toolkit >=3.0.8
     - pygments >=2.7.1
     - tabulate
+  run_constrained:
+    - __glibc >=2.32  # [aarch64]
```
jakirkham, line 47: For the ARM wheels, think we need to do something like

```
wheel tags --platform-tag=manylinux_2_32_aarch64 <filename>.whl
```

jakirkham: This is being handled in PR ( https://github.com/dask-contrib/dask-sql/pull/1152 ) with commit ( https://github.com/dask-contrib/dask-sql/pull/1152/commits/3284141fc83f50dbfbd38236f448f1e33f09043c )

continuous_integration/recipe/meta.yaml
```diff
@@ -41,6 +43,8 @@ requirements:
...
```
jakirkham, line 47: Suggesting adding the same comment that was added for wheels here

```suggestion
  run_constrained:
    # FIXME: aarch64 builds require glibc>=2.32 to prevent TLS allocation error when importing;
    # is there any way we could modify the TLS model of its shared object to avoid this?
    # https://bugzilla.redhat.com/show_bug.cgi?id=1722181
    # xref: https://github.com/dask-contrib/dask-sql/issues/1169
    - __glibc >=2.32  # [aarch64]
```

ayushdg (approved):

PR #1145, opened 2023-05-12
Obtain `filepath` from Dask DataFrame

sarahyurick:
Builds upon https://github.com/dask-contrib/dask-sql/pull/1074

Uses part of https://github.com/jdye64/dask-sql/blob/expand_predicate_pushdown/dask_sql/physical/rel/logical/table_scan.py#L112-134

~~For some reason, this doesn't work with https://github.com/dask-contrib/dask-sql/pull/1102 yet, indicating some error on the DPP side(?), even though DPP is able to obtain the correct filepaths by using this PR. So DPP still just returns the original plan when `create_table` uses a Dask DataFrame instead of a Parquet filepath.~~


dask_sql/context.py
```diff
@@ -247,6 +248,15 @@ def create_table(
         if type(input_table) == str:
             dc.filepath = input_table
             self.schema[schema_name].filepaths[table_name.lower()] = input_table
+        elif "dask" in str(type(input_table)):
```
jdye64, line 251: If the function achieves your desired result, could you use `dd.utils.is_dataframe_like` since we use it in other places. That function might possibly include a wider positive result for dataframe like things you don't wish to include here however.

sarahyurick: Thanks!

tests/unit/test_context.py
```diff
@@ -342,3 +343,30 @@ def test_filepath(tmpdir):
     assert c.schema["root"].tables["df"].filepath is None
     with pytest.raises(KeyError):
         c.schema["root"].filepaths["df"]
+
+
+def test_ddf_filepath(tmpdir):
```
ayushdg, line 348: Does the parquet_ddf fixture work for the test here, or does it explicitly need to be redefined here?

ayushdg (approved):

PR #1147, opened 2023-05-16
Give `conda-forge` higher priority than `nvidia` in GPU environments

charlesbluca:
Subtle difference between the GPU environments in this repo versus those used in the GPU CI runs, but important because having `conda-forge` above `nvidia` prevents us from installing `cuda-toolkit`, which would otherwise cause some conflicts with `libcudf`'s recent addition of `libcufile` as a dependency.


ayushdg:
Are the gpuCI errors here unrelated?

charlesbluca:
rerun tests

ayushdg:
rerun tests

ayushdg:
rerun tests

PR #1148, opened 2023-05-16
Enable case <expr> when syntax

ChrisJar:
Ex: `case a when 0 then NULL` which is equivalent to `case when a=0 then NULL`


ayushdg:
cc: @jdye64

jdye64 (approved):
Logic makes sense to me. Thanks for adding this @ChrisJar. Is there an issue for this? I ask because I likely will need to implement this in Arrow DataFusion Python later as well or this logic might accidentally be removed.

jdye64:
FYI - Made PR to mimic this behavior in Arrow DataFusion Python https://github.com/apache/arrow-datafusion-python/pull/388

ayushdg (approved):

charlesbluca (approved):

PR #1152, opened 2023-05-19
Install `protoc` in cibuildwheel

charlesbluca:
We need to install `protoc` as part of our wheel building CI so that it's available during compilation of `substrait`.

cc @ayushdg


.github/workflows/release.yml
```diff
@@ -2,6 +2,15 @@ name: Upload Python package
...
     types: [created]
+  pull_request:
+    paths:
+      - .github/workflows/release.yml
+
+# When this workflow is queued, automatically cancel any previous running
+# or pending jobs from the same branch
+concurrency:
+  group: release-${{ github.head_ref }}
+  cancel-in-progress: true
```
charlesbluca, line 13: cc @ayushdg okay with leaving this in for other PRs modifying the workflow?

ayushdg: Yeah. I'd actually prefer this so that we start testing CIBW updates in an automated way as well.

.github/workflows/release.yml
```diff
@@ -68,6 +91,9 @@ jobs:
           name: wheels for py3.${{ matrix.python }} on ${{ matrix.os }}
           path: dist/*
       - name: Publish package
+        if: |
+          github.event_name == 'release'
+          && github.repository == 'dask-contrib/dask-sql'
```
ayushdg, line 96: Minor nit: It might make sense to consolidate this and the case below to a common condition that sets an environment variable that can be picked up here and below

ayushdg (approved):
Minor suggestions but generally lgtm!

galipremsagar (approved):

.github/workflows/release.yml
```diff
@@ -56,6 +82,14 @@ jobs:
           use-mamba: true
           python-version: "3.8"
           channel-priority: strict
+      - name: Update platform tag for aarch64 wheel
```
bdice, line 85: If this is working around a bug or other issue, it would be good to leave some comments or a link to an issue describing that.

charlesbluca: Agreed, will leave a comment linking to the glibc bug itself:

https://bugzilla.redhat.com/show_bug.cgi?id=1722181

jakirkham: Looks like there is now a comment above this line with that note. Marking this resolved

.github/workflows/release.yml
```diff
@@ -56,6 +82,17 @@ jobs:
           use-mamba: true
           python-version: "3.8"
           channel-priority: strict
+      # FIXME: aarch64 builds require glibc>=2.32 to prevent TLS allocation error when importing;
+      # is there any way we could modify the TLS model of its shared object to avoid this?
+      # https://bugzilla.redhat.com/show_bug.cgi?id=1722181
```
jakirkham, line 87:
```suggestion
      # FIXME: aarch64 builds require glibc>=2.32 to prevent TLS allocation error when importing;
      # is there any way we could modify the TLS model of its shared object to avoid this?
      # https://bugzilla.redhat.com/show_bug.cgi?id=1722181
      # xref: https://github.com/dask-contrib/dask-sql/issues/1169
```

vyasr (approved):
Approving the wheel retagging, but agree with @jakirkham about adding additional comments linking to the relevant issue (looks like the glibc bug at least is linked though which is good).

jakirkham (approved):

jakirkham:
Thanks Charles for the PR and everyone for the reviews! 🙏

.github/workflows/release.yml
```diff
@@ -85,14 +85,15 @@ jobs:
       # FIXME: aarch64 builds require glibc>=2.32 to prevent TLS allocation error when importing;
       # is there any way we could modify the TLS model of its shared object to avoid this?
       # https://bugzilla.redhat.com/show_bug.cgi?id=1722181
+      # xref: https://github.com/dask-contrib/dask-sql/issues/1169
       - name: Update platform tag for aarch64 wheel
         if: matrix.os == 'ubuntu-latest'
         run: |
           mamba install wheel
 
           wheel tags --platform-tag=manylinux_2_32_aarch64 \
```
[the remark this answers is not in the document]
ayushdg: Thanks for the link. From PEP 599 it does seem like depending on glib > 2.17 would not make it manylinux2014 compliant too it probably doesn't make sense to add that to the tag.

ayushdg (approved):

.github/workflows/release.yml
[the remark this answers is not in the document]
jakirkham: [PEP 600]( https://peps.python.org/pep-0600/ ) outlines [the naming scheme]( https://peps.python.org/pep-0600/#core-definition ):

>`manylinux_${GLIBCMAJOR}_${GLIBCMINOR}_${ARCH}`

PR #1153, opened 2023-05-23
Fix join logic involving nulls

sarahyurick:
This PR fixes a bug in the join code that tries to assign a Series from an unfiltered DataFrame to a DataFrame after dropping nulls, which was causing a `ValueError: All columns must be of equal length`.

cc @ayushdg


ayushdg:
rerun tests

sarahyurick:
I think gpuCI will fail until we can get https://github.com/rapidsai/ops/issues/2672 uploaded.

ayushdg:
rerun tests

ayushdg (commented):
It might be worth adding a test with nulls as join keys to confirm the fix

ayushdg:
Also cc: @charlesbluca for review since you've looked at some of the join code before

charlesbluca:
@jdye64 I'm okay with xfailing the remaining cuML tests for now, as long as we mark them with a TODO pointing to https://github.com/rapidsai/cuml/pull/5434 so we can track when they should be un-xfailed

tests/integration/test_join.py
```diff
@@ -523,3 +523,15 @@ def test_broadcast_join(c, client, gpu):
...
+def test_null_key_join(c):
+    df1 = pd.DataFrame({"a": [None, None, None, None, 2], "c": [None, 1, 2, 3, 4]})
+    c.create_table("df1", df1)
+    df2 = pd.DataFrame({"b": [None, None, 2, 3, 4], "d": [None, None, None, 2, 3]})
+    c.create_table("df2", df2)
+
+    result_df = c.sql("SELECT d FROM df1 JOIN df2 ON d=c")
+    expected_df = c.sql("SELECT d FROM df1, df2 WHERE d=c")
+
+    assert_eq(result_df, expected_df)
```
charlesbluca, line 537: This test seems to pass for me without this patch? Think that the assign failures might be a little more complicated than just mismatched lengths on the frame/series, trying to make a reproducer of the failure now

charlesbluca: Was able to put together a small reproducer:

```python
import pandas as pd

from dask_sql import Context

df1 = pd.DataFrame({"a": [None, None, None, None, None, 1]})
df2 = pd.DataFrame({"b": [1.0, 2.0, 3.0, 4.0, 5.0, 6.0]})

c = Context()

c.create_table("df1", df1, gpu=True)
c.create_table("df2", df2, gpu=True)

print(c.explain("SELECT * FROM df1 JOIN df2 ON a=b"))

df = c.sql("SELECT * FROM (select * from df1 limit 5) JOIN (select * from df2 limit 5) ON a=b")
```

Looks like this only fails when using GPU dataframes? Which might reflect some API differences between `DataFrame.assign` that could be worth raising an issue over.

The underlying failure here is attempting to assign a non-empty column to an empty dataframe; note that we actually need to force the all nulls table with a `LIMIT` statement; this is because we don't enter inner join logic if a cast operation would be required as part of the join condition, instead pushing it to the post-join filtering step of an outer join.

sarahyurick: Thanks @charlesbluca !

ayushdg: Thanks. Given that it shouldn't be too hard to get a pandas vs cuDF repro from an assign based on the example above, it might also be worth raising on the cuDF side as an issue

ayushdg (approved):

Issue #607, opened 2022-06-29
[DF] [ENH] Add series.isin support for predicate pushdown


ayushdg:
**Is your feature request related to a problem? Please describe.**
With datafusion we now map `column in (val1,val2,val3)` like operations to `series.isin`. This operator does not support predicate pushdown

**Describe the solution you'd like**
Add support for predicate pushdown for `series.isIn` operation. 

**Describe alternatives you've considered**
N/A

**Additional context**
N/A

rjzamora:
**Minor update**: I started exploring this feature and found that we probably need to make a small change in `dask.dataframe` to make it easier to "regenerate" an `isin` layer. Right now, the `asin` implementation [is wrapping the iterable `values` argument in delayed](https://github.com/dask/dask/blob/352d934d11a144b9dafde79a933584a7a090b615/dask/dataframe/core.py#L3089). As far as I can tell, we should be able to avoid this behavior for simple `list`/`tuple` arguments.

With the upstream `isin` method modified, I was able to add `M.isin`/`dd._Frame.isin` to the logic in `dask-sql/dask_sql/physical/utils/filter.py` to produce the following behavior:

```python
import dask.dataframe as dd
from dask_sql.physical.utils.filter import attempt_predicate_pushdown
from dask.utils_test import hlg_layer

ddf = dd.read_parquet("testfile.parquet")
ddf = ddf[ddf["A"].isin([0,2,4,6,8])]
ddf2 = attempt_predicate_pushdown(ddf)

# Check filters
print(hlg_layer(ddf2.dask, "read-parquet").creation_info['kwargs']['filters'])

# Check compute
print("\n",ddf2.compute())
```

```
[[('A', 'in', (0, 2, 4, 6, 8))]]

    A
0  0
1  2
2  4
3  6
4  8
```

rjzamora:
>As far as I can tell, we should be able to avoid this behavior for simple list/tuple arguments.

It looks like I was wrong about this (see: https://github.com/dask/dask/pull/4727).  Rather than removing the `delayed` usage in Dask, I may need to add a simple `DelayedObject` layer to dask (so we get something other than a `MaterializedLayer`)

PR #1160, opened 2023-05-26
[WIP] Support predicate-pushdown for `isin` operations

rjzamora:
Closes https://github.com/dask-contrib/dask-sql/issues/607

This PR adds predicate-pushdown support for `dd._Frame.isin(values)`, and uses a somewhat-ugly workaround to account for the fact that `dask.dataframe` will produce a distinct `MaterializedLayer` object for the `values` argument to `isin`.

Note that this PR also addresses `"not in"` support. However, `dask.dataframe` will produce the wrong result for these filters until https://github.com/dask/dask/pull/10320 is merged.

### DIFF
--- a/dask_sql/physical/utils/filter.py
+++ b/dask_sql/physical/utils/filter.py
@@ -178,6 +178,7 @@ _regenerable_ops = set(_comparison_symbols.keys()) | {
     operator.and_,
     operator.or_,
     operator.getitem,
+    operator.inv,
     M.fillna,
     M.isin,
 }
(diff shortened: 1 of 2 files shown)


rjzamora:
@ayushdg @randerzander - Note that I added `isna` support to this PR, but I'd be happy to split that change into a separate PR if you prefer.

ayushdg (commented):
Changes generally lgtm!
Doesn't need to be in this pr, but as a followup it might also make sense to add simpler unit tests that just use the attempt predicate pushdown method on dask dataframes and add them to https://github.com/dask-contrib/dask-sql/tree/main/tests/unit

rjzamora:
@ayushdg - I'm thinking we should layer on something like #1140 as a follow up. Let me know what you think.

ayushdg:
Yeah it makes sense to add something that conceptually achieves what #1140 does without some of the extra logic/handling in that pr. We can address that in a followup and some of the logic from #1140 should be redundant with dask-cudf 22.06 and newer

ayushdg (approved):
Thanks a lot for looking into this @rjzamora!

PR #1166, opened 2023-05-30
Constrain `anyio<3.7` to unblock conda builds

charlesbluca:
It looks like there are some problems with the runtime dependencies of the latest `anyio` release on conda-forge that are causing us issues in our conda builds:

https://github.com/dask-contrib/dask-sql/actions/runs/5124690983/jobs/9216865588

Currently exploring if we can fix this in a new build in https://github.com/conda-forge/anyio-feedstock/pull/48, but this should unblock us in the meantime.

### DIFF
--- a/continuous_integration/environment-3.10-dev.yaml
+++ b/continuous_integration/environment-3.10-dev.yaml
@@ -3,6 +3,9 @@ channels:
 - conda-forge
 - nodefaults
 dependencies:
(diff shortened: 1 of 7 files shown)


ayushdg (approved):

charlesbluca:
Looks like https://github.com/conda-forge/anyio-feedstock/pull/48 merged in, so we might be good to close this if that unblocks the failures.

PR #1167, opened 2023-05-31
Used cached Rust registry for conda builds

charlesbluca:
Hopefully this should speed up conda builds on PRs and commits

### DIFF
--- a/continuous_integration/recipe/build.sh
+++ b/continuous_integration/recipe/build.sh
@@ -3,7 +3,7 @@
 export RUST_BACKTRACE=full
 
 # pull from existing rust cache, if it exists
-if [ -d "~/.cargo/registry" ]; then ln -s ~/.cargo/registry $CONDA_PREFIX/.cargo/registry; fi
-if [ -d "~/.cargo/git" ]; then ln -s ~/.cargo/git $CONDA_PREFIX/.cargo/git; fi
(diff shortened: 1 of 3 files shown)


continuous_integration/recipe/build.sh
```diff
@@ -0,0 +1,17 @@
...
+
+export RUST_BACKTRACE=full
+
+# set up caching if we are being executed in a github workflow
+if [[ -v GITHUB_WORKSPACE ]]; then
+    if [[ ! -d "$GITHUB_WORKSPACE/dask_planner/target" ]]; then
+        mkdir "$GITHUB_WORKSPACE/dask_planner/target"
+    fi
+    ln -s $GITHUB_WORKSPACE/dask_planner/target $CONDA_PREFIX/../work/dask_planner/target
+fi
```
charlesbluca, line 11: Adding this to the script makes it so we should now be sharing work between builds on the same job (i.e. if we're building all packages in one `conda build` command intermediate compilation only needs to be done once); however, since each `conda build` command generates its own unique directory (e.g. `/usr/share/miniconda3/envs/test/conda-bld/dask-sql_1685500788622`), this work can't be shared between jobs because the source path is used as part of detecting a cache hit.

Going to explore if we can either:
- Configure `conda build` to use a reproducible work directory on every run
- Configure the Rust caching to cache on relative path versus absolute
cc @ayushdg in case you've played around with the Rust caching at all

PR #1142, opened 2023-05-09
Remove TPOT-related NumPy restrictions

sarahyurick:
Since https://github.com/EpistasisLab/tpot/pull/1280 was merged, the changes from https://github.com/dask-contrib/dask-sql/pull/1135 can be reverted.

PR #1171, opened 2023-06-05
Remove `numpy` constraint, update `tpot` min version in CI environments

charlesbluca:
It looks like the fix to https://github.com/EpistasisLab/tpot/issues/1281 is now out in `tpot=0.12`, so we should be good to remove our constraint on `numpy` and bump to using this new `tpot` release in CI; relevant now because this constraint seems to conflict with the latest cuDF nightlies, so we need to remove it in order to continue testing against latest RAPIDS.

cc @sarahyurick @ayushdg

### DIFF
--- a/continuous_integration/environment-3.10-dev.yaml
+++ b/continuous_integration/environment-3.10-dev.yaml
@@ -14,9 +14,6 @@ dependencies:
 - maturin>=0.12.8
 - mlflow
 - mock
-# tpot imports fail with numpy >=1.24.0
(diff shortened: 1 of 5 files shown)


ayushdg (approved):

ayushdg:
Ah yes we might be able to close this in favor of #1142

charlesbluca:
Yup! Sorry I missed your PR @sarahyurick, will check out the ubuntu 3.8 failures there to see if they can be reproduced locally

PR #1173, opened 2023-06-07
Add support for custom filters to `attempt_predicate_pushdown`

rjzamora:
Follow-up to https://github.com/dask-contrib/dask-sql/pull/1160

Makes it possible to combine (1) pre-existing, (2) extracted, and (3) manually-specified filters within `attempt_predicate_pushdown`. Here, we assume that these three kinds of filters must always be combined as an `and` conjunction. However, any one of these three kinds of filters may correspond to a general `Or(And(...), ...)` DNF expression.


rjzamora:
@ayushdg - This kind of functionality should make it a lot easier to manually "add on" filters that cannot be extracted from the graph.

ayushdg (approved):
Thanks a lot for this @rjzamora

PR #1181, opened 2023-06-14
[REVIEW] Bump Arrow DataFusion Python dependency to 28.0.0

jdye64:


jdye64:
Seems like building and tests are failing due to some unrelated changes from this PR and rather some upstream pandas versioning from fugue. Will investigate and hopefully find a solution.

ayushdg:
Note: If we're migrating to Maturin, either in this PR or a followup pr, the CI release scripts might also need to be adapted to work with the new Maturin build. 
I'm not sure if Maturin works with python setup.py for source builds, so might need to update the instructions for building from source too.

jdye64:
> Note: If we're migrating to Maturin, either in this PR or a followup pr, the CI release scripts might also need to be adapted to work with the new Maturin build. I'm not sure if Maturin works with python setup.py for source builds, so might need to update the instructions for building from source too.

Agreed. I just adjusted a few things and CI likely still fails. Will iterate over the failures once they pop up.

.cargo/config.toml
```diff
@@ -9,3 +9,8 @@ rustflags = [
   "-C", "link-arg=-undefined",
   "-C", "link-arg=dynamic_lookup",
 ]
+
+[target.x86_64-unknown-linux-gnu]
+rustflags = [
+    "-C", "link-arg=-undefined"
+]
```
charlesbluca, line 16: Do we want to include these flags for all builds? Interested in if there's a way to set these globally and then have the OS X builds override with their additional build flags

jdye64: We can remove this now. This was added as part of an experiment I was performing to fix linux based cargo test builds with certain versions of GCC and I failed to remove it once I had performed that test.

src/sql/optimizer/dynamic_partition_pruning.rs
```diff
@@ -433,13 +434,13 @@ fn gather_aliases(plan: &LogicalPlan) -> HashMap<String, String> {
...
                     }
                     // Sometimes a TableScan is immediately followed by a Projection, so we can
                     // still use the alias for the table
                     LogicalPlan::Projection(ref p) => {
                         if let LogicalPlan::TableScan(ref t) = *p.input {
-                            aliases.insert(s.alias.clone(), t.table_name.to_string().clone());
+                            aliases.insert(s.alias.to_string(), t.table_name.to_string().clone());
                         }
                     }
                     _ => (),
```
sarahyurick, line 446: Add `Expr::Literal(ScalarValue::Date32(i)) => i64::from(i.unwrap()),` after line 784 (after `Expr::Literal(ScalarValue::TimestampNanosecond(i, None)) => i.unwrap(),`)

src/sql/optimizer/dynamic_partition_pruning.rs
```diff
@@ -536,7 +537,7 @@ fn read_table(
                 .project(projection.clone())
                 .ok();
             if let Some(row_iter) = row_iter_result {
-                rows.extend(row_iter);
+                rows.extend(row_iter.map(|r| r.expect("Parquet error encountered")));
```
charlesbluca, line 540: cc @sarahyurick just noting that with the changes pulled in by ADP 28, the row iterator here now gives `Result<Row, ParquetError>` instead of just the `Row` - I've gone ahead and made this minimal change to unblock compilation without changing much, not sure if there's a better way of handling things here

sarahyurick: Thanks! Looks fine to me.

sarahyurick:
One way to resolve the `TypeError: An error occurred while calling the read_parquet method registered to the pandas backend.
Original Message: int() argument must be a string, a bytes-like object or a number, not 'datetime.date'` in some of the TPC-DS errors is to add the following changes (which I've pushed onto my own branch) into `dask_sql/physical/utils/filter.py`: https://github.com/sarahyurick/dask-sql/blob/adp_28/dask_sql/physical/utils/filter.py

My solution is to perform a conversion to a `pd.Timestamp`, so while this resolves the error, it depends on Pandas and may not be the best way.

Please note that the branch I've linked does not contain all of the other code changes that this PR has, so only the `dask_sql/physical/utils/filter.py` script should be copied over.

jdye64:
> One way to resolve the `TypeError: An error occurred while calling the read_parquet method registered to the pandas backend. Original Message: int() argument must be a string, a bytes-like object or a number, not 'datetime.date'` in some of the TPC-DS errors is to add the following changes (which I've pushed onto my own branch) into `dask_sql/physical/utils/filter.py`: https://github.com/sarahyurick/dask-sql/blob/adp_28/dask_sql/physical/utils/filter.py
> 
> My solution is to perform a conversion to a `pd.Timestamp`, so while this resolves the error, it depends on Pandas and may not be the best way.
> 
> Please note that the branch I've linked does not contain all of the other code changes that this PR has, so only the `dask_sql/physical/utils/filter.py` script should be copied over.

Thanks @sarahyurick! Just added your code to the commit. Thanks a ton for pointing that out.

dask_sql/physical/utils/filter.py
```diff
@@ -388,6 +389,18 @@ def _regenerate_collection(
         regen_kwargs = self.creation_info.get("kwargs", {}).copy()
         regen_kwargs = {k: v for k, v in self.creation_info.get("kwargs", {}).items()}
         regen_kwargs.update((new_kwargs or {}).get(self.layer.output, {}))
+
+        if "read_parquet" in str(func):
+            for i in range(len(regen_kwargs["filters"])):
+                new_filters = []
+                for f in regen_kwargs["filters"][i]:
+                    if isinstance(f[2], np.datetime64):
```
sarahyurick, line 397: A small oversight on my part, but this check should be a bit more careful:
```suggestion
                    if len(f) == 3 and isinstance(f[2], np.datetime64):
```

dask_sql/physical/utils/filter.py
```diff
@@ -388,6 +389,18 @@ def _regenerate_collection(
...
+
+        if "read_parquet" in str(func):
+            for i in range(len(regen_kwargs["filters"])):
+                new_filters = []
+                for f in regen_kwargs["filters"][i]:
+                    if len(f) == 3 and isinstance(f[2], np.datetime64):
+                        dt = pd.Timestamp(f[2])
+                        new_filters.append((f[0], f[1], dt))
+                    else:
+                        new_filters.append(f)
```
sarahyurick, line 401: For query 83, this changes the error from `TypeError: An error occurred while calling the read_parquet method registered to the pandas backend.
Original Message: '<=' not supported between instances of 'Timestamp' and 'str'` to `ValueError: Metadata inference failed in 'and_'.`, which I've noticed is another error that appears for other queries.

```suggestion
                    if len(f) == 3 and isinstance(f[2], np.datetime64):
                        dt = pd.Timestamp(f[2])
                        new_filters.append((f[0], f[1], dt))
                    elif len(f) == 3 and f[1] == "in":
                        new_tuple = []
                        for dt in f[2]:
                            new_tuple.append(parse_datetime(dt))
                        new_filters.append((f[0], f[1], tuple(new_tuple)))
                    else:
                        new_filters.append(f)
```

dask_sql/physical/utils/filter.py
```diff
@@ -6,6 +6,7 @@
 
 import dask.dataframe as dd
 import numpy as np
+import pandas as pd
 from dask.blockwise import Blockwise
 from dask.highlevelgraph import HighLevelGraph, MaterializedLayer
 from dask.layers import DataFrameIOLayer
```
sarahyurick, line 12: Need to do a `from dask_sql.mappings import parse_datetime` below `from dask_sql._compat import PQ_IS_SUPPORT, PQ_NOT_IN_SUPPORT`.

sarahyurick (commented):
Working toward a fix for query 83.

sarahyurick (commented):
These changes remove the datetime-related errors for query 95, but I'm seeing otherwise strange behavior now. Whenever I run the query, it fails with a `MemoryError` the first time, then an `AssertionError` if I try to rerun it. I saw this several times as I would refresh the kernel; always a `MemoryError` on the first trial and an `AssertionError` on the following trials.

Still, I'm curious to see if the `MemoryError` pops up on gpuCI.

src/sql/optimizer/dynamic_partition_pruning.rs
```diff
@@ -781,6 +782,7 @@ fn satisfies_int64(long_value: Option<i64>, filter: Expr) -> bool {
                 Expr::Literal(ScalarValue::Int32(i)) => i64::from(i.unwrap()),
                 Expr::Literal(ScalarValue::Float64(i)) => i.unwrap() as i64,
                 Expr::Literal(ScalarValue::TimestampNanosecond(i, None)) => i.unwrap(),
+                Expr::Literal(ScalarValue::Date32(i)) => i64::from(i.unwrap()),
```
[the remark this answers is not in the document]
charlesbluca: Is this something you think we would want to apply for the other `satisfies_*` functions as well?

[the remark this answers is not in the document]
sarahyurick: Sure, I don't think it would hurt to add `Expr::Literal(ScalarValue::Utf8(_)) => return false,` to the equivalent places in `satisfies_int32` and `satisfies_float`.

.github/CODEOWNERS
```diff
@@ -2,4 +2,7 @@
 *              @ayushdg @charlesbluca @galipremsagar
 
 # rust codeowners
-dask_planner/  @ayushdg @charlesbluca @galipremsagar @jdye64
+.cargo/        @ayushdg @charlesbluca @galipremsagar @jdye64
+src/           @ayushdg @charlesbluca @galipremsagar @jdye64
+Cargo.toml     @ayushdg @charlesbluca @galipremsagar @jdye64
+Cargo.lock     @ayushdg @charlesbluca @galipremsagar @jdye64
```
jakirkham, line 8: Wonder if it makes sense to start adding teams to own portions of the code and then add people to those teams

continuous_integration/recipe/conda_build_config.yaml
```diff
@@ -4,5 +4,5 @@ rust_compiler_version:
     - 1.69
 libprotobuf:
     - 3
-setuptools_rust:
-    - 1.5.2
+maturin:
+    - 0.15.3
```
jakirkham, line 8: Could drop the patch release to simply get the latest of `0.15`

```suggestion
    - 0.15
```

continuous_integration/recipe/meta.yaml
```diff
@@ -24,14 +24,17 @@ requirements:
   build:
     - python                                 # [build_platform != target_platform]
     - cross-python_{{ target_platform }}     # [build_platform != target_platform]
+    - maturin                                # [build_platform != target_platform]
     - libprotobuf                            # [build_platform != target_platform]
+    - zlib                                   # [build_platform != target_platform]
```
jakirkham, line 29: Curious how `zlib` is being used during the build process

pyproject.toml
```diff
@@ -1,13 +1,82 @@
 [build-system]
-requires = ["setuptools", "wheel", "setuptools-rust"]
+requires = ["maturin>=0.15,<0.16"]
```
jakirkham, line 2: Curious what the issues are with newer versions of `maturin`. What happens when trying 0.16 (or newer)?

Also wondering if we can move to [1.1.0]( https://github.com/PyO3/maturin/releases/tag/v1.1.0 ) (where maybe things are more stable?)

pyproject.toml
```diff
@@ -1,13 +1,82 @@
...
+[project]
+name = "dask_sql"
+description = "SQL query layer for Dask"
+maintainers = [{name = "Nils Braun", email = "nilslennartbraun@gmail.com"}]
+license = {text = "MIT"}
+classifiers = [
+    "Development Status :: 5 - Production/Stable",
+    "Intended Audience :: Developers",
+    "Intended Audience :: Science/Research",
+    "License :: OSI Approved :: BSD License",
```
jakirkham, line 14: Since this is MIT (and not BSD)

```suggestion
    "License :: OSI Approved :: MIT License",
```

pyproject.toml
```diff
@@ -1,13 +1,82 @@
...
+    "Programming Language :: Python :: 3",
+    "Programming Language :: Python :: 3 :: Only",
+    "Programming Language :: Python :: 3.8",
+    "Programming Language :: Python :: 3.9",
+    "Programming Language :: Python :: 3.10",
+    "Topic :: Scientific/Engineering",
+    "Topic :: System :: Distributed Computing",
+]
+readme = "README.md"
+urls = {Homepage = "https://github.com/dask-contrib/dask-sql/"}
```
jakirkham, line 27:
```suggestion
urls = {
    Homepage = "https://github.com/dask-contrib/dask-sql",
    Documentation = "https://dask-sql.readthedocs.io",
    Source = "https://github.com/dask-contrib/dask-sql",
}
```

continuous_integration/gpuci/environment-3.10.yaml
```diff
@@ -14,15 +14,17 @@ dependencies:
...
 - mock
 - numpy>=1.21.6
 - pandas>=1.4.0
 - pre-commit
 - prompt_toolkit>=3.0.8
 - psycopg2
-- pyarrow>=6.0.1
+- pyarrow>=6.0.2
+# TODO: remove once fastapi has resolved downstream breakage
+- pydantic<1.10.12
```
jakirkham, line 27: Might be worth rechecking this now that a newer `fastapi` can be used ( https://github.com/dask-contrib/dask-sql/pull/1205 )

pyproject.toml
charlesbluca: Did this mostly to align with [ADP's maturin constraint](https://github.com/apache/arrow-datafusion-python/blob/1fde8e47ae6714f02c2d2a088577e92e814dfa34/pyproject.toml#L19), though I could try bumping us to a more recent version to see how that impacts things.

Will follow up on https://github.com/dask-contrib/dask-sql/pull/1181#discussion_r1278100781 accordingly depending on how that goes

continuous_integration/recipe/conda_build_config.yaml
charlesbluca: Yup that makes sense, going to try bumping us to a more recent maturin version but if that doesn't work will go with this suggestion

continuous_integration/recipe/meta.yaml
charlesbluca: Can try to dig up a link but seemed like the `linux-aarch64` builds seemed to fail if this wasn't included in the build requirements in addition to host

continuous_integration/gpuci/environment-3.10.yaml
charlesbluca: Yup, looks like the newer `fastapi` has a similar constraint so should be good to remove this

charlesbluca (approved):

PR #1182, opened 2023-06-14
Add support for left semi and left anti joins

ChrisJar:
To-do:
- [x] Fix expected return columns mismatch
- [x] Fix incorrect merge column name
- [x] Create CPU implementation for `leftanti`


dask_sql/physical/rel/logical/join.py
```diff
@@ -178,6 +179,7 @@ def merge_single_partitions(lhs_partition, rhs_partition):
         # and to rename them like the rel specifies
         row_type = rel.getRowType()
         field_specifications = [str(f) for f in row_type.getFieldNames()]
+        print(field_specifications)
```
sarahyurick, line 182:
```suggestion
        if join_type == "leftanti":
            field_specifications = field_specifications[:len(cc.columns)]
```

dask_sql/physical/rel/logical/join.py
```diff
@@ -178,6 +179,7 @@ def merge_single_partitions(lhs_partition, rhs_partition):
         # and to rename them like the rel specifies
```
sarahyurick, line 179: On lines 174-176, replace
```
        correct_column_order = list(df_lhs_renamed.columns) + list(
            df_rhs_renamed.columns
        )
```
with
```
        if join_type == "leftanti":
            correct_column_order = list(df_lhs_renamed.columns)
        else:
            correct_column_order = list(df_lhs_renamed.columns) + list(
                df_rhs_renamed.columns
            )
```

dask_sql/physical/rel/logical/join.py
```diff
@@ -178,6 +179,7 @@ def merge_single_partitions(lhs_partition, rhs_partition):
...
         row_type = rel.getRowType()
         field_specifications = [str(f) for f in row_type.getFieldNames()]
+        print(field_specifications)
 
         cc = cc.rename(
```
sarahyurick, line 184: On line 190, replace
```
        cc = self.fix_column_to_row_type(cc, row_type)
```
with
```
        if join_type == "leftanti":
            cc = self.fix_column_to_row_type(cc, row_type, join_type)
        else:
            cc = self.fix_column_to_row_type(cc, row_type)
```

dask_sql/physical/rel/logical/join.py
```diff
@@ -178,6 +179,7 @@ def merge_single_partitions(lhs_partition, rhs_partition):
...
             {
```
sarahyurick, line 185: On line 207, replace
```
        dc = self.fix_dtype_to_row_type(dc, rel.getRowType())
```
with
```
        # TODO: Debug this...
        if join_type != "leftanti":
            dc = self.fix_dtype_to_row_type(dc, rel.getRowType())
```

dask_sql/physical/rel/logical/join.py
```diff
@@ -178,6 +179,7 @@ def merge_single_partitions(lhs_partition, rhs_partition):
         # and to rename them like the rel specifies
         row_type = rel.getRowType()
         field_specifications = [str(f) for f in row_type.getFieldNames()]
+        print(field_specifications)
```
sarahyurick, line 183: In the `dask_sql/physical/rel/base.py` script, replace the `fix_column_to_row_type` method with
```
@staticmethod
    def fix_column_to_row_type(
        cc: ColumnContainer, row_type: "RelDataType", join_type: str = None,
    ) -> ColumnContainer:
        """
        Make sure that the given column container
        has the column names specified by the row type.
        We assume that the column order is already correct
        and will just "blindly" rename the columns.
        """
        field_names = [str(x) for x in row_type.getFieldNames()]
        if join_type == "leftanti":
            field_names = field_names[:len(cc.columns)]

        logger.debug(f"Renaming {cc.columns} to {field_names}")
        cc = cc.rename_handle_duplicates(
            from_columns=cc.columns, to_columns=field_names
        )

        # TODO: We can also check for the types here and do any conversions if needed
        return cc.limit_to(field_names)
```

sarahyurick (commented):
Here's a (hack-y) way of handling left anti joins that at least works with a very basic example. It only involves modifying the Python side and just slicing out the irrelevant righthand side columns. Not sure if this is the best way to do things or if it needs building upon to work for more complicated queries, but let me know what you think. Happy to help explore the Rust side more if this solution is insufficient.

ChrisJar:
@sarahyurick thanks for the suggestions!

ChrisJar:
What's the current best practice for checking if a dask-sql dataframe is cpu or gpu backed without importing cudf?

sarahyurick:
> What's the current best practice for checking if a dask-sql dataframe is cpu or gpu backed without importing cudf?

I think `is_cudf_type` should work: https://github.com/dask-contrib/dask-sql/blob/f8bf06ca09459cc78426b5a9bc67d870c4d6fba9/dask_sql/utils.py#L49

sarahyurick (approved):
LGTM!

dask_sql/context.py
```diff
@@ -845,15 +845,19 @@ def _get_ral(self, sql):
...
         if dc is None:
             return
 
+        # Optimization might remove some alias projects. Make sure to keep them here.
+        select_names = [field for field in rel.getRowType().getFieldList()]
+
         if select_names:
+            cc = dc.column_container
+
+            select_names = select_names[: len(cc.columns)]
```
charlesbluca, line 859: Do you know why `getFieldList` and `getFieldNames` are returning an extra field that we need to filter out here (and in `fix_column_to_row_type`)? I want to believe there's something we could do on the Rust end to avoid needing these workarounds in the Python code, but okay with filing that as a follow up issue to address.

cc @jdye64 if you have any idea what could be going on here

dask_sql/physical/rel/base.py
```diff
@@ -30,7 +30,7 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> dd.DataFra
 
     @staticmethod
     def fix_column_to_row_type(
-        cc: ColumnContainer, row_type: "RelDataType"
+        cc: ColumnContainer, row_type: "RelDataType", join_type: str = None
```
charlesbluca, line 33: Should the type annotation here be

```suggestion
        cc: ColumnContainer, row_type: "RelDataType", join_type: str | None = None
```

Since `join_type` is an optional argument?

dask_sql/physical/rel/base.py
```diff
@@ -84,7 +86,9 @@ def assert_inputs(
         return [RelConverter.convert(input_rel, context) for input_rel in input_rels]
 
     @staticmethod
-    def fix_dtype_to_row_type(dc: DataContainer, row_type: "RelDataType"):
+    def fix_dtype_to_row_type(
+        dc: DataContainer, row_type: "RelDataType", join_type: str = None
```
charlesbluca, line 90:
```suggestion
        dc: DataContainer, row_type: "RelDataType", join_type: str | None = None
```

dask_sql/physical/rel/logical/join.py
```diff
@@ -74,6 +76,9 @@ def convert(self, rel: "LogicalPlan", context: "dask_sql.Context") -> DataContai
 
         join_type = join.getJoinType()
         join_type = self.JOIN_TYPE_MAPPING[str(join_type)]
+        # TODO: update with correct implementation of leftsemi
```
charlesbluca, line 79: Should probably file/link an issue for this if one hasn't been made yet

charlesbluca (commented):
Thanks @ChrisJar! Some minor comments around type annotations / follow up work

dask_sql/context.py
ChrisJar: Yeah it seems like it's treating left semi and left anti joins like left joins where the column merged on from both left and right tables is included in the result whereas in leftsemi and leftanti joins only the left hand column should be returned

dask_sql/physical/rel/base.py
ChrisJar: Not sure why this change isn't passing CI but it looks like we do the `argument: str = None` for optional strings in several places ex: https://github.com/dask-contrib/dask-sql/blob/5421bbf9b363ab15c657432e0c9d367c6f236df7/dask_sql/context.py#L164 maybe to avoid this issue.

charlesbluca: Ah apologies I saw this syntax being used in other places but looks like this requires `python>=3.10`; think we could instead do something like

```suggestion
        cc: ColumnContainer, row_type: "RelDataType", join_type: Optional[str] = None
```

dask_sql/physical/rel/base.py
charlesbluca: Based on your comment above think this should work

```suggestion
        dc: DataContainer, row_type: "RelDataType", join_type: Optional[str] = None
```

dask_sql/context.py
charlesbluca: Would you mind filing an issue around this and linking it here so we could follow up on this later if possible?

ChrisJar: #1192

charlesbluca (approved):

PR #1184, opened 2023-06-20
Add compatibility for pandas 2

charlesbluca:
Looks like the upstream constraints on pandas in our CI environment have been removed such that we're now pulling in pandas 2, which is raising some new errors.

Noting that after unblocking these failures, we should probably explore a combination of upstream testing for pandas along with erroring on certain warnings in our Python tests, which should allow us to more gradually follow these breaking changes as their introduced (IIUC this is the approach that Dask/Distributed take).

### DIFF
--- a/tests/integration/fixtures.py
+++ b/tests/integration/fixtures.py
@@ -170,7 +170,7 @@ def gpu_string_table(string_table):
 @pytest.fixture()
 def gpu_datetime_table(datetime_table):
     if cudf:
-        # cudf doesn't have support for timezoned datetime data
+        # cudf doesn't have full support for timezone-aware data
         df = datetime_table.copy()
         df["timezone"] = df["timezone"].dt.tz_localize(None)
         df["utc_timezone"] = df["utc_timezone"].dt.tz_localize(None)
--- a/tests/integration/fixtures.py
+++ b/tests/integration/fixtures.py
@@ -169,12 +169,14 @@ def gpu_string_table(string_table):
 
 @pytest.fixture()
 def gpu_datetime_table(datetime_table):
-    # cudf doesn't have support for timezoned datetime data
-    datetime_table["timezone"] = datetime_table["timezone"].astype("datetime64[ns]")
-    datetime_table["utc_timezone"] = datetime_table["utc_timezone"].astype(
-        "datetime64[ns]"
-    )
-    return cudf.from_pandas(datetime_table) if cudf else None
+    if cudf:
+        # cudf doesn't have support for timezoned datetime data
+        datetime_table["timezone"] = datetime_table["timezone"].dt.tz_localize(None)
+        datetime_table["utc_timezone"] = datetime_table["utc_timezone"].dt.tz_localize(
+            None
+        )
+        return cudf.from_pandas(datetime_table)
+    return None
 
 
 @pytest.fixture()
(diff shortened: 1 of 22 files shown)


tests/integration/fixtures.py
```diff
@@ -169,12 +169,13 @@ def gpu_string_table(string_table):
...
-        "datetime64[ns]"
-    )
-    return cudf.from_pandas(datetime_table) if cudf else None
+    if cudf:
+        # cudf doesn't have support for timezoned datetime data
+        df = datetime_table.copy()
+        df["timezone"] = df["timezone"].dt.tz_localize(None)
+        df["utc_timezone"] = df["utc_timezone"].dt.tz_localize(None)
+        return cudf.from_pandas(df)
+    return None
```
charlesbluca, line 178: Before this change, we were actually overwriting the `datetime_table` CPU fixture such that it didn't contain timezone-aware data.

Now that we're no longer doing that, a failure has been exposed in `test_filter_cast_timestamp` because filtering a timezone-aware column by a timestamp would require non-UTC timezone handling for literals on the Python end:

https://github.com/dask-contrib/dask-sql/blob/899170679b44b8eebd5efe1af373eded98cbe9aa/dask_sql/physical/rex/core/literal.py#L181-L182

There are a few ways we could handle this, but all of them generally involve front-facing changes to timestamp literal handling so figure it might make sense to break that off into a separate PR.

Okay with xfailing that test for now to get this in? cc @ayushdg @jdye64

tests/integration/fixtures.py
```diff
@@ -169,12 +169,17 @@ def gpu_string_table(string_table):
...
+    if cudf:
+        # TODO: remove once `from_pandas` has support for timezone-aware data
+        # https://github.com/rapidsai/cudf/issues/13611
+        df = datetime_table.copy()
+        df["timezone"] = df["timezone"].dt.tz_localize(None)
+        df["utc_timezone"] = df["utc_timezone"].dt.tz_localize(None)
+        gdf = cudf.from_pandas(df)
+        gdf["timezone"] = gdf["timezone"].dt.tz_localize("Europe/Berlin")
+        gdf["utc_timezone"] = gdf["utc_timezone"].dt.tz_localize("UTC")
+        return gdf
```
ayushdg, line 181: Nit: Is it possible to get the tz from the cpu df rather than hardcoding to eu/berlin here?

Alternatively we can consider renaming the `timezone` column to make it clearer that it's EU/Berlin?

tests/integration/test_filter.py
```diff
@@ -110,6 +109,7 @@ def test_filter_cast_date(c, input_table, request):
         ),
     ],
 )
+@pytest.mark.xfail(reason="Need support for non-UTC timezoned literals")
```
ayushdg, line 112: Might be worth opening and linking to an issue here if possible (but not blocking)

ayushdg (approved):
Minor suggestions but nothing blocking for the merge.

tests/integration/fixtures.py
charlesbluca: Makes sense, I'm going to go with implicitly grabbing the tz to avoid a larger diff, ideally should be able to remove most of this once https://github.com/rapidsai/cudf/issues/13611 is resolved

ayushdg (approved):

PR #1194, opened 2023-06-30
`datetime_coercion` rule

### LINKED ISSUE #1189: [BUG] Int64-to-Timestamp coercion
Currently, something like
```
SELECT * FROM d_table d1, d_table d2 WHERE d2.d_date > d1.d_date + 5
```
fails with `OptimizationException('Plan("Timestamp(Nanosecond, None) + Int64 can't be evaluated because there isn't a common type to coerce the types to")')`. The assumption here is that we are adding 5 days to `d1.d_date`.

It appears that we are able to handle the unoptimized plan just fine, but the issue comes up once we try to run our optimizer.

So far, I've considered adding an optimizer rule on our side which converts an `Int64` to a timestamp for this scenario, but I was having trouble getting this to work as the rule did not appear to be executing for some reason. Perhaps we can try catching the error and doing a conversion before generating the unoptimized plan?

sarahyurick:
Closes https://github.com/dask-contrib/dask-sql/issues/1189


sarahyurick:
This PR is intended to help query 72, which was erroring with an `OptimizationException`. This PR fixes that bug, and the query now raises memory errors which will need to be explored further in future work.

dask_planner/src/sql/preoptimizer.rs
```diff
@@ -0,0 +1,212 @@
...
+fn extract_columns_and_literals(expr: &Expr) -> Vec<(Vec<Expr>, Expr)> {
+    let mut result = Vec::new();
+    if let Expr::BinaryExpr(b) = expr {
+        let left = *b.left.clone();
+        let right = *b.right.clone();
+        if let Operator::Plus
+        | Operator::Minus
+        | Operator::Multiply
+        | Operator::Divide
+        | Operator::Modulo = &b.op
```
sarahyurick, line 35: It doesn't really make sense to multiply/divide/modulo a datetime, so should maybe remove these.

tests/integration/test_rex.py
```diff
@@ -1238,3 +1238,34 @@ def test_scalar_timestamps(c, gpu):
...
+            "x": [1, 2, 3, 4],
+        }
+    )
+    c.create_table("d_table", d_table)
+
+    df = c.sql(
+        """
+        SELECT * FROM d_table d1, d_table d2
+        WHERE d2.x < d1.x + (1 + 2)
+        AND d2.d_date > d1.d_date + (2 + 3)
```
sarahyurick, line 1261: The parentheses don't matter here. This is just to demonstrate that this PR can handle something like `d1.d_date + 1`, `d1.d_date + 2 + 3`, `d1.d_date + 4 + 5 + 6`, etc. for as many mathematical operands as the user provides.

jdye64 (approved):
I really like this approach of introducing a "pre-processing" phase instead of trying to recover from errors from Datafusion and potentially attempt to re-run them. This is much more clean for our purposes. Thanks for adding this @sarahyurick

sarahyurick:
Thanks @jdye64 ! With https://github.com/dask-contrib/dask-sql/pull/1181 merged, this should be good to go.

charlesbluca (approved):

sarahyurick:
Hi @charlesbluca I wanted to ask about this getting merged in?

charlesbluca:
Sorry this has sat on the backburner for a while - might make sense to merge in the ongoing work in #1249, as I imagine we'll want to prioritize that for release and it includes a bump from ADP 28 -> 32 that could affect things here

sarahyurick:
Thanks @charlesbluca !

Tested locally with the latest main and can confirm that this PR still adds functionality that did not pass before. Should be ready to go.

sarahyurick:
@charlesbluca looks like https://github.com/dask-contrib/dask-sql/pull/1265 is related to the failures here? They're all giving a:
```
error: package `dask-sql v2023.11.0-rc1 (/Users/runner/work/dask-sql/dask-sql)` cannot be built because it requires rustc 1.73 or newer, while the currently active rustc version is 1.72.1
```

charlesbluca:
Yeah think the issue here is that the GHA runners for macOS have an older installation of Rust than the Linux/Windows runners? Out of simplcity think it might be easiest to drop the minimum dependency to 1.72 since IIUC we only need `rust>=1.70` (cc @jdye64), trying this out now in #1266

charlesbluca:
Thanks for the work here @sarahyurick!

Issue #1069, opened 2023-03-01
[ENH] Join reordering improvements


sarahyurick:
https://github.com/dask-contrib/dask-sql/pull/1027 added basic join reordering functionality. However, there are several things that need to be fixed and improved upon:

- [ ] Probably the biggest improvement we need is to support join filters correctly. This should allow us to run query 72, which is expected to have the largest performance gain with join reordering. Other queries affected include queries 75 and 93
- [ ] Ideally, the `fact_dimension_ratio` should be 0.3, but queries 17, 25, 29, and 85 currently fail without a stricter ratio. This suggests that more work needs to be done with reordering fact-to-fact joins
- [ ] Check for deterministic filter expressions
- [ ] Add Rust tests

PR #1199, opened 2023-07-07
Rust optimizer improvements

sarahyurick:
From https://github.com/dask-contrib/dask-sql/issues/1121:

- [x] Allow the fact_dimension_ratio and possibly other parameters to be specified by the user
- [x] Modify the c.explain() function to cut off large strings of INLIST vectors
- [x] DPP functionality PyTests
- [x] DPP config PyTests

From https://github.com/dask-contrib/dask-sql/issues/1069:

- [x] Consider exposing parameters to the user


dask_planner/src/sql/optimizer.rs
```diff
@@ -39,7 +39,12 @@ pub struct DaskSqlOptimizer {
 impl DaskSqlOptimizer {
     /// Creates a new instance of the DaskSqlOptimizer with all the DataFusion desired
     /// optimizers as well as any custom `OptimizerRule` trait impls that might be desired.
-    pub fn new() -> Self {
+    pub fn new(
+        fact_dimension_ratio: Option<f64>,
+        max_fact_tables: Option<usize>,
+        preserve_user_order: Option<bool>,
+        filter_selectivity: Option<f64>,
```
sarahyurick, line 46: It's kind of annoying/bulky to have to pass in these parameters every time we do a `DaskSqlOptimizer::new()`, but since there's really only 2 places in our code where we do this, I figured it's not too bad.

dask_planner/src/sql/optimizer.rs
```diff
@@ -91,9 +101,13 @@ impl DaskSqlOptimizer {
...
     // runs one time
-    pub fn dynamic_partition_pruner() -> Self {
-        let rule: Vec<Arc<dyn OptimizerRule + Sync + Send>> =
-            vec![Arc::new(DynamicPartitionPruning::new())];
+    pub fn dynamic_partition_pruner(fact_dimension_ratio: Option<f64>) -> Self {
+        let rule: Vec<Arc<dyn OptimizerRule + Sync + Send>>;
+        if let Some(f) = fact_dimension_ratio {
+            rule = vec![Arc::new(DynamicPartitionPruning::new(f))];
+        } else {
+            rule = vec![Arc::new(DynamicPartitionPruning::default())];
```
sarahyurick, line 109: Instead of having a `DynamicPartitionPruning::default` AND a `DynamicPartitionPruning::new()`, it might be cleaner to just do the latter and have a `fact_dimension_ratio.unwrap_or(0.3)` here.

dask_sql/sql-schema.yaml
```diff
@@ -69,6 +69,38 @@ properties:
         description: |
           Whether to apply the dynamic partition pruning optimizer rule.
 
+      verbose_optimizer:
```
sarahyurick, line 72: I really don't like the name `verbose_optimizer`, but I couldn't really think of a better name...

jdye64: Yeah I'm no naming expert. I like the feature though.

dask_sql/context.py
```diff
@@ -542,11 +550,16 @@ def explain(
             :obj:`str`: a description of the created relational algebra.
 
         """
+        dynamic_partition_pruning = dask_config.get("sql.dynamic_partition_pruning")
+        if not dask_config.get("sql.verbose_optimizer"):
```
jdye64, line 554: What am I missing here? Seems like getting the config value for `sql.verbose_optimizer` but then setting values related to `sql.dynamic_partition_pruning`?

jdye64 (changes requested):

dask_sql/context.py
sarahyurick: So when I was working on the DPP PR, I found it pretty annoying that the `c.explain()` output would contain such long `INLIST` expressions (hundreds and thousands of values), which aren't very useful for a user who maybe wouldn't even understand where all the values are coming from. With `verbose_optimizer` set to False, though, it just affects the `c.explain()` function, temporarily turning off DPP so that we don't get the long `INLIST` expressions and instead just get what the `LogicalPlan` would be if DPP was turned off. Then at the end we switch DPP back on (if it was originally set on), so that the query itself can still run with DPP.

So if DPP is on and `verbose_optimizer` is False, DPP will still run with `c.sql()`, but `c.explain()` won't contain the DPP values.

sarahyurick:
Hi @charlesbluca I wanted to ask about this getting merged in?

sarahyurick:
@charlesbluca same rustc version errors here?

sarahyurick:
@charlesbluca this should be OK now, thanks!

dask_sql/context.py
charlesbluca: Conceptually I like what this is aiming to accomplish, but don't really like the idea of implicitly skipping optimizations without warning the user. Feel like maybe it might make more sense to document the introduction of the verbose logical plans in the schema for `sql.dynamic_partition_pruning`, with a suggestion to disable this config for `c.explain()` if this behavior isn't desired.

dask_sql/sql-schema.yaml
charlesbluca: Generally would rather avoid this configuration for now, but if we do have it maybe it would make sense as something like `sql.explain.skipped_optimizers`? Then it could be a list where users pass any arbitrary rules, but it defaults to whatever rules we always want disabled for explains.

dask_sql/context.py
sarahyurick: Right now the default is to run DPP but to not have it in `c.explain()`. This is because having it in `c.explain()` would introduce filters that are not specified by the user (i.e., filters that are added by DPP) which potentially makes the `LogicalPlan` more difficult for the user to interpret. So we're not actually skipping DPP, we're just not printing how it affects the `LogicalPlan` for the user to see.

I like the idea of renaming it to `sql.explain.skipped_optimizers` or `sql.dynamic_partition_pruning.verbose` to make it more clear, although I do prefer having it default to not having DPP printed in `c.explain()` (but still run DPP by default).

sarahyurick:
Hi @charlesbluca LMK what you think of this.

charlesbluca (approved):
Thanks @sarahyurick!

PR #1205, opened 2023-07-27
Bump `fastapi` dependency to `>=0.92.0`

charlesbluca:
Doing this to avoid https://github.com/advisories/GHSA-3qj8-93xh-pwh2; interested in what needs to change to allow this


continuous_integration/environment-3.10-dev.yaml
```diff
@@ -5,9 +5,9 @@ channels:
 dependencies:
 - c-compiler
 - dask>=2022.3.0
-# FIXME: handling is needed for httpx-based fastapi>=0.87.0
-- fastapi>=0.69.0,<0.87.0
```
jakirkham, line 9: For more context, `fastapi` version `0.86.0` [pins to `starlette` version `0.20.4`]( https://github.com/tiangolo/fastapi/blob/0.86.0/pyproject.toml#L42 ), which was the last release before `0.87.0` (and so the latest version of `fastapi` allowed by this pin). Hence why the `starlette` installed has that CVE

continuous_integration/environment-3.9-dev.yaml
```diff
@@ -5,9 +5,9 @@ channels:
 dependencies:
 - c-compiler
 - dask>=2022.3.0
-# FIXME: handling is needed for httpx-based fastapi>=0.87.0
-- fastapi>=0.69.0,<0.87.0
+- fastapi>=0.92.0
```
jakirkham, line 8: Noticing that allowing any `fastapi` version `0.92.0` or greater results in [a bunch of test failures]( https://github.com/dask-contrib/dask-sql/actions/runs/5685404699/job/15410229312?pr=1205#step:8:42129 )

continuous_integration/environment-3.8-dev.yaml
```diff
@@ -5,8 +5,9 @@ channels:
 dependencies:
 - c-compiler
 - dask=2022.3.0
-- fastapi=0.69.0
+- fastapi=0.92.0
```
jakirkham, line 8: However restricting to `0.92.0` results in [only one test failure]( https://github.com/dask-contrib/dask-sql/actions/runs/5685404699/job/15410230957?pr=1205#step:8:33 )

<details>

````python
_________________________________ test_routes __________________________________
[gw1] linux -- Python 3.8.17 /usr/share/miniconda3/envs/dask-sql/bin/python3.8

app_client = <starlette.testclient.TestClient object at 0x7fe5582847f0>

    def test_routes(app_client):
        assert app_client.post("/v1/statement", data="SELECT 1 + 1").status_code == 200
>       assert app_client.get("/v1/statement", data="SELECT 1 + 1").status_code == 405
E       TypeError: get() got an unexpected keyword argument 'data'

tests/integration/test_server.py:61: TypeError
````

</details>

**Review by jakirkham — COMMENTED** [review]
Thanks Charles! 🙏

Noticing the latest `fastapi` is causing more test failures than some of the older ones. We might consider keeping an upper bound, but just bumping that upper bound as well (based on whichever version causes the least issues)

That way we could then fix at most a handful of issues (or maybe just one)

Alternatively the issues in the newer version seem to repeat across tests. So maybe there is only one needed change to adapt to the newer versions of `fastapi`

File: continuous_integration/environment-3.9-dev.yaml (hunk 2 above)
Comment (charlesbluca) [reply continuous_integration/environment-3.9-dev.yaml:8]: IIUC, those test failures are unrelated and caused by (what I assume) is new default handling for pyarrow strings in Dask, shouldn't be too difficult to resolve and then merge into this PR

Comment (charlesbluca) [reply continuous_integration/environment-3.9-dev.yaml:8]: Opened https://github.com/dask-contrib/dask-sql/pull/1206 to verify that these issues are isolated from the `fastapi` version; if things are green there, can merge with this branch and push on the one remaining test failure

File: continuous_integration/environment-3.8-dev.yaml (hunk 3 above)
Comment (jakirkham) [inline continuous_integration/environment-3.8-dev.yaml:8]: Think we want `>=` here to match what we have elsewhere in this PR

```suggestion
- fastapi>=0.92.0
```

jakirkham:
Looks like this is now passing! 🎉

continuous_integration/environment-3.8-dev.yaml
ayushdg: The py 3.8 environment file generally tests tests against our minimum specified deps (Added in #555), so I think this might be intentional to test if things still work with 0.92 even if conda picks up a newer version.

jakirkham: Ah ok. Wasn't aware of that. Happy to revert

ayushdg (approved):

continuous_integration/environment-3.9-dev.yaml
jakirkham: Fixed by PR ( https://github.com/dask-contrib/dask-sql/pull/1206 ). Marking resolved

continuous_integration/environment-3.8-dev.yaml
jakirkham: Fixed by PR ( https://github.com/dask-contrib/dask-sql/pull/1206 ). Marking resolved

jakirkham:
Thanks all! 🙏

Merging so we can get this fix out

Happy to follow up on anything else as needed 🙂

PR #1242, opened 2023-10-12
See what happens if we ignore `libprotobuf`'s runtime exports

charlesbluca:
Exploring alternatives to https://github.com/conda-forge/dask-sql-feedstock/pull/74

cc @bdice @jakirkham


continuous_integration/recipe/meta.yaml
```diff
@@ -48,6 +48,9 @@ requirements:
     - pygments >=2.7.1
     - tabulate
 
+  ignore_run_exports:
```
bdice, line 51: Does this have to go in the top level `build` key? https://docs.conda.io/projects/conda-build/en/stable/resources/define-metadata.html#export-runtime-requirements

PR #1244, opened 2023-10-13
Remove versioneer

charlesbluca:
- Bump ucx-py version
- Remove versioneer


Cargo.toml
```diff
@@ -1,7 +1,7 @@
 [package]
 name = "dask-sql"
 repository = "https://github.com/dask-contrib/dask-sql"
-version = "2023.10.0"
+version = "0.1.0"
```
jdye64, line 4: Is this version right? Could be and I'm out of the loop just want to check

charlesbluca: When we made the initial switch to maturin I opted to use the version of the python package here similar to ADP, but in hindsight we really don't need to toil over a Rust package version since we aren't publishing to crates.io; figured I'd set this to a generic semantic version in that case

Does that seem reasonable to you?

jdye64 (approved):

PR #1269, opened 2023-11-20
More restrictive input validation during predicate pushdown filter extraction

charlesbluca:
While running some TPC-H queries using dask-sql + dask-cudf, I noticed some failures around predicate pushdown that seemed to be caused by us incorrectly pushing down blockwise comparisons between columns.

This PR attempts to tighten the logic of `_blockwise_comparison_dnf` to block this edge case; marking this as draft as I'm not sure if we want to block all RHS `getitem` calls yet.

cc @rjzamora

### DIFF
--- a/dask_sql/physical/utils/filter.py
+++ b/dask_sql/physical/utils/filter.py
@@ -518,6 +518,10 @@ def _inv(symbol: str):
 def _blockwise_comparison_dnf(op, indices: list, dsk: RegenerableGraph) -> DNF:
     # Return DNF expression pattern for a simple comparison
     left = _get_blockwise_input(0, indices, dsk)
+
+    # at this point, RHS must be a scalar and not a column name (i.e. getitem task)
+    if indices[1][1] is not None:
+        raise ValueError
     right = _get_blockwise_input(1, indices, dsk)
 
     if is_arraylike(left) and hasattr(left, "item") and left.size == 1:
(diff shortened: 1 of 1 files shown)


dask_sql/physical/utils/filter.py
```diff
@@ -518,6 +518,10 @@ def _inv(symbol: str):
 def _blockwise_comparison_dnf(op, indices: list, dsk: RegenerableGraph) -> DNF:
     # Return DNF expression pattern for a simple comparison
     left = _get_blockwise_input(0, indices, dsk)
+
+    # at this point, RHS must be a scalar and not a column name (i.e. getitem task)
+    if indices[1][1] is not None:
+        raise ValueError
```
rjzamora, line 524: Yeah, I think this criteria makes sense. The right side of the comparison must be literal/scalar user input. I know this is still a draft, but here are some thoughts while I happen to be looking at this:

Maybe the `ValueError` should contain an informative message?

Is it possible for the current structure of `indices` to produce something like `(1, ">", "col-A")` instead of `("col-A, "<", 1)`? If so, maybe we want to normalize the `left`, `right`, and `symbol` upfront:

```python
def _blockwise_comparison_dnf(op, indices: list, dsk: RegenerableGraph) -> DNF:

    if indices[0][1] is not None and indices[1][1] is None:
        left = _get_blockwise_input(0, indices, dsk)
        right = _get_blockwise_input(1, indices, dsk)
        symbol = _comparison_symbols[op]
    elif indices[0][1] is None and indices[1][1] is not None:
        left = _get_blockwise_input(1, indices, dsk)
        right = _get_blockwise_input(0, indices, dsk)
        symbol = _inv(_comparison_symbols[op])
    else:
        raise ValueError("Failed to build valid filter expression.")

    if is_arraylike(right) and hasattr(right, "item") and right.size == 1:
        right = right.item()
    return DNF((left, symbol, right))
```
(Note: Not completely sure this is "correct")

PR #1280, opened 2023-12-04
[REVIEW] Fix `dynamic_partition_pruning` to handle a single Parquet file

### LINKED ISSUE #1268: [BUG] `dynamic_partition_pruning::read_table` errors on single-file Parquet datasets
**What happened**:
While playing around with [hooking dask-sql into Coiled's benchmarks](https://github.com/charlesbluca/benchmarks/blob/tpch-rapids-dask-sql/tests/tpch/test_dask_sql.py), I noticed some issues around DPP with `test_query_3`:

```
pyo3_runtime.PanicException: called `Result::unwrap()` on an `Err` value: Os { code: 20, kind: NotADirectory, message: "Not a directory" }
```

Think this is because `dynamic_partition_pruning::read_table` assumes we're working with a directory of chunked parquet files and doesn't have handling for the case where we have a single parquet file:

https://github.com/dask-contrib/dask-sql/blob/2f068f95ecb39e5d9c40addfb50d12abd00db8f9/src/sql/optimizer/dynamic_partition_pruning.rs#L488

**What you expected to happen**:
In general, I would expect this to emit a warning and skip DPP rather than bubble up to an error, though I don't think it should be too difficult to modify the handling of `tables` for the single file case? cc @sarahyurick 

**Environment**:

- dask-sql version: 2023.10.1
- Python version: 3.9
- Operating System: ubuntu20.04
- Install method (conda, pip, source): conda

sarahyurick:
Closes https://github.com/dask-contrib/dask-sql/issues/1268


sarahyurick:
Hi @charlesbluca it looks like the gpuCI failure is unrelated?

sarahyurick:
Hi @charlesbluca LMK what you think of this.

charlesbluca (commented):
Thanks @sarahyurick! Can verify locally that this resolved #1268; is there a trivial way we can add a test around this behavior?

charlesbluca:
rerun tests

charlesbluca (approved):
Thanks @sarahyurick!

PR #1290, opened 2024-01-11
Bump datafusion-python to 34.0.0

charlesbluca:


src/sql/optimizer.rs
```diff
@@ -81,8 +79,8 @@ impl DaskSqlOptimizer {
             Arc::new(SimplifyExpressions::new()),
             Arc::new(UnwrapCastInComparison::new()),
             // Arc::new(CommonSubexprEliminate::new()),
-            Arc::new(PushDownProjection::new()),
-            Arc::new(EliminateProjection::new()),
```
andygrove, line 85: I think that these two rules were combined into a new `OptimizeProjections` rule, so perhaps we should be using that now?

andygrove (approved):
LGTM. I left a comment about a missing optimizer rule. Nice to see the project move back to a using a versioned release of DataFusion.

Michael-J-Ward:
Is there something that is keeping `dask-sql` using the pinned version of `datafusion-python`?

Asking because if there is, I'd like to fix it.

PR #1296, opened 2024-01-23
Drop latest datafusion testing from nightly tests

charlesbluca:
I've found that in practice, the nightly tests against latest DataFusion don't really provide much actionable information, and increases the toil around diagnosing failures coming from latest Dask, which are relatively common.

I've still left the logic for using latest DataFusion in PRs, so we still have a mechanism to it if needed.

cc @ayushdg


ayushdg:
Agreed, thanks for the cleanup @charlesbluca.

PR #1301, opened 2024-01-31
Pin dask to 2024.1.1

charlesbluca:
As part of large Dask pinning discussions, agreed with @pentschev @rjzamora that for now it makes sense to pin dask-sql to the same Dask dependency as RAPIDS, but _not_ to fully invest in `rapids-dask-dependency`.


charlesbluca:
Looks like we'll need to drop `tpot` from our 3.12 deps as it is unsupported and added a new constraint recently:

https://github.com/conda-forge/conda-forge-repodata-patches-feedstock/pull/673

continuous_integration/environment-3.9.yaml
```diff
@@ -26,6 +26,7 @@ dependencies:
 - pytest-xdist
 - pytest
 - python=3.9
+- py-xgboost=1.7.0
```
charlesbluca, line 29: Noting that we'll need to make a follow-up PR to dask-build-environment as this pinning will definitely conflict with the RAPIDS xgboost.

PR #1328, opened 2024-04-24
Avoid using the ``to/from_dask_dataframe`` dask-cudf API

rjzamora:
Using this API is not compatible with dask-expr.  See: https://github.com/rapidsai/cudf/pull/15592


charlesbluca (approved):
Thanks for this @rjzamora! Looks like there are some linting issues and some GPU CI errors that I don't _think_ are related, but do seem worth investigating.

Do you mind running `pre-commit run --all-files` and pushing up the changes so we can see if the GPU failures are intermittent?

